PR #7, opened 2019-12-17
Change voe_signal property from dict to array in ingest script.

ThomasSchellenbergNextCentury:
Hey Clark, I'd like to change the voe_signal property from a dict:

```
voe_signal: {
  1: A,
  2: B,
  3: C,
  ...
}
```

to an array:

```
voe_signal: [{
  frame: 1,
  value: A
}, {
  frame: 2,
  value: B
}, {
  frame: 3,
  value: C
}, {
  ...
}]
```

Please test these changes to verify they work as expected.  You can change the nested property names ("frame" and "value") if desired.  No rush.  Thanks!

### DIFF
--- a/ingest_results/create_json_ingest.py
+++ b/ingest_results/create_json_ingest.py
@@ -263,7 +263,7 @@ class JsonImportCreator:
         # Handle case where this did not work
 
     def get_voe_signal(self, filename):
-        voe_signal = self.nested_dict(4, float)
+        voe_signal = self.nested_dict(3, float)
         with zipfile.ZipFile(filename) as my_zip:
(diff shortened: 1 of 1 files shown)


brianpippin:
If these changes go through, they will probably break the analysis_ui page, until it can be updated to accommodate these changes.   Not saying that we shouldn't make the changes, just wanted a side effect to be known.  The fix shouldn't take too long, but analysis_ui wouldn't show the plots and might even show an error until it was fixed.

ThomasSchellenbergNextCentury:
@brianpippin Ah, good to know, thank you!  This is why we do PRs!

For background, this change would allow us to show the line charts in Neon.  I'm not sure if that's important to do right now, but I thought, if it's an easy change, why not.

PR #8, opened 2020-02-20
MCS-21

rartiss55:
Adding support for RotateLook by allowing rotation and horizon arguments to be passed in. Currently, the run_mcs_environment script will rotate/look to the upper left, then rotate and look towards the right and the floor for a few steps. There will be an additional step for seeing what happens when an invalid rotation or horizon angle is specified. 

Associated ai2thor fork changes here:
https://github.com/NextCenturyCorporation/ai2thor/pull/2

### DIFF
--- a/ai2thor_wrapper/controller_ai2thor.py
+++ b/ai2thor_wrapper/controller_ai2thor.py
@@ -51,46 +51,42 @@ class Controller_AI2THOR(Controller_MCS):
         self.__step_number = 0
         return self.wrap_output(self.__controller.step(self.wrap_step(action='Initialize', sceneConfig=config_data)))
 
-    def validate_params(self, **kwargs):
-        isValid = True
-
+    # TODO: may need to reevaluate validation strategy/error handling in the future
+    # Need a validation/conversion step for what ai2thor will accept as input
+    # to keep parameters more simple for the user (in this case, wrapping
+    # rotation degrees into an object)
+    def validate_and_convert_params(self, **kwargs):
         rotation = kwargs.get(self.ROTATION_KEY, 0)
         horizon = kwargs.get(self.HORIZON_KEY, 0)
 
         if rotation > self.MAX_ROTATION or rotation < self.MIN_ROTATION:
-            print('Value of rotation needs to be between ' + str(self.MIN_ROTATION) + ' and ' + str(self.MAX_ROTATION) + '. Current value: ' + str(rotation))
-            isValid = False
+            print('Value of rotation needs to be between ' + str(self.MIN_ROTATION) + \
+                ' and ' + str(self.MAX_ROTATION) + '. Current value: ' + str(rotation) + \
+                '. Will be reset to 0.')
+            rotation = 0
 
         if horizon > self.MAX_HORIZON or horizon < self.MIN_HORIZON:
-            print('Value of horizon needs to be between ' + str(self.MIN_HORIZON) + ' and ' + str(self.MAX_HORIZON) + '. Current value: ' + str(horizon))
-            isValid = False
+            print('Value of horizon needs to be between ' + str(self.MIN_HORIZON) + \
+                ' and ' + str(self.MAX_HORIZON) + '. Current value: ' + str(horizon)+ \
+                '. Will be reset to 0.')
+            horizon = 0
 
-        return isValid
-
-    # Need a conversion step for what ai2thor will accept as input
-    # to keep parameters more simple for the user (in this case, wrapping
-    # rotation degrees into an object)
-    def convert_params(self, **kwargs):
         rotation_vector = {}
-        rotation_vector['y'] = kwargs.get(self.ROTATION_KEY, 0)
+        rotation_vector['y'] = rotation
 
         return dict(
             rotation=rotation_vector,
-            horizon=kwargs.get(self.HORIZON_KEY, 0)
+            horizon=horizon
         )
 
     # Override
     def step(self, action, **kwargs):
         self.__step_number += 1
 
-        # TODO: we might want a different strategy than "Pass" for invalid input in the future
-        if self.validate_params(**kwargs) == False:
-            print('Passing due to validation errors')
-            action='Pass'
+        params = self.validate_and_convert_params(**kwargs)
 
-        params = self.convert_params(**kwargs)
-
-        return self.wrap_output(self.__controller.step(self.wrap_step(action=action, rotation=params.get(self.ROTATION_KEY), horizon=params.get(self.HORIZON_KEY))))
+        return self.wrap_output(self.__controller.step(self.wrap_step(action=action, \
+            rotation=params.get(self.ROTATION_KEY), horizon=params.get(self.HORIZON_KEY))))
 
     def retrieve_action_list(self, scene_event):
         # TODO Return the list of AI2-THOR actions based on the player's simulated age, position (lying, crawling, or standing), and nearby or held objects
--- a/ai2thor_wrapper/controller_ai2thor.py
+++ b/ai2thor_wrapper/controller_ai2thor.py
@@ -17,6 +17,14 @@ class Controller_AI2THOR(Controller_MCS):
 
     __step_number = 0
 
+    MAX_ROTATION = 360
+    MIN_ROTATION = -360
+    MAX_HORIZON = 90
+    MIN_HORIZON = -90
+
+    ROTATION_KEY = 'rotation'
+    HORIZON_KEY = 'horizon'
+
     def __init__(self, unity_app_path):
         super().__init__()
         self.__controller = ai2thor.controller.Controller(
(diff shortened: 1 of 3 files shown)


ai2thor_wrapper/controller_ai2thor.py
```diff
@@ -43,10 +43,46 @@ def reset_scene(self, config_name, config_data):
...
+        )
+
     # Override
-    def step(self, action):
+    def step(self, action, **kwargs):
         self.__step_number += 1
-        return self.wrap_output(self.__controller.step(self.wrap_step(action=action)))
+
+        # TODO: we might want a different strategy than "Pass" for invalid input in the future
+        if self.validate_params(**kwargs) == False:
```
rartiss55: I didn't put too much effort in here now since we are going to refactor things, but we might want to think about what we want our strategy to be for invalid inputs/what we want to pass onto ai2thor, and whether error handling needs to be part of the Python API or the ai2thor code. Like the note says, for now I am simply changing the action to "Pass" if there's any issues with input.

ai2thor_wrapper/run_mcs_environment.py
```diff
@@ -15,10 +15,24 @@ def run_scene(controller, config_name, config_data):
...
     print('step=' + str(output.step_number))
 
-    for i in range(1, 31):
-        output = controller.step('Pass')
+    #for i in range(1, 31):
+    #    output = controller.step('Pass')
+    #    print('step=' + str(output.step_number))
+
+    # Testing RotateLook
+    rotateAngle = -45
```
rartiss55: We'll need to have values where we track/increment the rotation and horizon since rotation is absolute in terms of ai2thor but we will want it to be relative to the user, although this may change later.

ai2thor_wrapper/controller_ai2thor.py
```diff
@@ -43,10 +51,42 @@ def reset_scene(self, config_name, config_data):
         self.__step_number = 0
         return self.wrap_output(self.__controller.step(self.wrap_step(action='Initialize', sceneConfig=config_data)))
 
+    # TODO: may need to reevaluate validation strategy/error handling in the future
+    # Need a validation/conversion step for what ai2thor will accept as input
+    # to keep parameters more simple for the user (in this case, wrapping
+    # rotation degrees into an object)
+    def validate_and_convert_params(self, **kwargs):
```
rartiss55: I didn't put too much effort in here now since we are going to refactor things, but we might want to think about what we want our strategy to be for invalid inputs/what we want to pass onto ai2thor, and whether error handling needs to be part of the Python API or the ai2thor code (we likely want to start by adding it the Python API side of things to help better navigate maintaining the ai2thor fork, but that could change as things get more complex). Per the Python API Confluence doc, I updated this to simply set invalid user inputs to zero, and for now printing an error (we can figure out the return strings of SUCCESSFUL/SUCCESSFUL_WITH_INVALID_INPUTS/etc later when this is refactored).

ai2thor_wrapper/controller_ai2thor.py
```diff
@@ -43,10 +51,46 @@ def reset_scene(self, config_name, config_data):
...
-        return self.wrap_output(self.__controller.step(self.wrap_step(action=action)))
+
+        # TODO: we might want a different strategy than "Pass" for invalid input in the future
+        if self.validate_params(**kwargs) == False:
+            print('Passing due to validation errors')
+            action='Pass'
+
+        params = self.convert_params(**kwargs)
+
+        return self.wrap_output(self.__controller.step(self.wrap_step(action=action, rotation=params.get(self.ROTATION_KEY), horizon=params.get(self.HORIZON_KEY))))
```
ThomasSchellenbergNextCentury: @rartiss55 For the argument of the `wrap_step` function, we should find a way to avoid listing specific parameters like `rotation` and `horizon` here, because we'll be adding some more parameters in the future, and I don't think we want to list all of them out.  I believe (check me on this) you can do this with the following code:
```
wrap_step(action=action, **params)
```

ai2thor_wrapper/controller_ai2thor.py
```diff
@@ -43,10 +51,46 @@ def reset_scene(self, config_name, config_data):
...
+
+        if horizon > self.MAX_HORIZON or horizon < self.MIN_HORIZON:
+            print('Value of horizon needs to be between ' + str(self.MIN_HORIZON) + ' and ' + str(self.MAX_HORIZON) + '. Current value: ' + str(horizon))
+            isValid = False
+
+        return isValid
+
+    # Need a conversion step for what ai2thor will accept as input
+    # to keep parameters more simple for the user (in this case, wrapping
+    # rotation degrees into an object)
```
ThomasSchellenbergNextCentury: In Python, if you want to write a block comment (like for class or function headers), you can use three double quotes like this:
```python
"""
Block comment
"""
def my_function():
```

ai2thor_wrapper/controller_ai2thor.py
ThomasSchellenbergNextCentury: This here seems fine to me, but I'd be happy to have a longer discussion on the topic.  @brianpippin do you have any thoughts on the current implementation (settings the invalid parameters to a default value and showing a warning message)?

ai2thor_wrapper/run_mcs_environment.py
ThomasSchellenbergNextCentury: @rartiss55 So I believe what you're saying is that, if we want to let the performers specify degrees in "relative" terms (corresponding to their current camera view), we need to do a conversion?  And, the next question would be, should we do that conversion in our Python API or in Unity?  Hmm... I feel like this would be easier to do in Unity, since there we have direct access to the performer's current rotation.

ai2thor_wrapper/controller_ai2thor.py
brianpippin: I kind of error on the side of trying to keep as much out of the fork as possible, but i think the current implementation of setting to a default value and a warning message is a good.

ai2thor_wrapper/run_mcs_environment.py
```diff
@@ -15,10 +15,24 @@ def run_scene(controller, config_name, config_data):
     output = controller.reset_scene(config_name, config_data)
     print('step=' + str(output.step_number))
 
-    for i in range(1, 31):
-        output = controller.step('Pass')
+    #for i in range(1, 31):
+    #    output = controller.step('Pass')
+    #    print('step=' + str(output.step_number))
```
brianpippin: If we no longer need this commented out code we should probably remove it, unless you are saving it for something else later, in which case maybe add a comment about why it is being saved?

ai2thor_wrapper/run_mcs_environment.py
brianpippin: Yes it seems like that would be easier in Unity, it would seem tedious to have to call Unity to get the current rotation, increment it and send the new rotation back, instead of just sending the increment and having Unity adjust accordingly.

ThomasSchellenbergNextCentury: Technically, the output from AI2-THOR's `step` function returns the performer's current rotation, but it still seems like the conversion code should belong in Unity.  https://ai2thor.allenai.org/ithor/documentation/metadata/#agent-attributes-

ai2thor_wrapper/controller_ai2thor.py
rartiss55: Yup, makes sense, and that worked. Thanks!

ai2thor_wrapper/controller_ai2thor.py
rartiss55: Thanks for the heads up -- updated.

ai2thor_wrapper/run_mcs_environment.py
rartiss55: Ah good catch -- left a comment. Thought it might be good to keep the previous behavior around but commented out in case we wanted to quickly view a scene. I can take it out though if its not needed.

ai2thor_wrapper/run_mcs_environment.py
rartiss55: So I think moving this from the Python API to Unity code makes sense, but now that I'm fiddling with it I think this involves revisiting the validation logic/moving some of that to Unity as well. For example, if my current rotation is 355 degrees and I say "rotate 20 degrees more" then that absolute value would be over 360/invalid input for our current spec. But I wouldn't really know until the Unity code. 

We could just treat that case as valid input (and for Unity it would be -- Unity would accept values like 720 degrees for example) but then do we even need validation logic? I'll think about this some more tonight but wanted to leave this here in case you guys had any additional thoughts.

ThomasSchellenbergNextCentury: Hmm, I think for the case you mention, it shouldn't be an error; we should just set them to 15.  Same if you're at 5 and then go -20 (you should end at 345).  I think the "invalid input" warnings are for situations that we know are never valid, like tilting your head too high/low or setting the rotation to a string when it should be always number.  An open question is whether  rotating left/right by more than 360 degrees or less than 0 degrees should be an error.  We could easily `% 360` the rotation input, but I'm not sure if that would be more intuitive than showing a warning.

rartiss55: I updated this PR and added code in our ai2thor fork to handle the relative RotateLook functionality we want: https://github.com/NextCenturyCorporation/ai2thor/pull/2

@ThomasSchellenbergNextCentury I added some error handling on the Python API side of things for non-numbers and removed the error message for the rotation inputs discussed, but left it in for look angles (that still needs to be -90 to 90). I added a check for that on the ai2thor side of things as well.

PR #10, opened 2020-03-03
MCS-47/MCS-57 - Add return status to step output

rartiss55:
Added return status to step output and updated the tests. Also updated run_mcs_environment script to test this with move actions (can run it with wall_obstruct.json file included in this PR).

(Note that MCS-47 is the ticket to add the return status and MCS-57 is the subtask to make sure it all works for move actions). 

Associated ai2thor changes here: https://github.com/NextCenturyCorporation/ai2thor/pull/4


ai2thor_wrapper/mcs_return_status.py
```diff
@@ -13,5 +13,6 @@ class MCS_Return_Status(Enum):
     SUCCESSFUL = "SUCCESSFUL"
     SUCCESSFUL_WITH_INVALID_PARAMETERS = "SUCCESSFUL_WITH_INVALID_PARAMETERS"
     UNDEFINED = "UNDEFINED"
-    WRONG_POSE = "WRONG_POSE"
+    WRONG_POSE = "WRONG_POSE",
```
brianpippin: Not sure you need the comma there

rartiss55: Ah, you're right! Fixed.

PR #12, opened 2020-03-12
Mcs 12 human input mode

brianpippin:
Please Let me know anywhere you think I can improve the python code.


ai2thor_wrapper/mcs_action_api_desc.py
```diff
@@ -0,0 +1,23 @@
+from enum import Enum, unique
+
+@unique
+class MCS_Action_API_DESC(Enum):
+    CLOSE_OBJECT = "Close a nearby object. (objectId=string, amount=float)"
```
chrislong: I would suggest combining this with the action keys enum and also using lists & tuples so you don't have to parse the string description later. For example:

`     CLOSE_OBJECT = "Close a nearby object.", "1", [("objectId", str), ("amount", float)]`
Then you can deconstruct the pieces later but not parse any strings.

ai2thor_wrapper/run_mcs_human_input.py
```diff
@@ -0,0 +1,126 @@
...
+        print(*userInput)
+        
+        # Run commands that have no parameters
+        if len(userInput) < 2:
+            output = controller.step(userInput[0])
+            print('step=' + str(output.step_number))
+            return input_commands()
+        # Run commands that have parameters, parse params to send to step function
+        else:
+            # controller.step won't take params as string, need to recreate
```
chrislong: Rather than dealing with each parameter count function separately, you could gather them all in a list and use the `*` operator to pass them all at once, like `controller.step(*args)` or `controller.step(**kwargs)`

ai2thor_wrapper/mcs_action_api_desc.py
brianpippin: Some of that was there, so that the user knew which parameters to enter, but I assume you also want to use the last part of the tuple for formatting the string that gets sent to the step function?

chrislong: Oh, ok, I missed that. Yeah, you'd have to format the parameter list for the help. I think it's still cleaner than parsing strings yourself, but YMMV.

ai2thor_wrapper/run_mcs_human_input.py
```diff
@@ -0,0 +1,126 @@
...
+    print('Enter your command:')
+    userInput = input().split(',')
+
+    if(userInput[0] == 'exit'):
+        print("Exiting Human Input Mode")
+        return
+
+    try:
+        # Map shortcut keys to command value
+        print(len(userInput[0]))
```
ThomasSchellenbergNextCentury: I don't think you need this print here.

ai2thor_wrapper/run_mcs_human_input.py
```diff
@@ -0,0 +1,126 @@
...
+# Execute Input Commands until the user exits the system
+def input_commands(): 
+    print('Enter your command:')
+    userInput = input().split(',')
+
+    if(userInput[0] == 'exit'):
+        print("Exiting Human Input Mode")
+        return
+
+    try:
```
ThomasSchellenbergNextCentury: Having a `try` around this entire block catches all of the errors, including errors thrown in `controller.step`.  I think your intent is to catch errors from bad/unsupported input, so you may want to separate that logic from the logic of actually running the action.  Ideally, if we could parse the action and each parameter in a separate try/catch block, we could pinpoint which part of the input actually causes an error.

ai2thor_wrapper/run_mcs_human_input.py
```diff
@@ -0,0 +1,126 @@
...
+
+                    output = controller.step(userInput[0], param1key=param1value.strip(), param2key=param2value.strip(), param3key=param3value.strip(), param4key=param4value.strip())
+                    print('step=' + str(output.step_number))
+                    return input_commands()
+                else:
+                    output = controller.step(userInput[0], param1key=param1value.strip(), param2key=param2value.strip())
+                    print('step=' + str(output.step_number))
+                    return input_commands()
+            else:
+                output = controller.step(userInput[0], param1key=param1value.strip())
```
ThomasSchellenbergNextCentury: It looks like the `step` parameters are always set to `"param1key"` when you want to set it to the string that the variable `param1key` is holding.  You can do this by creating a dict:

```python
params = {}
params[param1key] = param1value.strip
output = controller.step(userInput[0], params)
```

ai2thor_wrapper/run_mcs_human_input.py
ThomasSchellenbergNextCentury: Or a dict, probably.

ai2thor_wrapper/run_mcs_human_input.py
```diff
@@ -0,0 +1,126 @@
...
+        # Run commands that have parameters, parse params to send to step function
+        else:
+            # controller.step won't take params as string, need to recreate
+            param1key, param1value = userInput[1].split('=')
+            param1key = param1key.strip()
+
+            # this could probably be done more easily, open to suggestions
+            if len(userInput) > 3:
+                param2key, param2value = userInput[2].split('=')
+                param2key = param2key.strip()
```
ThomasSchellenbergNextCentury: This looks like a great place to add a function that takes a dict of the existing params and returns the same dict updated with the new params.  Heck, if you're feeling especially ambitious, you could even make it recursive (keep going until you run out of userInput)!

ThomasSchellenbergNextCentury:
Hey Brian, overall this looks great, and I'm excited to use it a whole lot once we finish working out the kinks.  Using it, I had a couple of thoughts, but they all seem like low priority, nice-to-haves and can thus wait for now (we can make new JIRA tickets):

- At first, it's not clear how you're supposed to enter the parameters (Separated by commas, or spaces? Surrounded by parentheses? Surrounded by quotation marks?).  Maybe an example use of each action would be helpful.
- If you fat-finger an action, I think(?) it should just prompt you for input again, rather than trying to execute the invalid action and fail.
- It would be nice if the up arrow let you repeat your past commands.  Maybe we could keep a command history in the script?
- It would also be nice if CTRL-C let you automatically exit the script.
- We'll want to clean up the step output that is printed to the console, because it gets much longer with my recent addition of object metadata.  Maybe we'll just show a few things like the object IDs and return status.

Let me know what you think.

chrislong:
I would strongly suggest checking out the cmd module, which is part of the standard Python library (https://docs.python.org/3.6/library/cmd.html). When used with the readline library (also from the standard library), it allows command completion, history, and editing.

brianpippin:
Made a whole slew of changes that should hopefully improve performance and address some of the comments:

* Gathering all the params up in a dict, iterating over them and building the param list that way (I wasn't aware of that python convention, thanks for pointing it out, read about it last night, it is super useful)
* Tried to clean up where and how we are using the try catches
* Added try catch for mapping shortcut key
* Added try catch to make sure command is available before defaulting to pass
* Added try catch for parsing params
* CTRL+C should exit the program properly now.
* Added print command that will allow you to print commands
* Added some example commands

Things to still address:
* Cleaning up enums potentially
* Arrow keys or way to see past commands
* Clean up step output

ai2thor_wrapper/run_mcs_human_input.py
```diff
@@ -0,0 +1,135 @@
...
+        return input_commands()
+ 
+    # Run commands that have no parameters
+    if len(userInput) < 2:
+        output = controller.step(userInput[0])
+        print('step=' + str(output.step_number))
+        return input_commands()
+    else:
+        # Create Params List
+        print("Creating param list")
```
ThomasSchellenbergNextCentury: I think this is an extra print statement.

ai2thor_wrapper/run_mcs_human_input.py
```diff
@@ -33,11 +33,23 @@ def print_commands():
         print("ShortcutKey: " + commandListItem.key)
         print("*******************")
 
+    print("Example commands: ")
+    print("MoveAhead")
+    print("RotateLook, rotation=45, horizon=15")
     print(" ")
-    print("Enter 'exit' to exit the program")
+    print("Enter 'print' to print the commands again.")
```
chrislong: How about 'help', instead?

brianpippin: That's a good idea, updated.

PR #14, opened 2020-03-13
MCS-80 initial scene generator

chrislong:
Just makes the same static scene every time for now.

### DIFF
--- a/scene_generator/scene_generator.py
+++ b/scene_generator/scene_generator.py
@@ -9,7 +9,7 @@ import json
 
 DEFAULT_JSON = """
 {
-  "name": "file_name_prefix_uuid",
+  "name": "",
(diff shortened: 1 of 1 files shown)


scene_generator/scene_generator.py
```diff
@@ -0,0 +1,66 @@
...
+
+import sys
+import argparse
+import os
+import os.path
+import json
+
+DEFAULT_JSON = """
+{
+  "name": "file_name_prefix_uuid",
```
ThomasSchellenbergNextCentury: Maybe we should just set the name to an empty string rather than this silly dummy string I gave you?  (I hope that doesn't sound confrontational; just a nit-pick.)

chrislong: Sure, that's fine, I'll push a change

PR #16, opened 2020-03-16
Mcs 48 convert action params

brianpippin:


ai2thor_wrapper/mcs_controller_ai2thor.py
```diff
@@ -97,44 +126,109 @@ def start_scene(self, config_data):
...
+
+        objectDirectionX = kwargs.get(self.OBJECT_DIRECTION_X_KEY, self.DEFAULT_DIRECTION)
+        objectDirectionY = kwargs.get(self.OBJECT_DIRECTION_Y_KEY, self.DEFAULT_DIRECTION)
+        objectDirectionZ = kwargs.get(self.OBJECT_DIRECTION_Z_KEY, self.DEFAULT_DIRECTION)
+        receptacleObjectDirectionX = kwargs.get(self.RECEPTACLE_DIRECTION_X, self.DEFAULT_DIRECTION)
+        receptacleObjectDirectionY = kwargs.get(self.RECEPTACLE_DIRECTION_Y, self.DEFAULT_DIRECTION)
+        receptacleObjectDirectionZ = kwargs.get(self.RECEPTACLE_DIRECTION_Z, self.DEFAULT_DIRECTION)
+
+        # Check params that should be numbers
+        if self.is_number(rotation, self.ROTATION_KEY) == False:
```
ThomasSchellenbergNextCentury: FYI In Python you can just do:
```python
if not self.is_number(rotation, self.ROTATION_KEY):
```

ai2thor_wrapper/mcs_controller_ai2thor.py
```diff
@@ -97,44 +126,109 @@ def start_scene(self, config_data):
...
+            rotation = self.DEFAULT_ROTATION
+
+        if self.is_number(horizon, self.HORIZON_KEY) == False:
+            horizon = self.DEFAULT_HORIZON
+
+        if self.is_number(amount, self.AMOUNT_KEY) == False:
+            amount = self.DEFAULT_AMOUNT
```
ThomasSchellenbergNextCentury: So as I was playing around with the Open and Close actions a few minutes ago, I realized that we probably want the default "amount" to be 1 for both (either open fully or close fully).  I have a few different thoughts:

- Maybe it's just easier to use a differently named parameter for Open and Close, so it's not the same as the parameter for the Move actions?  The name doesn't matter to me, and it's easy enough to update the documentation.  Maybe "percent"?

- If we want to keep "amount", we could check to see if the action name is in the OBJECT_MOVE_ACTIONS, like:

```python
amount = (self.DEFAULT_OBJECT_MOVE_AMOUNT if action in self.OBJECT_MOVE_ACTIONS else self.DEFAULT_MOVE_AMOUNT)
```

ai2thor_wrapper/mcs_action_api_desc.py
```diff
@@ -2,22 +2,22 @@
 
 @unique
 class MCS_Action_API_DESC(Enum):
-    CLOSE_OBJECT = "Close a nearby object. (objectId=string, amount=float)"
+    CLOSE_OBJECT = "Close a nearby object. (objectId=string, amount=float, objectDirectionX=float, objectDirectionY=float, objectDirectionZ=float)"
```
ThomasSchellenbergNextCentury: Do we want to add the default values here too, or is denoting the defaults in the API page sufficient?

ai2thor_wrapper/mcs_controller_ai2thor.py
```diff
@@ -97,44 +126,109 @@ def start_scene(self, config_data):
...
             rotation=rotation_vector,
-            horizon=horizon
+            horizon=horizon,
+            moveMagnitude=moveMagnitude,
+            objectDirectionX=objectDirectionX,
+            objectDirectionY=objectDirectionY,
+            objectDirectionZ=objectDirectionZ,
+            receptacleObjectDirectionX=receptacleObjectDirectionX,
+            receptacleObjectDirectionY=receptacleObjectDirectionY,
+            receptacleObjectDirectionZ=receptacleObjectDirectionZ
```
ThomasSchellenbergNextCentury: Whoops, I totally forgot this when I wrote the ticket:  we need to send "objectDirection" and "receptacleObjectDirection" properties to Unity, not individual X, Y, and Z properties.  We can handle them like we do with "rotation" and "rotation_vector", where we convert the given X, Y, and Z properties into dicts with X, Y, and Z properties.  Sorry for the mixup.

ai2thor_wrapper/mcs_controller_ai2thor.py
brianpippin: I just had copied the convention there, but that would be a good way to clean it up, I can add that.  (Also good to know)

ai2thor_wrapper/mcs_controller_ai2thor.py
brianpippin: Isn't there already a check that would discern this?  I think I forgot to change the default to one like you mentioned:  

        if action in self.OBJECT_MOVE_ACTIONS:
            moveMagnitude = amount

        if action in self.MOVE_ACTIONS:
            moveMagnitude = amount * self.MAX_MOVE_DISTANCE

ai2thor_wrapper/mcs_action_api_desc.py
brianpippin: Yes we should add default values here, if we are going to be able to have the human input spit out the API as help, we should keep them in sync.

ai2thor_wrapper/mcs_controller_ai2thor.py
brianpippin: Okay that should be easy enough to fix up, I didn't realize it took them like rotation, but thinking about it, that makes more sense.

ai2thor_wrapper/mcs_controller_ai2thor.py
```diff
@@ -97,44 +126,109 @@ def start_scene(self, config_data):
...
-           horizon = 0
-
-        if horizon > self.MAX_HORIZON or horizon < self.MIN_HORIZON:
-            print('Value of horizon needs to be between ' + str(self.MIN_HORIZON) + \
-                ' and ' + str(self.MAX_HORIZON) + '. Current value: ' + str(horizon)+ \
-                '. Will be reset to 0.')
-            horizon = 0
+    def validate_and_convert_params(self, action, **kwargs):
+        moveMagnitude = self.MAX_MOVE_DISTANCE
+        rotation = kwargs.get(self.ROTATION_KEY, self.DEFAULT_HORIZON)
```
chrislong: Should this be DEFAULT_ROTATION?

ai2thor_wrapper/mcs_controller_ai2thor.py
ThomasSchellenbergNextCentury: Oh, yeah.  I guess I just figured if you were jumping through hoops to account for the different types of "amount" then we can just change the name of one of the "amounts" to something else like "percent".

ai2thor_wrapper/mcs_controller_ai2thor.py
brianpippin: It should.  They are the same value, and at one point I had them combined into the same default, but adding in more defaults I ended up splitting them up, I must have missed this one when cleaning up the code.

brianpippin:
Merged on command line and forgot to push

PR #18, opened 2020-03-17
Increasing baby force for force actions, briefly testing PushObject, …

rartiss55:
…PullObject (MCS-62, MCS-63).

ai2thor changes here: https://github.com/NextCenturyCorporation/ai2thor/pull/12

### DIFF
--- a/python_api/machine_common_sense/mcs_controller_ai2thor.py
+++ b/python_api/machine_common_sense/mcs_controller_ai2thor.py
@@ -31,8 +31,8 @@ class MCS_Controller_AI2THOR(MCS_Controller):
     MAX_MOVE_DISTANCE = 0.5
 
     # The amount of force to offset force values, that seems appropriate for a baby
-    # TODO Check with psych team about this about what we should use for a baby, defaulting to .25 now
-    MAX_BABY_FORCE = 0.25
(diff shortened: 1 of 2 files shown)


python_api/machine_common_sense/mcs_controller_ai2thor.py
```diff
@@ -31,8 +31,8 @@ class MCS_Controller_AI2THOR(MCS_Controller):
     MAX_MOVE_DISTANCE = 0.5
 
     # The amount of force to offset force values, that seems appropriate for a baby
-    # TODO Check with psych team about this about what we should use for a baby, defaulting to .25 now
-    MAX_BABY_FORCE = 0.25
+    # TODO Check with psych team about this about what we should use for a baby, defaulting to 25 now
+    MAX_BABY_FORCE = 25.0
```
rartiss55: Bumped this up since the object movement before was barely noticeable (I'm sure this will change again later).

ThomasSchellenbergNextCentury: So my intent was: the number that the TA1s give to us should be between 0 and 1, but the number that we give to AI2-THOR should be between 0 and X, where X is some reasonable "baby force" number.  `0.25` is definitely too low (sorry I didn't notice that before Brian), and `25` seems fine to me based on my limited prior testing.

PR #20, opened 2020-03-18
Mcs 27 Generate Scenes with Many Different Player Starting Locations

chrislong:
Implements MCS-27

### DIFF
--- a/scene_generator/scene_generator.py
+++ b/scene_generator/scene_generator.py
@@ -58,7 +58,7 @@ def generate_file(name):
     body['name'] = os.path.basename(name)
     position = body['performerStart']['position']
     position['x'] = random_position()
-    position['y'] = random_position()
+    position['z'] = random_position()
     body['performerStart']['rotation']['y'] = random_rotation()
 
     with open(name, 'w') as out:
--- a/scene_generator/scene_generator.py
+++ b/scene_generator/scene_generator.py
@@ -6,8 +6,10 @@ import argparse
 import os
 import os.path
 import json
+import copy
+import random
 
-DEFAULT_JSON = """
+OUTPUT_TEMPLATE_JSON = """
 {
   "name": "file_name_prefix_uuid",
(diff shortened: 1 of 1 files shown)


scene_generator/scene_generator.py
```diff
@@ -28,38 +30,69 @@
...
 
 def generate_file(name):
-    global DEFAULT_JSON
-    body = json.loads(DEFAULT_JSON)
+    global OUTPUT_TEMPLATE
+    body = copy.deepcopy(OUTPUT_TEMPLATE)
     body['name'] = os.path.basename(name)
+    position = body['performerStart']['position']
+    position['x'] = random_position()
+    position['y'] = random_position()
```
wagnerwj: This should be position['z']

scene_generator/scene_generator.py
```diff
@@ -28,38 +30,69 @@
...
+    body['performerStart']['rotation']['y'] = random_rotation()
```
wagnerwj: shouldn't this me position['rotation']['y']?

scene_generator/scene_generator.py
chrislong: Oops, yes, it sure should. I'll push a fix. Thanks.

scene_generator/scene_generator.py
chrislong: I don't think so. From what I've seen, I'm pretty sure `rotation` is part of `performerStart`, not `position`.

ThomasSchellenbergNextCentury: Chris, you are correct.  In Unity, `position`, `rotation`, and `scale` are three separate objects (all under an object's `transform` property).

ThomasSchellenbergNextCentury: You don't have to worry about `transform` here though; I just mentioned it to illustrate that Unity manages `position` and `rotation` separately.

PR #25, opened 2020-03-20
Mcs 28

wagnerwj:
Added stuff for object inclusion=- and a basic collision detection function


scene_generator/scene_generator.py
```diff
@@ -51,21 +57,105 @@ def random_rotation():
...
+
+def dot_prod_dict(v1, v2):
+    return sum (v1[key]*v2.get(key,0) for key in v1)
+
+def collision(test_rect, test_point):
+    #assuming test_rect is an array4 points in order... Clockwise or CCW does not matter
+    #points are {x,y,z}
+    A=test_rect[0]
+    B=test_rect[1]
+    C=test_rect[2]
```
chrislong: I'm confused. If it's a rectangle, shouldn't two points be sufficient to define it? Or is it really a quadrilateral?

scene_generator/scene_generator.py
```diff
@@ -51,21 +57,105 @@ def random_rotation():
...
+    
+    vectorAB={'x': B['x']-A['x'], 'y': B['y']-A['y'], 'z': B['z']-A['z']}
+    vectorBC={'x': C['x']-B['x'], 'y': C['y']-B['y'], 'z': C['z']-B['z']}
+    
+    vectorAM={'x': test_point['x']-A['x'], 'y': test_point['y']-A['y'], 'z': test_point['z']-A['z']}
+    vectorBM={'x': test_point['x']-B['x'], 'y': test_point['y']-B['y'], 'z': test_point['z']-B['z']}
+    
+    return (0<=dot_prod_dict(vectorAB, vectorAM)<=dot_prod_dict(vectorAB, vectorAB)) & (0<=dot_prod_dict(vectorBC, vectorBM)<=dot_prod_dict(vectorBC, vectorBC))
```
chrislong: How about a citation for where this comes from?

scene_generator/scene_generator.py
```diff
@@ -51,21 +57,105 @@ def random_rotation():
...
+        radian_amount = rotation_amount*math.pi/180.0
+        new_x = random_position()
+        new_z = random_position()
+        rotate_sin = math.sin(radian_amount)
+        rotate_cos = math.cos(radian_amount)
+        a = { 'x': new_x+(dx*rotate_cos)-(dz*rotate_sin) , 'y' : 0 , 'z': new_z+(dx*rotate_sin+dz*rotate_cos)}
+        b = { 'x': new_x+(dx*rotate_cos)-(dz*rotate_sin) , 'y' : 0 , 'z': new_z-(dx*rotate_sin+dz*rotate_cos)}
+        c = { 'x': new_x-(dx*rotate_cos)+(dz*rotate_sin) , 'y' : 0 , 'z': new_z-(dx*rotate_sin+dz*rotate_cos)}
+        d = { 'x': new_x-(dx*rotate_cos)+(dz*rotate_sin) , 'y' : 0 , 'z': new_z+(dx*rotate_sin+dz*rotate_cos)} 
+        rect = []
```
chrislong: you could also do `rect = [ a, b, c, d ]`

scene_generator/scene_generator.py
```diff
@@ -88,11 +178,13 @@ def main(argv):
     parser.add_argument('--prefix', required=True, help='Prefix for output filenames')
     parser.add_argument('-c', '--count', type=int, default=1, help='How many scenes to generate [default=1]')
     parser.add_argument('--seed', type=int, default=None, help='Random number seed [default=None]')
+    parser.add_argument('--objects', required=False, help='File containing a list of objects to insert')
```
chrislong: Shouldn't objects be required? Won't the program die if it's not provided?

scene_generator/scene_generator.py
```diff
@@ -8,13 +8,17 @@
 import json
 import copy
 import random
+import uuid
+import math
+from _hashlib import new
```
chrislong: Looks like this isn't used anywhere

scene_generator/scene_generator.py
wagnerwj: Actually, I had originally planned on leaving a default value- but forgot- so good catch

scene_generator/scene_generator.py
```diff
@@ -51,21 +57,104 @@ def random_rotation():
         rotation = int(rotation)
     return rotation
 
+def get_object(object_file_name):
+    with open(object_file_name) as object_file:
+        objects = json.load(object_file)
```
ThomasSchellenbergNextCentury: We should probably just load this fine once, so we aren't opening and reading it for each individual object we want to add.

scene_generator/scene_generator.py
```diff
@@ -51,21 +57,104 @@ def random_rotation():
...
+def calc_obj_pos(performer_position, new_object, old_object):
+    #generate position and rotation
+    #find the corners
+    #calculate intercept of points
+
+    dx = old_object['dimensions']['x']
+    dz = old_object['dimensions']['z']
+    
+    
+    while True:
```
ThomasSchellenbergNextCentury: I can imagine some rare-yet-possible scenarios in which it would be impossible to add the new object (there isn't enough open space for it anywhere) and this would loop indefinitely.  We should add a check for this case... and select a replacement object in `generate_file`, I guess?

I'm not sure exactly what this check should be... Maybe just try X times, and then give up?

wagnerwj: Take a look at these now, T

scene_generator/scene_generator.py
brianpippin: +1

scene_generator/scene_generator.py
ThomasSchellenbergNextCentury: Looks fine

PR #27, opened 2020-03-23
Mcs 30

wagnerwj:
MCS-30: adding Salient Materials
https://nextcentury.atlassian.net/browse/MCS-30
I reserve the right to change things :)

### DIFF
--- a/scene_generator/materials.py
+++ b/scene_generator/materials.py
@@ -1,0 +1,68 @@
+CERAMIC_MATERIALS = [
+    "AI2-THOR/Materials/Ceramics/ConcreteBoards1",
+    "AI2-THOR/Materials/Ceramics/GREYGRANITE",
+    "AI2-THOR/Materials/Ceramics/KitchenFloor",
+    "AI2-THOR/Materials/Ceramics/RedBrick",
+    "AI2-THOR/Materials/Ceramics/TexturesCom_BrickRound0044_1_seamless_S",
+    "AI2-THOR/Materials/Ceramics/WhiteCountertop"
+]
+FABRIC_MATERIALS = [
+    "AI2-THOR/Materials/Fabrics/Carpet2",
+    "AI2-THOR/Materials/Fabrics/Carpet4",
+    "AI2-THOR/Materials/Fabrics/CarpetDark",
+    "AI2-THOR/Materials/Fabrics/CarpetWhite 3",
+    "AI2-THOR/Materials/Fabrics/HotelCarpet3",
+    "AI2-THOR/Materials/Fabrics/RugPattern224"
+]
+METAL_MATERIALS = [
+    "AI2-THOR/Materials/Metals/Brass 1",
+    "AI2-THOR/Materials/Metals/BrownMetal 1",
+    "AI2-THOR/Materials/Metals/GenericStainlessSteel",
+    "AI2-THOR/Materials/Metals/Metal"
+]
+PLASTIC_MATERIALS = [
+    "AI2-THOR/Materials/Plastics/GreenPlastic",
+    "AI2-THOR/Materials/Plastics/OrangePlastic",
+    "AI2-THOR/Materials/Plastics/YellowPlastic2"
+]
+RUBBER_MATERIALS = [
+    "AI2-THOR/Materials/Plastics/BlueRubber"
+]
+WOOD_MATERIALS = [
+    "AI2-THOR/Materials/Wood/BedroomFloor1",
+    "AI2-THOR/Materials/Wood/LightWoodCounters 1",
+    "AI2-THOR/Materials/Wood/LightWoodCounters4",
+    "AI2-THOR/Materials/Wood/TexturesCom_WoodFine0050_1_seamless_S",
+    "AI2-THOR/Materials/Wood/WoodFloorsCross",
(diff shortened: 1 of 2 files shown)


scene_generator/scene_generator.py
```diff
@@ -83,6 +85,8 @@ def collision(test_rect, test_point):
     
     return (0<=dot_prod_dict(vectorAB, vectorAM)<=dot_prod_dict(vectorAB, vectorAB)) & (0<=dot_prod_dict(vectorBC, vectorBM)<=dot_prod_dict(vectorBC, vectorBC))
 
+#Returns true if we managed to put the object in the frame
+#False if we couldnt
 def calc_obj_pos(performer_position, new_object, old_object):
     #generate position and rotation
```
chrislong: How about changing these comments to a python docstring?

scene_generator/scene_generator.py
```diff
@@ -112,46 +116,56 @@ def calc_obj_pos(performer_position, new_object, old_object):
         if  not collision(rect,performer_position):
             break;
         tries += 1
-        
-    rotation = { 'x' : 0, 'y': rotation_amount, 'z': 0 }
+     
+    if tries < 6 :
```
chrislong: It would be better if this 6 were a constant declared at the top of the file, rather than a magic embedded literal.

scene_generator/scene_generator.py
```diff
@@ -112,46 +116,56 @@ def calc_obj_pos(performer_position, new_object, old_object):
...
+        rotation = { 'x' : 0, 'y': rotation_amount, 'z': 0 }
+        position = { 'x' : new_x, 'y': old_object['position_y'], 'z' : new_z}
+        new_object['rotation'] = rotation
+        new_object['position'] = position
```
chrislong: To me these intermediate variables (rotation and position) don't contribute much, but YMMV, I guess

scene_generator/scene_generator.py
```diff
@@ -112,46 +116,56 @@ def calc_obj_pos(performer_position, new_object, old_object):
...
+        for attribute in selected_object['attributes']:
+            new_object[attribute]= True
     
-    shows_object['scale'] = selected_object['scale']
-    calc_obj_pos(position, shows_object, selected_object)
-
-    body['objects'].append(new_object)
+        shows = []
+        new_object['shows'] = shows;
+        shows.append(shows_object)
```
chrislong: or combine 156 and 158 into `shows = [shows_object]`

scene_generator/scene_generator.py
```diff
@@ -112,46 +116,56 @@ def calc_obj_pos(performer_position, new_object, old_object):
...
+        shows_object['stepBegin'] = 0;
+        shows_object['scale'] = selected_object['scale']
+        if 'salientMaterials' in selected_object:
+            salientMaterialsIndex = selected_object['salientMaterials'][0].upper()+'_MATERIALS'
+            salientMaterial = random.choice(globals().get(salientMaterialsIndex, None))
```
chrislong: Based on this, maybe put a comment in the materials.py file to indicate the names in there aren't aribtrary, but have to correspond to the values "salientMaterials" can take on.

scene_generator/scene_generator.py
```diff
@@ -12,7 +12,9 @@
 import math
 
 
+from materials import *
 
+#
```
chrislong: why?

scene_generator/scene_generator.py
wagnerwj: Part of the requirement from MCS-28

chrislong: Sorry, I was too terse. I mean, I would combine the assignments and eliminate the intermediate variables, like
`new_object['rotation'] = { 'x' : 0, 'y': rotation_amount, 'z': 0 }`
and similarly for position.

scene_generator/materials.py
```diff
@@ -1,3 +1,4 @@
+#Holding the bunches of materials- will probably be changed by whatever Thomas has planned
```
chrislong: Sorry if this is pedantic, but I didn't mean just any comment, I meant a comment that indicates that the variable names in this file (e.g., `CERAMIC_MATERIALS`) have to be named according to valid values for the salientMaterials property of objects. Otherwise the magic is too obscured, and it's a future bug waiting to happen when someone renames these. (I am esp. concerned because there may be a lot of churn on MCS and other developers are likely to need to know about this.) It would also be really great to have a pointer to where the list of valid salientMaterials is.

brianpippin: I agree with Chris that as much you can put in the comments that will help the next developers to look at this the better, especially as MCS evolves.

scene_generator/scene_generator.py
```diff
@@ -112,46 +114,53 @@ def calc_obj_pos(performer_position, new_object, old_object):
         if  not collision(rect,performer_position):
             break;
         tries += 1
-        
-    rotation = { 'x' : 0, 'y': rotation_amount, 'z': 0 }
+     
+    if tries < 6 :
```
brianpippin: Should this 6 be set to MAX_TRIES as well, because I assume if someone changed the MAX_TRIES value up top, this would break here right?

PR #28, opened 2020-03-24
Mcs 29

wagnerwj:
Multiple objects in scene

Can't remember who else Brian wanted review options

### DIFF
--- a/scene_generator/scene_generator.py
+++ b/scene_generator/scene_generator.py
@@ -48,7 +48,7 @@ MIN_ROTATION = 0
 MAX_ROTATION = 359
 ROTATION_DIGITS = 0
 MAX_TRIES = 6
-
+MAX_OBJECTS = 5
 
 def random_position():
     return round(random.uniform(MIN_PERFORMER_POSITION, MAX_PERFORMER_POSITION), POSITION_DIGITS)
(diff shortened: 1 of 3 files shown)


scene_generator/scene_generator.py
```diff
@@ -86,38 +84,39 @@ def collision(test_rect, test_point):
...
-        b = { 'x': new_x+(dx*rotate_cos)-(dz*rotate_sin) , 'y' : 0 , 'z': new_z-(dx*rotate_sin+dz*rotate_cos)}
-        c = { 'x': new_x-(dx*rotate_cos)+(dz*rotate_sin) , 'y' : 0 , 'z': new_z-(dx*rotate_sin+dz*rotate_cos)}
-        d = { 'x': new_x-(dx*rotate_cos)+(dz*rotate_sin) , 'y' : 0 , 'z': new_z+(dx*rotate_sin+dz*rotate_cos)} 
-        rect = [a, b, c, d]
-
         
-        if  not collision(rect,performer_position):
-            break;
+        rect = calc_obj_coords(new_x, new_z, dx, dz, rotation)
+        if not any(collision(rect, point) for point in other_points):
```
chrislong: Shouldn't this be calculating collisions of rectangles with rectangles? Just testing rectangles against corners could allow one rect inside another, for example, right?

wagnerwj: Its testing if the rectangle point is in another rectangle- therefore a collision..

However, thinking about  it, there may be a problem unless I check both ways; I.E. the case where my rect rectangle is inside the rectangle with points....
Making the check both ways should alleviate this concern.

chrislong: That would help, but it won't be sufficient. Consider two axis-aligned rectangles, one wide and skinny with corners (-10, -1) and (10, 1) and another tall sknny one (-1, 10), (1, -10). No corners of either are in the other, but the rectangles intersect around the origin. I think you need an actual rectangle intersection algorithm, or maybe a general polygon one since they may not be axis-aligned. (I'm just not sure if rectangle-specific algorithms exist if they're not axis aligned.)

wagnerwj: Ah! Good counter!
And yes, off axis ones tend to be hard-to-find. Let me do a little research- because this is a non-trivial problem.

wagnerwj: Oh Future question to @ThomasSchellenbergNextCentury 
Will we need to worry about 3D intersection here? 
To clarify, right now we're just checking for intersections in the X-Z plane, will we need to also worry about Y plane intersections?

scene_generator/separating_axis_theorem.py
```diff
@@ -0,0 +1,89 @@
...
+# of the Hyperplane separation theorem, also known as Separating axis theorem (SAT).
+#
+# For more information visit:
+# https://en.wikipedia.org/wiki/Hyperplane_separation_theorem
+#
+# Copyright (C) 2016, Juan Antonio Aldea Armenteros
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
```
chrislong: This will force us to release the scene generator with the GPL 3+ license, too, if we distribute the scene generator.

ThomasSchellenbergNextCentury: Oh, will that contaminate our whole repo?  :/

wagnerwj: Not the repo-just this code..

chrislong: Yeah, just the scene generator, because it's the program that's calling the GPL'ed code.

wagnerwj: TBH- after further testing, this alg doesn't work if I only have a single point of overlap ( which I consider an intersection) putting something else together- line segment intersection.

PR #29, opened 2020-03-26
Mcs 33 - Generate Scenes with Simple Interaction Goals

chrislong:
Implements goal generation


wagnerwj:
I don't see an issue- but adding other reviewers

brianpippin:
Nothing jumped out at me, looked solid, will see if @ThomasSchellenbergNextCentury has any.

scene_generator/goals.py
```diff
@@ -0,0 +1,83 @@
...
+    }
+    return goal
+
+
+def _generate_transportation_goal(objects):
+    if len(objects) < 2:
+        raise ValueError(f'need at least 2 for this goal, was only given {len(objects)}')
+    pickupables = [x for x in objects if 'pickupable' in x['attributes']]
+    if len(pickupables) == 0:
+        raise ValueError('at least one object must be "pickupable"')
```
ThomasSchellenbergNextCentury: Are we catching these errors anywhere?

scene_generator/scene_generator.py
```diff
@@ -164,12 +166,15 @@ def generate_file(name, objects):
                 new_object['salientMaterials'] = selected_object['salientMaterials']
             body['objects'].append(new_object)
         
+    if add_goal:
+        body['goal'] = generate_goal(body['objects'])
+
```
ThomasSchellenbergNextCentury: @chrislong Not to complicate things, but I was just thinking about this task some more, and I had an idea:  instead of generating the objects first and then the goal, why don't we reverse it?  We decide which goal type we want first, then choose which objects we want based on the chosen goal type (so we make sure we have at least X objects and they each have specific attributes if needed), then fill out the goal metadata with the chosen objects.  Does that make sense, and, if so, would that be hard to change now?

scene_generator/goals.py
chrislong: I was thinking it would be developers (us) using this, so if they bubbled up and halted the program, that would be ok. (And that halting is probably the desired behavior regardless.) If you think it should produce nicer error messages, I could add a catch to the main program.

scene_generator/scene_generator.py
chrislong: That could work. If later on, goals have more specific requirements than these, it might be a much better way to do it, actually. I can refactor it to do it that way.

ThomasSchellenbergNextCentury: That would be nice, yes.

scene_generator/goals.py
ThomasSchellenbergNextCentury: So, in the long term, we'll be running this in batch, and, right now, it's very possible for the script to randomly choose number_of_objects=1 and goal=transportation and then fail.  This specific case just seems very likely given how the script is currently coded.  (Unless I'm missing something?)

However, this doesn't really matter if you implement my other suggestion (choosing the goal before choosing the objects).

chrislong: Yeah, this shouldn't matter with the new strategy of picking the goal before the objects.

scene_generator/goals.py
```diff
@@ -0,0 +1,144 @@
...
+    def find_all_valid_objects_all_constraints(self, objects):
+        """For each constraint, find all valid members of objects."""
+        valid_objects = []
+        for constraint_list in self.get_object_constraint_lists():
+            objects = self.find_all_valid_objects(constraint_list, objects)
+            valid_objects.append(objects)
+        return valid_objects
+
+
+class IdGoal(Goal):
```
ThomasSchellenbergNextCentury, line 78: FYI The specific names of these goals have evolved over time, but I'll be making a new ticket to implement a few minor tweaks like name changes (I'm sure you just want to get this code approved and merged by now!).

scene_generator/goals.py
```diff
@@ -0,0 +1,144 @@
...
+        NEXT_TO = 'next to'
+        ON_TOP_OF = 'on top of'
+
+    TEMPLATE = {
+        'type_list': ['interaction', 'identification', 'objects', 'places'],
+        'task_list': ['navigation', 'identification', 'transportation']
+    }
+
+    def get_object_constraint_lists(self):
+        return [[AttributeConstraint(list.__contains__, 'attributes', 'pickupable')], []]
```
ThomasSchellenbergNextCentury, line 112: Fancy!

scene_generator/scene_generator.py
```diff
@@ -51,125 +50,150 @@
...
     body['performerStart']['rotation']['y'] = random_rotation()
-    
-    other_rects = [ ]
-    object_count = random.randint(1,MAX_OBJECTS)
-    for x in range (0, object_count):
-        selected_object = copy.deepcopy(random.choice(objects))
+
+    if add_goal:
+        goal_obj = goals.generate_goal()
+        min_obj_count = len(goal_obj.get_object_constraint_lists())
```
ThomasSchellenbergNextCentury, line 149: Could we save the object constraint list in a variable so we don't have to call the function again on line 159?

scene_generator/goals.py
```diff
@@ -0,0 +1,144 @@
...
+            obj_ok = True
+            for constraint in constraint_list:
+                if not constraint.is_true(obj):
+                    obj_ok = False
+                    break
+            if obj_ok:
+                valid_objects.append(obj)
+        return valid_objects
+
+    def find_all_valid_objects_all_constraints(self, objects):
```
ThomasSchellenbergNextCentury, line 69: Is this function not used anywhere?

scene_generator/scene_generator.py
```diff
@@ -51,125 +50,150 @@
...
+    other_rects = []
+    all_objects = []
+    object_count = random.randint(min_obj_count, MAX_OBJECTS)
+    for x in range(object_count):
+        valid_objects = objects
+        if add_goal:
+            constraint_lists = goal_obj.get_object_constraint_lists()
+            if x < len(constraint_lists):
+                constraint_list = constraint_lists[x]
+                valid_objects = goal_obj.find_all_valid_objects(constraint_list, objects)
```
ThomasSchellenbergNextCentury, line 162: So, we can push this to a future ticket, but I was thinking about how we'll eventually want to run this in batch jobs, and we could probably improve the performance here by pre-finding all this different lists of "valid objects" that we'll need (like the list of "pickupable" objects), so we only have to do it once.

scene_generator/goals.py
```diff
@@ -0,0 +1,144 @@
...
+    """An abstract Goal."""
+
+    def get_object_constraint_lists(self):
+        return []
+
+    def get_goal(self, objects):
+        return {}
+
+    @staticmethod
+    def find_valid_object(constraint_list, objects):
```
ThomasSchellenbergNextCentury, line 31: Are the find_valid_object, find_valid_objects, or final_all_valid_objects_all_constraints functions used anywhere?

scene_generator/goals.py
```diff
@@ -0,0 +1,144 @@
...
+        return self.attribute in obj and self.predicate(obj[self.attribute], *self.arguments)
+
+
+class Goal:
+    """An abstract Goal."""
+
+    def get_object_constraint_lists(self):
+        return []
+
+    def get_goal(self, objects):
```
ThomasSchellenbergNextCentury, line 27: This is a nitpick, but Goal.get_goal seems like it could be more descriptive.  Maybe something like get_config instead?

scene_generator/goals.py
```diff
@@ -0,0 +1,144 @@
...
+        goal['metadata'] = {
+            'target_1': {
+                'id': target1['id'],
+                'info': target1['info']
+            },
+            'target_2': {
+                'id': target2['id'],
+                'info': target2['info']
+            },
+            'relationship': [relationship.value, 'target_1', 'target_2']
```
ThomasSchellenbergNextCentury, line 134: It seemed like the popular opinion is to change the order of this array so the relationship's value is in the middle.

scene_generator/scene_generator.py
```diff
@@ -51,125 +50,150 @@
...
     with open(object_file_name) as object_file:
         objects = json.load(object_file)
-    return objects    
-        
+    return objects
+
+
 def dot_prod_dict(v1, v2):
-    return sum (v1[key]*v2.get(key,0) for key in v1)
+    return sum(v1[key] * v2.get(key, 0) for key in v1)
```
ThomasSchellenbergNextCentury, line 72: FYI Other reviewers: I'd recommend checking "Hide whitespace changes" in the GitHub diff settings at the top.

scene_generator/scene_generator.py
chrislong: Yeah, sure.

scene_generator/goals.py
deanwetherby: Should Goal inherit from ABC (abstract base class) and then decorate class methods with @abstractmethod?
from abc import ABC
class Goal(ABC):
    @abstractmethod
    def get_object_constraint_lists(self):

scene_generator/scene_generator.py
chrislong: I'm not sure what optimization you're thinking of here. The constraints for each goal "parameter" (each `constraint_list`) is only searched once. I guess we could do memoization or something where we remember the results of any individual constraint search and can use that again if/when it reappears in another `constraint_list`.
Anyway, doing it on a future ticket sounds good. :)

scene_generator/goals.py
chrislong: No. I'll remove them.

scene_generator/goals.py
chrislong: Ok, sure

scene_generator/goals.py
chrislong: I bow to the will of the people

scene_generator/scene_generator.py
ThomasSchellenbergNextCentury: Yeah, either memoization or just pre-computing.  If we called `generate_file` 10,000 times and had 5,000 TransportationGoals, then we'd be calling `find_all_valid_objects` 5,000 times, and it would return the same list of `valid_objects` each time.

scene_generator/goals.py
chrislong: That's a good idea, Dean, I didn't know about that package.

PR #31, opened 2020-03-27
Minor changes to human input mode

ThomasSchellenbergNextCentury:
Changes to human input mode script:
- Added support for the `action_list` returned in the step output that restricts the available actions for the next step.
- Condensed the "help" print statements so they take up less space in the terminal.

Brian, please let me know what you think.

### DIFF
--- a/python_api/machine_common_sense/run_mcs_human_input.py
+++ b/python_api/machine_common_sense/run_mcs_human_input.py
@@ -17,33 +17,48 @@ class command:
         self.desc = desc
 
 # Define all the possible human input commands
-def build_commands(): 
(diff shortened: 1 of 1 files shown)


brianpippin:
Thomas.  I struggled with putting all the commands on one line versus separate lines.  It's a weird console design choice, because the one line it looks a little busy and multiple lines it is clean but also really long and hard to necessarily see what you might need.  I think the parens help though.

ThomasSchellenbergNextCentury:
@brianpippin I agree that there are pros and cons to each.  Thanks for taking a look.

PR #32, opened 2020-03-27
Generating walls

wagnerwj:


wagnerwj:
Hold off on reviewing this until MCS-33 has been merged into master. I think I'm going to have to do a touch of refactoring then

chrislong:
Maybe it would be better to merge master into this branch and resolve conflicts here?

wagnerwj:
> Maybe it would be better to merge master into this branch and resolve conflicts here?

I plan on doing that- but I want your merges into master first. Which is why I requested a delay

scene_generator/scene_generator.py
```diff
@@ -50,6 +51,14 @@
 ROTATION_DIGITS = 0
 MAX_TRIES = 6
 MAX_OBJECTS = 5
+MAX_WALLS = 3
+MIN_WALLS = 0
```
brianpippin: I don't know if this will or won't come up, but shouldn't a room always have 4 walls?

chrislong:
Oh, right, MCS-33. Sorry, failure to read.

scene_generator/scene_generator.py
wagnerwj: These walls aren't the boundaries of the room- but additional obstacles inside it

wagnerwj: Plus- what have you got against triangular rooms????? 😄

ThomasSchellenbergNextCentury: Yeah @brianpippin the idea of this ticket was to have additional walls within the existing room  (for a total of 5, 6, or 7) .

brianpippin: That makes sense, should the code have a comment about that somewhere?  so if someone is just looking at the code they understand that?  

(Note - I heard triangular rooms are coming back in style)

wagnerwj: Good idea- added a comment to that effect

wagnerwj:
@ThomasSchellenbergNextCentury @rartiss55 @brianpippin  Why don't we go ahead and review this? Its been waiting long enough

ThomasSchellenbergNextCentury:
Sorry Billy, I thought we were waiting to merge Chris's PR first.  I'll take a look at it now.

scene_generator/scene_generator.py
```diff
@@ -163,7 +210,12 @@ def generate_file(name, objects):
                 new_object['info'].append(selected_object['salientMaterials'][0])
                 new_object['salientMaterials'] = selected_object['salientMaterials']
             body['objects'].append(new_object)
-        
+            
+            
+    wall_count = random.randint(MIN_WALLS,MAX_WALLS)
+    for x in range (0,wall_count):
+        generate_wall(ceil_wall_mat_choice, position, other_rects, body['objects'])
```
ThomasSchellenbergNextCentury, line 217: I just wanted to say that I think it's good that we're adding the walls AFTER we've added the objects.  I'd rather fail adding a wall due to repeated collisions than to fail adding an object.

scene_generator/scene_generator.py
```diff
@@ -163,7 +210,12 @@ def generate_file(name, objects):
                 new_object['info'].append(selected_object['salientMaterials'][0])
                 new_object['salientMaterials'] = selected_object['salientMaterials']
             body['objects'].append(new_object)
-        
+            
+            
+    wall_count = random.randint(MIN_WALLS,MAX_WALLS)
```
ThomasSchellenbergNextCentury, line 215: Now that I'm thinking about this, I'd prefer if the wall count probability was calculated more like this:
- 0 walls: 60%
- 1 wall: 20%
- 2 walls: 10%
- 3 walls: 10%

Would that be a hard change?

wagnerwj: No- I'll make something in a config though, so we can change as we see fit.

wagnerwj: Ok- done

wagnerwj:
@chrislong Can you review this? I was seeing errors with Goal Generation using this code.

chrislong:
@wagnerwj Sure, I'll take a look

brianpippin:
Good for me as well unless Chris has anything.

PR #33, opened 2020-03-31
Add Reset to Human Input Mode

ThomasSchellenbergNextCentury:
Done:
- Added reset command to human input mode script.
- Fixed bug with script's optional command line arguments.

Also, should we continue doing PRs for small changes like this, or should we just push them directly into master?  I'm fine either way, I just don't know if you'd think small reviews like this are a "waste of time".


deanwetherby:
It's hard to say on these little changes but it does bring up a question that I've had. Shouldn't we be using argparse, click, or something along those lines for CLI parameter parsing?

ThomasSchellenbergNextCentury:
@deanwetherby Yes, probably. I think that script has grown more complex than we originally expected.  I'll add a ticket to JIRA.

PR #40, opened 2020-04-10
Mcs 137 Refactor Scene Generator to be Driven by Goal

chrislong:


brianpippin:
Did the goals_test and scene_generator_test not to be updated with new refactor?   Should we add another ticket to add more tests to that code?

chrislong:
Oh, drat. I forgot about the tests. I doubt I'll have time to update them today, but should some time Monday. If you all want to include them in this ticket, that's fine, or if you want to go ahead and merge this so other stuff can use it, that's fine with me, too.

scene_generator/goals.py
```diff
@@ -112,7 +248,8 @@ def get_config(self, objects):
             },
             'relationship': ['target_1', relationship.value, 'target_2']
         }
-        goal['description'] = f'Find and pick up the {" ".join(target1["info"])} and move it {relationship.value} the {" ".join(target2["info"])}.'
+        goal[
+            'description'] = f'Find and pick up the {" ".join(target1["info"])} and move it {relationship.value} the {" ".join(target2["info"])}.'
```
ThomasSchellenbergNextCentury, line 252: I think your auto-code-formatter did something silly here.

scene_generator/goals.py
```diff
@@ -46,18 +125,53 @@ def find_all_valid_objects(constraint_list, objects):
         return valid_objects
 
 
+class EmptyGoal(Goal):
```
ThomasSchellenbergNextCentury, line 128: Is there a specific reason why you have an abstract goal class, and then an empty goal class, rather than just using the empty goal as the superclass for the other goals?  Both seem fine, and I have no preference either way -- just curious.

scene_generator/goals.py
chrislong: So it did. I'll fix it.

scene_generator/goals.py
chrislong: I think maybe it's a little cleaner this way, but it's mainly historical: I didn't need EmptyGoal when I first made the class hierarchy and I already had the model of Goal as abstract in my head and didn't about changing it.

chrislong:
I'll work on the tests today and also resolve conflicts

ThomasSchellenbergNextCentury:
@chrislong Would you mind handling the merge conflict here?

chrislong:
@ThomasSchellenbergNextCentury Not at all, I was planning to resolve it

PR #46, opened 2020-04-15
Updating compute_objects to match new approach

deanwetherby:
TraversalGoal didn't match the abstract methods from the base class so no traversal files were being generated.


deanwetherby:
Merged master back into this branch. It should be consistent now.

PR #47, opened 2020-04-16
Mcs 84: Generate Scenes with Pickupable Objects Inside Closed Receptacles

chrislong:
Does not do anything for optimal path because that code isn't in master yet.


scene_generator/geometry.py
```diff
@@ -83,4 +102,23 @@ def calc_obj_pos(performer_position, other_rects, old_object):
         other_rects.append(rect)
         return new_object
 
+    logging.debug(f'could not place object: {old_object}')
+    return None
+
+
+def can_enclose(objectA, objectB):
+    """Return True iff each 'dimensions' of objectA is >= the corresponding dimension of objectB."""
```
ThomasSchellenbergNextCentury, line 110: Wow, is this "iff" as in "if and only if"?  I haven't seen that since college...

scene_generator/goals.py
```diff
@@ -90,6 +97,28 @@ def instantiate_object(object_def, object_location):
...
+        if area_index is not None:
+            # try to place the container before we accept it
+            container_location = geometry.calc_obj_pos(performer_position, bounding_rects, container_def)
+            if container_location is not None:
+                found_container = instantiate_object(container_def, container_location)
+                found_area = container_def['enclosed_areas'][area_index]
+                all_objects.append(found_container)
+                target['locationParent'] = found_container['id']
+                target['shows'][0]['position'] = found_area['position'].copy()
+                target['shows'][0]['rotation'] = geometry.ORIGIN.copy()
```
ThomasSchellenbergNextCentury, line 117: Didn't know that you can just call `copy()` on a dict.  Neat.

scene_generator/goals.py
```diff
@@ -45,8 +46,13 @@ def instantiate_object(object_def, object_location):
         'id': str(uuid.uuid4()),
         'type': object_def['type'],
         'info': object_def['info'],
-        'mass': object_def['mass']
+        'mass': object_def['mass'],
```
brianpippin, line 49: Should this comma be here, looks like it's the last element of the object.

scene_generator/geometry.py
chrislong: Yeah, well, this is what  you risk getting when you invite a pointed-headed, over-educated nerd onto your team. ;-)

scene_generator/goals.py
chrislong: Also lists and sets. It's just a shallow copy, though, you need to use copy.deepcopy if you want a nested structure copied.

scene_generator/goals.py
chrislong: It doesn't hurt anything, but it's not required. Sharp eyes. I'll remove it.

PR #49, opened 2020-04-16
MCS-112 Calculating goal rewards

deanwetherby:
Let the arrows fly!

Note: I added 'sympy' to the setup.py package list. You will have to get the sympy into your venv in your preferred method. Sympy is used to do polygon intersection, point in polygon, etc. I didn't want to do all of that from scratch given our timeline.

Created MCS_Reward utility class with static methods for calculating the various rewards (retrieval, transferral, traversal) and all the helper methods necessary to do that. Added fairly extensive unit tests as well.

Removed 'machine_common_sense' from all imports in the package because they weren't needed as far as I can tell. 'from machine_common_sense.mcs_thing' became a relative import ie 'from .mcs_thing'. If this breaks something I'm not aware of, let me know. I left the unit tests alone because they need that additional namespace because of their location.

Questions:
1) Should the goal be 'traversal' or 'goto'?
2) Should we have a constants.py module? I had to move two constants in ai2thor controller outside the class so I could import in MCS_Reward.
3) I notice we rely on dictionaries a lot. I tried to avoid key errors by using dict.get('attr', None) and checking for None when I could.
4) I used type hinting on the MCS_Rewards class because it helps me when coding. Curious what others think about it. Hinting won't hurt anything as it's just a suggestion.


python_api/machine_common_sense/__init__.py
```diff
@@ -1,14 +1,30 @@
...
+from .mcs_controller_ai2thor import MCS_Controller_AI2THOR
+from .mcs_goal import MCS_Goal
+from .mcs_goal_category import MCS_Goal_Category
+from .mcs_material import MCS_Material
+from .mcs_object import MCS_Object
+from .mcs_pose import MCS_Pose
+from .mcs_return_status import MCS_Return_Status
+from .mcs_reward import MCS_Reward
+from .mcs_step_output import MCS_Step_Output
+from .mcs_util import MCS_Util
```
ThomasSchellenbergNextCentury, line 29: Does this still work if you reinstall the `machine_common_sense` library without the `-e` flag?  Please double-check it using the GitHub link that the TA1s use (well, modified to use this branch):

```
pip install git+https://github.com/NextCenturyCorporation/MCS@MCS-112
```

ThomasSchellenbergNextCentury: When I run this I get an error, and the stacktrace seems to imply some sort of problem finding packages, but it's not clear.

python_api/machine_common_sense/mcs_controller_ai2thor.py
```diff
@@ -360,6 +364,9 @@ def retrieve_object_output(self, object_metadata, object_id_to_color):
         rgb = object_id_to_color[object_metadata['objectId']] if object_metadata['objectId'] in object_id_to_color \
                 else [None, None, None]
 
+        #if object_metadata['objectBounds'] is None:
+        #    print(f"No bounds for {object_metadata['objectId']}")
```
ThomasSchellenbergNextCentury, line 368: Is this still needed?

python_api/machine_common_sense/__init__.py
deanwetherby: I get an error when installing to a fresh venv. I'll look into it.

python_api/machine_common_sense/mcs_goal.py
```diff
@@ -28,20 +28,20 @@ class MCS_Goal:
...
-        info_list=None,
+        info_list=[],
         last_preview_phase_step=0,
         last_step=None,
-        task_list=None,
-        type_list=None,
-        metadata=None
+        task_list=[],
+        type_list=[],
+        metadata={}
```
ThomasSchellenbergNextCentury, line 36: This doesn't work like you expect.  If you use a list or dict as a default argument, the function (or constructor in this case) uses the exact same list/dict in each call, rather than creating a new list/dict each time.

https://docs.python-guide.org/writing/gotchas/

python_api/machine_common_sense/mcs_controller_ai2thor.py
deanwetherby: Does it do anything? I was getting KeyErrors. I can change it to object_metadata.get('objectBounds' None) and then check the return value for None.

python_api/machine_common_sense/mcs_goal.py
deanwetherby: Thanks.

python_api/machine_common_sense/run_mcs_just_rotate.py
```diff
@@ -22,6 +24,12 @@
...
-    for i in range(1, 12):
+    for i in range(1, 11):
         output = controller.step('RotateLook', rotation=30)
 
+    # testing .... remove
+    output = controller.step('RotateLook', rotation=-15)
+    time.sleep(1)
+    output = controller.step('MoveAhead')
+    output = controller.step('MoveAhead')
+    output = controller.step('MoveAhead')
```
ThomasSchellenbergNextCentury, line 35: I use this script for creating some of the example gifs for our GitHub (sorry, I should probably write that down somewhere), so if you want to test out something, if you wouldn't mind using `run_mcs_environment.py` or making your own test script, I would appreciate it.

deanwetherby: accidental commit for testing.

python_api/machine_common_sense/__init__.py
deanwetherby: Moved sympy to install_requires. pip install was successful and sympy shows up in pip freeze.

deanwetherby: With a fresh install, I'm getting "No module named 'PIL'". I can add it to the install_requires.

deanwetherby: same with ai2thor

deanwetherby: Now I can pip install from the repo and use mcs.

>>> import machine_common_sense as mcs
>>> dir(mcs)
['MCS_Action', 'MCS_Action_Keys', 'MCS_Controller', 'MCS_Controller_AI2THOR', 'MCS_Goal', 'MCS_Goal_Category', 'MCS_Material', 'MCS_Object', 'MCS_Pose', 'MCS_Return_Status', 'MCS_Reward', 'MCS_Step_Output', 'MCS_Util', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__', 'mcs_action', 'mcs_action_keys', 'mcs_controller', 'mcs_controller_ai2thor', 'mcs_goal', 'mcs_goal_category', 'mcs_material', 'mcs_object', 'mcs_pose', 'mcs_return_status', 'mcs_reward', 'mcs_step_output', 'mcs_util']
>>> mcs.MCS_Action
<enum 'MCS_Action'>

python_api/machine_common_sense/mcs_goal.py
deanwetherby: I noticed some strangeness in testing that this explains now that I think about it. Now I'm wondering if this caused me issues in other projects.

python_api/machine_common_sense/__init__.py
deanwetherby: I'm thinking now that the ai2thor dependency would come from our own fork.

ThomasSchellenbergNextCentury: The install now works for me, and so does the `mcs_run_in_human_input_mode` script.  Thanks!

ThomasSchellenbergNextCentury: You can delete the commented-out imports now.

python_api/machine_common_sense/mcs_reward.py
```diff
@@ -0,0 +1,239 @@
...
+            objects: list of object dictionaries
+            target_id: str ID of the mcs_object to find
+
+        Returns:
+            target: object dictionary if found or None
+        '''
+        return next((o for o in objects if o['objectId'] == target_id), None)
+
+    @staticmethod
+    def _convert_bounds_to_polygons(goal_object: Dict) -> Tuple[sympy.Polygon, sympy.Polygon]:
```
ThomasSchellenbergNextCentury, line 33: Is there a reason why some of the functions start with one underscore and some start with two?  Do we want to be consistent?

deanwetherby: I had the single underscores as a suggestion of privacy but still may be called directly. The double underscores are used for name mangling and meant to be more private although nothing is private in Python. I can do whatever the team wants here.

python_api/machine_common_sense/mcs_reward.py
```diff
@@ -0,0 +1,239 @@
...
+
+        return reward
+
+    @staticmethod
+    def _calculate_default_reward(goal: MCS_Goal, scene_metadata: Dict) -> int:
+        '''Returns the default reward of 0; not achieved.'''
+        return GOAL_NOT_ACHIEVED
+
+    @staticmethod
+    def calculate_reward(goal: MCS_Goal, scene_metadata: Dict) -> int:
```
ThomasSchellenbergNextCentury, line 214: Rather than passing in the whole `scene_metadata` dict, would it be cleaner to pass in the list of `objects` and the `position` of the `agent`?

deanwetherby: I thought about that. Wasn't sure initially what I would need out of the dictionary but I can make that change now that it is more solidified.

python_api/machine_common_sense/mcs_reward.py
ThomasSchellenbergNextCentury: Personally, I think if we want something private (or as private as Python can make it), we should use the double underscores, and have no starting underscore for anything else.  I'm open to other opinions though.

python_api/machine_common_sense/run_mcs_just_rotate.py
```diff
@@ -22,6 +24,5 @@
 
     output = controller.start_scene(config_data)
 
-    for i in range(1, 12):
+    for i in range(1, 10):
```
ThomasSchellenbergNextCentury, line 27: Should be `12` here

ThomasSchellenbergNextCentury:
In response to your overall questions:

1. This is just nitpicking, but my problem with "go to" is that it's a verb (for most of the population who aren't old-school programmers) and the other goals are nouns.  There are other nouns we can use if "traversal" is too similar to "transferral".

2. Adding a `constraints.py` module (maybe with the `mcs_` prefix) sounds good to me.

3. That's good to note.

4. I personally like strong typing.  Didn't realize you could do it in Python.  Good to know!

deanwetherby:
1. It's already done as traversal so we'll leave it alone
2. Maybe we put in a ticket for later to have an mcs_constants.py module.
3. ...
4. So funny thing is that it's type "hinting". You can actually pass whatever you want into a function with a type hint. It was Guido's desire to get python to strong typing eventually and hinting was the first step towards that goal.

python_api/machine_common_sense/mcs_reward.py
chrislong: I used single underscore in goals.py to indicate protected methods (that are for subclasses to be able to call or override, but that shouldn't be used by "outsiders")

ThomasSchellenbergNextCentury: Ah, good point about protected methods for class inheritence.

ThomasSchellenbergNextCentury: We're not doing any of that here though, right?

deanwetherby: We aren't but I was thinking maybe a dev wanted to call a specific reward calculation rather than going through the switch statement in retrieve_reward. The double underscores were really meant to mangled. That was my thinking anyway but @chrislong is correct about protected/private which is how I've used the underscores before.

python_api/machine_common_sense/mcs_reward.py
```diff
@@ -0,0 +1,240 @@
...
+            polygons: Tuple(sympy.Polygon, sympy.Polygon)
+        '''
+        # split 3d object bounds into upper and lower boxes
+        bbox3d = goal_object['objectBounds']['objectBoundsCorners']
+        lower_box, upper_box = bbox3d[:4], bbox3d[4:]
+
+        # conver lower box plane to sympy Polygon
+        lower_corners = [(pt['x'], pt['z']) for pt in lower_box]
+        a, b, c, d = lower_corners
+        lower_polygon = sympy.Polygon(a, b, c, d)
```
chrislong, line 50: Just FYI, you can let Python unpack it:
`lower_polygon = sympy.Polygon(*lower_corners)`

python_api/machine_common_sense/mcs_reward.py
```diff
@@ -0,0 +1,240 @@
...
+
+        # convert upper box plane to sympy Polygon
+        upper_corners = [(pt['x'], pt['z']) for pt in upper_box]
+        e, f, g, h = upper_corners
+        upper_polygon = sympy.Polygon(e, f, g, h)
+
+        return lower_polygon, upper_polygon
+
+    @staticmethod
+    def __calc_distance_to_goal(action_obj_xz_pos: Tuple[float, float], goal_obj: Dict) -> float:
```
chrislong, line 60: Couldn't you use Polygon.distance (from sympy) (possibly with encloses_point, too)?

python_api/machine_common_sense/mcs_reward.py
```diff
@@ -0,0 +1,240 @@
...
+
+        Returns:
+            int: 1 for goal achieved, 0 otherwise
+
+        '''
+        reward = GOAL_NOT_ACHIEVED
+        goal_id = goal.metadata.get('target_id', None)
+        goal_object = MCS_Reward.__get_object_from_list(objects, goal_id)
+        if goal_object and ('isPickedUp' in goal_object) and goal_object['isPickedUp']:
+            reward = GOAL_ACHIEVED
```
chrislong, line 122: Not a big deal, but you could replace this condition with `goal_object and goal_object.get('isPickedUp', False)`

python_api/test/test_mcs_goal.py
```diff
@@ -0,0 +1,54 @@
...
+
+    @classmethod
+    def tearDownClass(cls):
+        pass
+
+    def test_action_list(self):
+        self.assertIsNone(self.goal.action_list)
+
+    def test_info_list(self):
+        self.assertFalse(self.goal.info_list)
```
chrislong, line 31: Is this supposed to test that info_list is an empty list? I think it would be a lot clearer to check that `len(info_list) == 0`

python_api/test/test_mcs_reward.py
```diff
@@ -0,0 +1,334 @@
...
+from machine_common_sense.mcs_util import MCS_Util
+
+
+class Test_MCS_Reward(unittest.TestCase):
+
+    def test_default_reward(self):
+        goal = MCS_Goal()
+        reward = MCS_Reward.calculate_reward(goal, objects={}, agent={})
+        self.assertFalse(reward)
+        self.assertIsInstance(reward, int)
```
chrislong, line 17: It's weird to me to test if something is False and then immediately check if it's an int. If you want to check that it's 0, I think it would be clearer to check that explicitly.

python_api/test/test_mcs_reward.py
```diff
@@ -0,0 +1,334 @@
...
+    def test_target_in_object_list(self):
+        obj_list = []
+        for i in range(10):
+            obj = {"objectId":str(i)}
+            obj_list.append(obj)
+        target_id = '7'
+        results = MCS_Reward._MCS_Reward__get_object_from_list(obj_list, target_id)
+        self.assertTrue(results)
+        self.assertIsInstance(results, dict)
+        self.assertEqual(results['objectId'], '7')
```
chrislong, line 39: Maybe use `target_id` here instead of `'7'`?

python_api/machine_common_sense/mcs_reward.py
```diff
@@ -0,0 +1,240 @@
...
+            return GOAL_NOT_ACHIEVED
+        
+        if action_object is None or ('isPickedUp' in action_object and action_object['isPickedUp']):
+            return GOAL_NOT_ACHIEVED
+
+        action_object_xz_center = action_object['position']['x'], \
+            action_object['position']['z']
+
+        # actions are next_to or on_top_of (ie; action obj next to goal obj)
+        if action == 'next_to':
```
chrislong, line 193: It's kind of icky to hard-code magic strings in the middle of code like this. Maybe these should be defined as constants somewhere? (also in the test code)

python_api/machine_common_sense/mcs_reward.py
deanwetherby: yes. good point

python_api/machine_common_sense/mcs_reward.py
deanwetherby: I'm not following.

python_api/machine_common_sense/mcs_reward.py
deanwetherby: Yes. Thank you. I use the dictionary get everywhere but had a block on this one.

python_api/machine_common_sense/mcs_reward.py
deanwetherby: Agreed. Suggested a future story to add an mcs_constants.py module for these things.

python_api/test/test_mcs_goal.py
deanwetherby: This should be more human readable if that's your concern.

self.assertEqual(len(self.goal.info_list), 0)

python_api/machine_common_sense/mcs_reward.py
chrislong: Maybe I'm not understanding what this does, but I thought it was basically computing the distance between a polygon and a point. Which you could use Polygon.distance for, without having to mess with edges or sorting.

deanwetherby: I started out one way to get the reward calculation done but if sympy does what you suggest, then this gets simplified dramatically. Let me check.

PR #51, opened 2020-04-17
Mcs 34

wagnerwj:


ThomasSchellenbergNextCentury:
@wagnerwj Some of the actions are being incorrectly nested in an extra layer of list.

Example:

```json
  "answer": {
    "actions": [
      [
        {
          "action": "RotateLook",
          "params": {
            "rotation": -118.0,
            "horizon": 0.0
          }
        },
        {
          "action": "MoveAhead",
          "params": {}
        },
        {
          "action": "MoveAhead",
          "params": {}
        },
        {
          "action": "MoveAhead",
          "params": {}
        },
        {
          "action": "MoveAhead",
          "params": {}
        },
        {
          "action": "MoveAhead",
          "params": {}
        },
        {
          "action": "MoveAhead",
          "params": {}
        },
        {
          "action": "MoveAhead",
          "params": {
            "amount": 0.07
          }
        }
      ],
      {
        "action": "PickupObject",
        "params": {
          "objectId": "16d862de-10ae-46b6-bfee-7268edfd6f23"
        }
      }
    ]
  }
```

The actions from the RotateLook to the last MoveAhead should not be inside the nested list.

ThomasSchellenbergNextCentury:
@wagnerwj So, I think there's a problem here, but I didn't think about it until now (and I'm not blaming you for it).  If you try to PickupObject but the object is on the ground and small enough to be below your camera's field-of-view, the action will fail.  I think we either need to adjust the first RotateLook so you rotate up/down (the "horizon") to face the center of the target object, or add a second RotateLook after moving but before picking up.

scene_generator/goals.py
```diff
@@ -220,6 +227,56 @@ def add_objects(self, object_list, rectangles, performer_position):
...
+                theta = 90
+            else:
+                theta = 270
+        else:
+            theta = degrees(arctan(dz/dx))
+            if dx < 0:
+                theta += 180
+            #IF my calculations are correct, this should be right no matter what
+            # I'm assuming a positive angle is a clockwise rotation- so this should work
+            #I think
```
chrislong, line 247: Would math.atan2 work, instead of all this?

scene_generator/goals.py
```diff
@@ -220,6 +227,56 @@ def add_objects(self, object_list, rectangles, performer_position):
...
+            action = {
+                'action': 'RotateLook',
+                'params': {
+                    'rotation': round(delta_t,0),
+                    'horizon': 0.0
+                    }
+                }
+            actions.append(action)
+        distance = sqrt( dx ** 2 + dz ** 2 )
+        frac, whole = math.modf(distance / 0.5)
```
chrislong, line 265: What's this about? Is 0.5 a magic constant from somewhere?

scene_generator/goals.py
```diff
@@ -220,6 +227,56 @@ def add_objects(self, object_list, rectangles, performer_position):
...
+        for wholes in range(int(whole)):
+            actions.append({
+                    "action": "MoveAhead",
+                    "params": {}
+                    })
```
chrislong, line 270: FYI, you can also get a list with duplicated elements with `*`, like
`[{"action":...}]*int(whole)`

scene_generator/goals.py
```diff
@@ -220,6 +227,56 @@ def add_objects(self, object_list, rectangles, performer_position):
...
+        actions.append({
+                "action": "MoveAhead",
+                "params": {
+                    "amount": round(frac,POSITION_DIGITS)
+                    }
+            })
```
ThomasSchellenbergNextCentury, line 276: Something funny is going on here, but I'm not sure exactly where.

So in the example I generated, I get 7 MoveAhead actions, but only the last one has its own "amount" parameter.

The "amount" parameter is a number between 0 and 1, and that's the percentage to multiply by the maximum move of `0.5`.  The default "amount" is `0.5` (50%), which means the default move is `0.5 * 0.5 = 0.25`.

Shouldn't my example's answer have 4 MoveAhead actions instead of 7, where the first 3 all have `"amount": 1.0`?

scene_generator/goals.py
```diff
@@ -237,7 +294,11 @@ def generate_walls(self, material, performer_position, bounding_rects):
             else:
                 logging.warning('could not generate wall')
         return walls
-        
+    @abstractmethod
+    def find_answer(self, goal_objects, all_objects):
+        '''Compute the optimal set of moves and up[date the body object'''
```
chrislong, line 299: I'd prefer something more descriptive, like `find_optimal_path`, but it's not a big deal. Also, stray `[` in the comment.

scene_generator/goals.py
```diff
@@ -368,12 +457,14 @@ class RelationshipType(Enum):
     def __init__(self):
         super(TransferralGoal, self).__init__()
 
+
```
chrislong, line 460: FYI, the python standard is one blank line between methods (and two between top level functions). See https://www.python.org/dev/peps/pep-0008/#blank-lines

scene_generator/goals.py
```diff
@@ -401,7 +492,42 @@ def get_config(self, objects):
         goal['description'] = f'Find and pick up the {target1["info"][-1]} and move it {relationship.value} ' \
             f'the {target2["info"][-1]}.'
         return goal
-
+    
+    def find_answer(self, goal_objects, all_objects):
```
chrislong, line 496: Seems like a lot of duplicated code between the other `find_answer` and this one. Maybe factor out the repeated code?

scene_generator/optimal_path.py
```diff
@@ -0,0 +1,32 @@
+from extremitypathfinder.extremitypathfinder import PolygonEnvironment as Environment
+
+from  geometry import MAX_SCENE_POSITION, MIN_SCENE_POSITION
```
chrislong, line 3: extra space after from

scene_generator/optimal_path.py
```diff
@@ -0,0 +1,32 @@
...
+from math import degrees
+from numpy.ma.core import arctan
+
+
+
+# passing the other_rects from my  calculations previously
+# This is the source for extremitypathfinder: https://github.com/MrMinimal64/extremitypathfinder
+#other_loc will probably contain the target- we'll see
```
chrislong, line 11: Do you mean `target_loc`?

scene_generator/optimal_path.py
```diff
@@ -0,0 +1,32 @@
...
+def generatepath(source_loc, target_loc, other_rects):
+    '''Boundary has to be CCW, Holes CW'''
+    boundary_coordinates = [ (MAX_SCENE_POSITION, MAX_SCENE_POSITION),(MIN_SCENE_POSITION, MAX_SCENE_POSITION),(MIN_SCENE_POSITION, MIN_SCENE_POSITION), (MAX_SCENE_POSITION,MIN_SCENE_POSITION)]
+    holes= []
```
chrislong, line 15: So you're defining the objects in the room as holes? That seems weird, but maybe it's just me.

scene_generator/goals.py
deanwetherby: What about using a standard formatter like yapf or black or even a linter? Just throwing it out there. @chrislong had mentioned something about this earlier and I'm seconding here.

scene_generator/goals.py
deanwetherby: Yes!
>>> ['i']*5
['i', 'i', 'i', 'i', 'i']

scene_generator/goals.py
ThomasSchellenbergNextCentury: Linter ticket is in the backlog (MCS-38).  I figured we'd discuss it when we had some downtime.

scene_generator/optimal_path.py
ThomasSchellenbergNextCentury: They're holes in the available space within which the "player" may move, right?  That makes sense to me.  Is this just a terminology thing?

scene_generator/goals.py
ThomasSchellenbergNextCentury: This may be the maximum move distance for a single Move action.  It would be good to make this a constant, though.

scene_generator/goals.py
ThomasSchellenbergNextCentury: Chris, don't you like space?  It's the final frontier!

scene_generator/goals.py
deanwetherby: MAX_MOVE_DISTANCE is now importable from mcs_controller_ai2thor. I needed it for mcs_reward.

from .mcs_controller_ai2thor import MAX_MOVE_DISTANCE

scene_generator/optimal_path.py
```diff
@@ -0,0 +1,32 @@
...
+    #need to convert to lists of points 
+    for object in other_rects:            
+        holes.append([(point['x'],point['z']) for point in object])
+        
+        
+    environment = Environment()
+    environment.store(boundary_coordinates, holes, validate = True)
+    environment.prepare()
+    
+    path,length = environment.find_shortest_path(source_loc, target_loc)
```
ThomasSchellenbergNextCentury, line 26: I got an error on this line from the extremitypathfinder during one run:  `ValueError: start or goal do not lie within the map`

Ideally, could we:
- Catch the error
- Print a useful log message (maybe with the inputs to `generatepath`)
- Return an empty path, or something?

scene_generator/goals.py
ThomasSchellenbergNextCentury: Yeah, I'm generating some answers with RotateLook rotation values that don't make sense.  For example, I start facing some furniture, and the answer turns me to the left, facing a wall, instead of turning me to the right.

(Also, positive angle IS clockwise rotation.)

ThomasSchellenbergNextCentury:
Is there any logic to detect if a path is simply impossible (because you're completely blocked off from reaching the target object)?  If not, we can probably shift that to a new ticket, since I don't think it's very likely.

wagnerwj:
> i



> @wagnerwj So, I think there's a problem here, but I didn't think about it until now (and I'm not blaming you for it). If you try to PickupObject but the object is on the ground and small enough to be below your camera's field-of-view, the action will fail. I think we either need to adjust the first RotateLook so you rotate up/down (the "horizon") to face the center of the target object, or add a second RotateLook after moving but before picking up.

Go ahead and write that ticket- it should be easy to figure

ThomasSchellenbergNextCentury:
Billy, since there are a lot of comments, here's a summary of the comments so far:
- There's an extra layer of listing within `answer.actions`
- The rotation values sometimes are off
- Set the default MoveAhead `amount` to `1.0` (translates to 100% of move magnitude `0.5`)
- Catch errors in the `extremitypathfinder` (`find_shortest_path` function)
- Chris had made a few code style suggestions

Separate tickets:
- Add RotateLook to the action list before PickupObject or PutObject actions https://nextcentury.atlassian.net/browse/MCS-161
- Detect any impossible paths https://nextcentury.atlassian.net/browse/MCS-162

scene_generator/goals.py
wagnerwj: We can do that- Your expected results had a blank parameter... plus, when testing, when I put in a "amount": 1 parameter, I moved forward 1, and the moves were off

scene_generator/optimal_path.py
wagnerwj: ITs what the algorithm "thinks". IF you can't go there, its a hole in the environment.

scene_generator/optimal_path.py
wagnerwj: This should really never happen- when this happens, one of the points is inside a 'hole'- currently one of the objects in the space.. since I take out (well, HOPEFULLY take out) the hole containing the point, we shouldn't see this- so this is most likely a bug in removing holes..
PRinting out the exact conditions would be useful

scene_generator/optimal_path.py
chrislong: Oh, ok. I thought maybe there was support for polygons with holes in them. (Although now that I think about it, I don't know why it would need support for that since such holes would be inaccessible.)

wagnerwj:
Alright  @ThomasSchellenbergNextCentury @chrislong @deanwetherby  IF you could take a look at this, I think I have those things T listed done.. please review

wagnerwj:
Going to merge. IF you find any more issues, please bug it

PR #55, opened 2020-04-20
Scene Generator:  Choose Dimensions

ThomasSchellenbergNextCentury:
Fixed a bug in the scene generator that caused it to fail if an object definition's dimensions was in its 'choose' list.

### DIFF
--- a/scene_generator/goals.py
+++ b/scene_generator/goals.py
@@ -29,19 +29,24 @@ WALL_COUNTS = [0, 1, 2, 3]
 WALL_PROBS = [60, 20, 10, 10]
 
 
+def finalize_object_definition(object_def):
+    object_def_copy = copy.deepcopy(object_def)
+
+    # apply choice if necessary
+    if 'choose' in object_def_copy:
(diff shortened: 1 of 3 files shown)


scene_generator/goals.py
```diff
@@ -29,19 +29,24 @@
...
     if object_def is None or object_location is None:
         raise ValueError('instantiate_object cannot take None parameters')
 
-    # apply choice if necessary
-    if 'choose' in object_def:
-        object_def = object_def.copy()
-        choice = random.choice(object_def['choose'])
-        for key in choice:
-            object_def[key] = choice[key]
-
```
chrislong, line 44: Shouldn't this call `finalize_object_definitiion`? Either that, or else everywhere we call `instantiate_object` we have to be sure to call `finalize_object_definition` first, and it seems safer to me to do it here.

ThomasSchellenbergNextCentury: I think I handled those cases, but I also think it wouldn't hurt to call the function again, since `finalize_object_definition` deletes the `choose` property from the copied definition.  I'll add it now.

PR #56, opened 2020-04-20
MCS-155 Added dimensions and new distance properties to MCS Object class

ThomasSchellenbergNextCentury:
- Added dimensions and new distance properties to the MCS Object class as used in the MCS AI2-THOR Controller.
- Fixed issues with Python class constructors.


rartiss55:
I think the distance reference in generate_pretty_object_output (MCS_Util) needs to be updated to distance_in_steps -- otherwise looks good!

ThomasSchellenbergNextCentury:
Nice catch @rartiss55 , thanks!

PR #57, opened 2020-04-21
MCS-159 Prevent duplicate goal objects

deanwetherby:
Finally got a unit test working. I would prefer more extensive testing but with the impending release more eyes on it will have to do.

### DIFF
--- a/scene_generator/goals.py
+++ b/scene_generator/goals.py
@@ -200,6 +200,7 @@ class Goal(ABC):
 
     def __init__(self):
         self._performer_start = None
+        self._targets = []
 
     def update_body(self, body):
         """Helper method that calls other Goal methods to set performerStart, objects, and goal."""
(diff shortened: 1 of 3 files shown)


scene_generator/goals.py
```diff
@@ -242,11 +243,16 @@ def compute_objects(self):
     def add_objects(self, object_list, rectangles, performer_position):
         """Add random objects to fill object_list to some random number of objects up to MAX_OBJECTS. If object_list
         already has more than this randomly determined number, no new objects are added."""
+        print("Goal add_objects called")
```
brianpippin, line 246: Do we want to keep all these print statements?

deanwetherby: probably not

scene_generator/goals.py
```diff
@@ -326,6 +332,9 @@ def _set_goal_objects(self):
 
     def add_objects(self, all_objects, bounding_rects, performer_position):
         """Maybe add a container and put the target inside it. If so, maybe put other objects in other objects, too."""
+        print("interaction add_objects called")
```
brianpippin: Same for these print statements

PR #58, opened 2020-04-21
MCS-167 Base Target must be stackable

deanwetherby:


scene_generator/goals.py
```diff
@@ -419,6 +419,9 @@ def get_config(self, objects):
         target1, target2 = objects
         if not target1.get('pickupable', False):
             raise ValueError(f'first object must be "pickupable": {target1}')
+        if not target2.get('stackTarget', False):
+            
+            raise ValueError(f'second object must be "stackable": {target2}')
```
ThomasSchellenbergNextCentury, line 424: What happens if this fails?  Does the whole generator fail?  Maybe we should keep picking new objects until we get a valid stackTarget?

deanwetherby: I'm just following the existing pattern used for the pickupable object. I assume either condition fails the generator.

ThomasSchellenbergNextCentury: We have code elsewhere to ensure that target1 is pickuable (see line 317), so the error check for that isn't even needed here.  We don't have the same assurances for target2, though -- it's just any random object (see line 411).

deanwetherby: > In addition, I think `TransferralGoal._set_goal_objects` should be changed to select a `target2_def` that is stackable. Otherwise this will only sometimes succeed.

I'm not seeing OBJECTS_STACKABLE in objects.py. How can we get a stackable object?

ThomasSchellenbergNextCentury: Can we just keep calling choose_object_def until we find one that has the stackTarget attribute?

deanwetherby: I can do that but what if none are found? Is that possible?

ThomasSchellenbergNextCentury: It's highly unlikely but still possible.  You can loop for MAX_TRIES and then throw an error if you still can't find a valid stackTarget, I guess?

Alternatively, you can create an OBJECTS_STACKABLE list in the objects.py file like we're doing with ENCLOSED_CONTAINERS and get_enclosed_containers, but I think the get_enclosed_containers function actually has a bug in it (doesn't account for the 'choose' property).

chrislong: I think it would be cleaner to avoid randomly trying things until they succeed where we can. I would just call `objects.get_all_object_defs`, shuffle the result, and loop through it until a valid stackTarget was found.

chrislong: @ThomasSchellenbergNextCentury That's an interesting point about get_enclosed_containers and the choose property. You're right, that's a bug.

ThomasSchellenbergNextCentury: @chrislong I don't think it will cause a big issue with generating the training data, and the fix may be non-trivial.

PR #59, opened 2020-04-21
MCS-169: Allow generating specific goal

chrislong:
enhanced --goal to accept specific goal name


scene_generator/goals.py
```diff
@@ -518,8 +519,16 @@ def choose_goal(goal_type):
     if goal_type is None:
         return EmptyGoal()
     else:
-        return random.choice(GOAL_TYPES[goal_type])()
+        if goal_type in GOAL_TYPES:
+            return random.choice(GOAL_TYPES[goal_type])()
+        else:
+            class_name = goal_type + 'Goal'
```
ThomasSchellenbergNextCentury, line 525: Does goal_type need to have its first letter capitalized for this to work?

chrislong: Yes. This will be enforced in the command line parser. I think it may be preferable, anyway, because it makes it clearer for the user whether they're asking for a specific goal or a goal type.

PR #60, opened 2020-04-21
Make Optimal Path Optional

ThomasSchellenbergNextCentury:
Added a flag to the scene generator to make calculating the optimal path for interaction goals an optional behavior used for evaluation data (since I think we decided that we don't want it for our training data).

### DIFF
--- a/scene_generator/goals.py
+++ b/scene_generator/goals.py
@@ -205,7 +205,8 @@ class Goal(ABC):
     get_config. Users of a goal object should normally only need to call 
     update_body."""
 
-    def __init__(self):
+    def __init__(self, find_path=False):
+        self._find_path = find_path
         self._performer_start = None
         self._targets = []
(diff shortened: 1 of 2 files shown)


scene_generator/goals.py
```diff
@@ -205,7 +205,8 @@ class Goal(ABC):
     get_config. Users of a goal object should normally only need to call 
     update_body."""
 
-    def __init__(self):
+    def __init__(self, find_path=False):
+        self._find_path = find_path
```
chrislong, line 209: I'd rather have it as a parameter to `update_body` than `__init__`. Then it doesn't have to propagate as many places, including `choose_goal`, where it feels like it really doesn't belong.

ThomasSchellenbergNextCentury: Good idea.  It's done.

PR #61, opened 2020-04-23
Mcs 168

wagnerwj:
Bug fix for faulty transformations


scene_generator/goals.py
```diff
@@ -666,3 +667,9 @@ def choose_goal(goal_type):
 
 def get_goal_types():
     return GOAL_TYPES.keys()
+
+
+def test__object_collision():
+    r1=geometry.calc_obj_coords(-1.97,1.75, 1.1,.89,315)
+    r2=geometry.calc_obj_coords(-3.04,.85,3.35,.1,315)
+    sat_entry(r1,r2)
```
chrislong, line 675: Should there be an `assert` here?

wagnerwj: OH! Thanks

wagnerwj: IF you can look at it again, please?

PR #62, opened 2020-04-23
Mcs 32: IntPhys Scene Generator: Goals

chrislong:


scene_generator/goals.py
```diff
@@ -643,8 +644,144 @@ def find_optimal_path(self, goal_objects, all_objects):
         return actions
 
 
+class IntPhysGoal(Goal, ABC):
```
deanwetherby, line 647: Does the class need multiple inheritance with both Goal and ABC?

chrislong: To tag it as an abstract class, I'm pretty sure it needs to inherit directly from ABC. Otherwise it could be a concrete class, which we don't want.

PR #63, opened 2020-04-24
Mcs 171: calc_obj_pos in goals.py is broken

chrislong:
fixed multiple bugs

* missing arguments for calc_obj_coords
* missing arg for point_within_room
* broken test for pickupable-ness and stackTarget-ness
* missing arguments to superclass method call for update_body
* JSON-ification bug

### DIFF
--- a/scene_generator/geometry.py
+++ b/scene_generator/geometry.py
@@ -3,7 +3,6 @@ import math
 import random
 
 from separating_axis_theorem import sat_entry
-from setuptools.sandbox import AbstractSandbox
 
 MAX_TRIES = 100
 # the following mins and maxes are inclusive
@@ -101,6 +100,12 @@ object in the frame, None otherwise."""
 
     dx = old_object['dimensions']['x']/2.0
     dz = old_object['dimensions']['z']/2.0
+    if 'offset' in old_object:
+        offset_x = old_object['offset']['x']
+        offset_z = old_object['offset']['z']
+    else:
+        offset_x = 0.0
+        offset_z = 0.0
 
     tries = 0
     while tries < MAX_TRIES:
@@ -108,7 +113,7 @@ object in the frame, None otherwise."""
         new_x = random_position()
         new_z = random_position()
 
-        rect = calc_obj_coords(new_x, new_z, dx, dz, rotation)
+        rect = calc_obj_coords(new_x, new_z, dx, dz, offset_x, offset_z, rotation)
         if not collision(rect, performer_position) and \
                 rect_within_room(rect) and \
                 (len(other_rects) == 0 or not any(sat_entry(rect, other_rect) for other_rect in other_rects)):
--- a/scene_generator/goals.py
+++ b/scene_generator/goals.py
@@ -143,9 +143,9 @@ def generate_wall(wall_mat_choice, performer_position, other_rects):
         new_x = random_position()
         new_z = random_position()
         new_x_size = round(random.uniform(MIN_WALL_WIDTH, MAX_WALL_WIDTH), POSITION_DIGITS)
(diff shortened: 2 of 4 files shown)


scene_generator/goals.py
```diff
@@ -496,16 +496,16 @@ def _set_goal_objects(self):
...
         if len(objects) < 2:
             raise ValueError(f'need at least 2 objects for this goal, was given {len(objects)}')
-        target1, target2 = objects
-        print(target2)
-        if 'pickupable' not in target1.get('attributes', []):
+        target1, target2 = objects[0:2]
+        if not target1.get('pickupable', False):
             raise ValueError(f'first object must be "pickupable": {target1}')
-        if 'stackTarget' not in target2.get('attributes', []):
+        if not target2.get('stackTarget', False):
```
deanwetherby, line 508: stackTarget and pickupable are part of the attributes field and they're not booleans. This won't work.
objects.py
143    "attributes": ["moveable", "pickupable", "stackTarget"],

chrislong: They're attributes in the object definitions, but they turn into fields when instantiated. See `instantiate_object`. I think it's a bit confusing, I'm not sure why it was decided to be that way.

deanwetherby: Ok. I see it now.
    for attribute in object_def['attributes']:
        new_object[attribute] = True
It is confusing.

scene_generator/geometry.py
```diff
@@ -101,14 +100,20 @@ def calc_obj_pos(performer_position, other_rects, old_object):
 
     dx = old_object['dimensions']['x']/2.0
     dz = old_object['dimensions']['z']/2.0
+    if 'offset' in old_object:
+        offset_x = old_object['offset']['x']
+        offset_z = old_object['offset']['z']
+    else:
+        offset_x = 0.0
+        offset_z = 0.0
```
ThomasSchellenbergNextCentury, line 108: This isn't super important, but can't this be

```
offset_x = old_object['offset']['x'] if 'offset' in old_object else 0.0
offset_z = old_object['offset']['z'] if 'offset' in old_object else 0.0
```

scene_generator/goals.py
ThomasSchellenbergNextCentury: If you all think it's less confusing to replace the attributes in the object definitions with separate properties, and just copy each property over into the new_object, that's fine.  Just make a ticket for it.

scene_generator/geometry.py
chrislong: Yeah, that would work, too. Strangely, I usually prefer concision, but I like it the more verbose way this time. But I don't care that much either way.

scene_generator/goals.py
```diff
@@ -496,16 +496,16 @@ def _set_goal_objects(self):
             raise ValueError(f'No stack targets found for transferral goal')
         target2_location = calc_obj_pos(self._performer_start['position'], self._bounding_rects, target2_def)
         target2 = instantiate_object(target2_def, target2_location)
+        logging.debug(f'target2 = {target2}')
```
brianpippin, line 499: Do we need this log checked in?

chrislong: I guess not, but the advantage of using logging is that it won't show up unless the user enables log level at least debug.

brianpippin: I guess what I should have asked was is this logging important for logging in the program or was it more of a development debug logging that we didn't need.  I should have been more specific.

chrislong: I'm not sure. This PR has been merged, so I'll take it out in the branch I'm working on now (MCS-118).

PR #64, opened 2020-04-24
Mcs 116: IntPhys Scene Generator: Objects Moving Across

chrislong:


scene_generator/pretty_json/pretty_json.py
```diff
@@ -33,7 +33,7 @@ def encode(self, obj):
             # see https://stackoverflow.com/a/15012814/355230
             id = int(match.group(1))
             no_indent = PyObj_FromPtr(id)
-            json_obj_repr = json.dumps(no_indent.value, sort_keys=self.__sort_keys, separators=(',', ':'))
+            json_obj_repr = json.dumps(no_indent.value, cls=PrettyJsonEncoder, sort_keys=self.__sort_keys, separators=(',', ':'))
```
ThomasSchellenbergNextCentury, line 36: So this is... recursively calling itself?  Does that work?  Won't that just loop infinitely?

chrislong: I confess I don't fully understand the control flow, but this works and it didn't before.

scene_generator/goals.py
```diff
@@ -695,8 +704,109 @@ def compute_objects(self):
...
+            for position in position_by_step:
+                if location in ('a', 'b', 'c', 'd'):
+                    position = object_position_x - position
+                else:
+                    position = object_position_x + position
+                new_positions.append(position)
+            if location in ('a', 'b', 'e', 'f'):
+                max_x = 3.55 + obj_def['scale']['x'] / 2.0 * 1.2
+            else:
+                max_x = 4.2 + obj_def['scale']['x'] / 2.0 * 1.2
```
ThomasSchellenbergNextCentury, line 790: Sorry, I know I put these magic numbers in the ticket without much of an explanation.  The `3.55` or `4.2` is the position at which the object will leave the camera's viewport, and is dependent on the object's Z position (either `1.6` or `2.7`).  The `* 1.2` is to account for the camera's perspective.  You can leave that info in a comment here and/or make global variables for the magic numbers.

scene_generator/goals.py
```diff
@@ -695,8 +704,109 @@ def compute_objects(self):
...
+            # TODO: later this will get imported from objects (or somewhere else)
+            from objects_intphys_v1 import OBJECTS_INTPHYS
+            obj_def = finalize_object_definition(random.choice(OBJECTS_INTPHYS))
+            remaining_intphys_options = obj_def['intphys_options'].copy()
+            while len(remaining_intphys_options) > 0:
+                intphys_option = random.choice(remaining_intphys_options)
+                if location in velocity_ordering and velocity_ordering[location] in location_assignments:
+                    # ensure the objects won't collide
+                    other_obj = location_assignments[velocity_ordering[location]]
+                    # TODO: compute value for collision
```
ThomasSchellenbergNextCentury, line 754: I created [MCS-188](https://nextcentury.atlassian.net/browse/MCS-188) for this, though I still need to add a description.

scene_generator/goals.py
chrislong: Got it, will do.

scene_generator/goals.py
```diff
@@ -695,8 +704,109 @@ def compute_objects(self):
...
+        num_objects = random.choices((1, 2, 3), (40, 30, 30))[0]
+        object_positions = {
+            'a': (4.2, 1.6),
+            'b': (5.3, 1.6),
+            'c': (4.8, 2.7),
+            'd': (5.9, 2.7),
+            'e': (-4.2, 1.6),
+            'f': (-5.3, 1.6),
+            'g': (-4.8, 2.7),
+            'h': (-5.9, 2.7)
```
ThomasSchellenbergNextCentury, line 716: Would it be better to make the different positions into an enum or variables or something?  Also, we could change the names, if you want, from A-H to maybe something like "left_first_close", "right_last_far", etc.

scene_generator/pretty_json/pretty_json.py
ThomasSchellenbergNextCentury: Huh, interesting.  It's from stackoverflow (the link is at the top) if you're interested (I just renamed the two classes).

scene_generator/goals.py
chrislong: Sounds good. I'll add a reference to the ticket here

scene_generator/pretty_json/pretty_json.py
ThomasSchellenbergNextCentury: As long as it works, then everything's good.

scene_generator/goals.py
chrislong: That seems like a pain, but if it would make the code a lot easier to understand, then it's probably worth it.

ThomasSchellenbergNextCentury: Would it minimize the pain if, instead of doing enums or anything super fancy, we just replaced each string `'a'` through `'h'` with a global variable?

chrislong: Actually, just changing the literals (`'a'`, etc.) to enum values (e.g., `LEFT_FIRST_NEAR`) wouldn't be that bad. I'll do that. For naming, there are 3 parts: left/right and near/far, but what should we call the other? first/last? (like to distinguish (4.2, 1.6) and (5.3, 1.6))

chrislong: Also, I just realized I don't know which direction is which. I positive right or left?

ThomasSchellenbergNextCentury: First and last seem fine to me.  Positive is right.

chrislong:
Addressed comments, ready for review again.

PR #67, opened 2020-04-29
Mcs 117: IntPhys Scene Generator: Objects Falling Down Behind Occluders

chrislong:


scene_generator/goals.py
```diff
@@ -36,6 +37,13 @@
 WALL_PROBS = [60, 20, 10, 10]
 
 
+def random_real(a, b, step):
+    """Return a random real number N where a <= N <= b and N - a is divisible by step."""
+    steps = int((b - a) / step)
+    n = random.randint(0, steps)
+    return a + n * step
```
brianpippin, line 44: I know you probably don't need parens here because people should know order of operations, but I feel like having them would make the code read easier.

scene_generator/goals.py
```diff
@@ -689,18 +698,86 @@ def generate_walls(self, material, performer_position, bounding_rects):
...
+        return [], objs + occluders, []
 
     def _get_objects_moving_across(self):
         # TODO: in a future ticket
         return []
 
-    def _get_objects_falling_down(self):
-        # TODO: in a future ticket
-        return []
+    def _get_objects_falling_down(self, wall_material_name):
```
brianpippin, line 710: So this function has a bunch of numbers in it, should a lot of these be constants with names that say why you picked them, some of them have comments about them, but some of them don't, and I am not sure why some of the numbers were picked.

chrislong: Yeah, that's fair. The 1.6 and 2.7 I will consolidate (NEAR_Z and FAR_Z), but for the others I'll need input from @ThomasSchellenbergNextCentury

scene_generator/goals.py
```diff
@@ -722,14 +710,14 @@ class Position(Enum):
...
-            Position.LEFT_FIRST_FAR: (-4.8, 2.7),
-            Position.LEFT_LAST_FAR: (-5.9, 2.7)
+            Position.RIGHT_FIRST_NEAR: (4.2, IntPhysGoal.OBJECT_NEAR_Z),
+            Position.RIGHT_LAST_NEAR: (5.3, IntPhysGoal.OBJECT_NEAR_Z),
+            Position.RIGHT_FIRST_FAR: (4.8, IntPhysGoal.OBJECT_FAR_Z),
+            Position.RIGHT_LAST_FAR: (5.9, IntPhysGoal.OBJECT_FAR_Z),
+            Position.LEFT_FIRST_NEAR: (-4.2, IntPhysGoal.OBJECT_NEAR_Z),
+            Position.LEFT_LAST_NEAR: (-5.3, IntPhysGoal.OBJECT_NEAR_Z),
+            Position.LEFT_FIRST_FAR: (-4.8, IntPhysGoal.OBJECT_FAR_Z),
+            Position.LEFT_LAST_FAR: (-5.9, IntPhysGoal.OBJECT_FAR_Z)
```
ThomasSchellenbergNextCentury, line 720: You can add a comment here about these numbers being X positions that start outside of the camera viewport and ensure that objects with scales of 1 don't collide with each other.

ThomasSchellenbergNextCentury: The greater the Z, the greater the X due to the angle of the camera's perspective.

scene_generator/goals.py
```diff
@@ -824,9 +812,81 @@ class Position(Enum):
...
+                    distance = abs(obj['shows'][0]['position']['x'] - x_position)
+                    too_close = distance < min_obj_distance
+                if not too_close:
+                    found_space = True
+            if not found_space:
+                raise GoalException(f'Could not place {i+1} objects to fall down')
+            location = {
+                'position': {
+                    'x': x_position,
+                    'y': 3.8,
```
ThomasSchellenbergNextCentury, line 845: This Y position ensures that the object starts above the camera's viewport.

scene_generator/goals.py
```diff
@@ -824,9 +812,81 @@ class Position(Enum):
...
+            obj['shows'][0]['stepBegin'] = random.randint(13, 20)
+            object_list.append(obj)
+        # place required occluders, then (maybe) some random ones
+        num_occluders = 2 if num_objects == 2 else random.choice((1, 2))
+        occluders = []
+        non_wall_materials = [m for m in materials.CEILING_AND_WALL_MATERIALS
+                              if m[0] != wall_material_name]
+        for i in range(num_objects):
+            paired_obj = object_list[i]
+            min_scale = min(max(paired_obj['shows'][0]['scale']['x'], MIN_OCCLUDER_SCALE), 1)
```
ThomasSchellenbergNextCentury, line 860: This `1` could probably be `MAX_OCCLUDER_SCALE`

scene_generator/goals.py
```diff
@@ -824,9 +812,81 @@ class Position(Enum):
...
+            x_position = paired_obj['shows'][0]['position']['x']
+            paired_z = paired_obj['shows'][0]['position']['z']
+            factor = NEAR_X_PERSPECTIVE_FACTOR if paired_z == IntPhysGoal.OBJECT_NEAR_Z else FAR_X_PERSPECTIVE_FACTOR
+            # Determine the biggest scale we could use for the new
+            # occluder (up to 1) so it isn't too close to any of the
+            # others.
+            max_scale = 1
```
ThomasSchellenbergNextCentury, line 867: I think this is also `MAX_OCCLUDER_SCALE`

scene_generator/goals.py
```diff
@@ -824,9 +812,81 @@ class Position(Enum):
...
+        num_objects = random.choice((1, 2))
+        object_list = []
+        for i in range(num_objects):
+            found_space = False
+            # It doesn't matter how close the objects are to each
+            # other, but each one must have an occluder, and those
+            # have to be a certain distance apart, so these objects
+            # do, too.
+            for _ in range(MAX_POSITION_TRIES):
+                x_position = random_real(-2.5, 2.5, 0.05)
```
ThomasSchellenbergNextCentury, line 833: These min/max X positions will ensure that the occluders won't be positioned partially/completely outside of the camera's viewport, and also leave a gap so we can see when an object enters/leaves the scene.

scene_generator/goals.py
ThomasSchellenbergNextCentury, line 867: I think this starting value is also `MAX_OCCLUDER_SCALE`

scene_generator/goals.py
```diff
@@ -824,9 +812,81 @@ class Position(Enum):
...
+                    max_scale = scale
+            x_scale = random_real(min_scale, max_scale, 0.05)
+            adjusted_x = x_position * factor
+            occluder_pair = objects.create_occluder(random.choice(non_wall_materials),
+                                                    random.choice(materials.METAL_MATERIALS),
+                                                    adjusted_x, x_scale, True)
+            occluders.extend(occluder_pair)
+        for i in range(num_occluders - num_objects):
+            x_scale = random_real(MIN_OCCLUDER_SCALE, 1, 0.05)
+            limit = 3 - x_scale / 2.0
```
ThomasSchellenbergNextCentury, line 883: Similar to my comment posted above:

> These min/max X positions will ensure that the occluders won't be positioned partially/completely outside of the camera's viewport, and also leave a gap so we can see when an object enters/leaves the scene.

scene_generator/goals_test.py
```diff
@@ -380,8 +380,25 @@ class TestGoal(IntPhysGoal):
...
+    wall_material = random.choice(materials.CEILING_AND_WALL_MATERIALS)
+    obj_list, occluders = goal._get_objects_falling_down(wall_material[0])
+    assert 1 <= len(obj_list) <= 2
+    assert len(obj_list)*2 <= len(occluders) <= 4
+    for obj in obj_list:
+        assert -2.5 <= obj['shows'][0]['position']['x'] <= 2.5
+        assert obj['shows'][0]['position']['y'] == 3.8
+        z = obj['shows'][0]['position']['z']
+        assert z == 1.6 or z == 2.7
+        assert 13 <= obj['shows'][0]['stepBegin'] <= 20
```
ThomasSchellenbergNextCentury, line 404: Open question:  Would it be better to set a random seed and expect specific values, rather than checking for ranges?

scene_generator/goals.py
```diff
@@ -824,9 +812,81 @@ class Position(Enum):
...
+            # occluder (up to 1) so it isn't too close to any of the
+            # others.
+            max_scale = 1
+            for occluder in occluders:
+                distance = abs(occluder['shows'][0]['position']['x'] - x_position)
+                scale = 2 * (distance - occluder['shows'][0]['scale']['x'] / 2.0 - MIN_OCCLUDER_SEPARATION)
+                if scale < 0:
+                    raise GoalException('Placed objects too close together after all')
+                if scale < max_scale:
+                    max_scale = scale
```
ThomasSchellenbergNextCentury, line 874: @chrislong I think we need to do this check in the next loop as well.  We want the `0.5` gap to apply to both the occluders-with-corresponding-objects and the occluders-without-corresponding-objects.

scene_generator/goals.py
```diff
@@ -824,9 +812,81 @@ class Position(Enum):
...
+            x_scale = random_real(min_scale, max_scale, 0.05)
+            adjusted_x = x_position * factor
+            occluder_pair = objects.create_occluder(random.choice(non_wall_materials),
+                                                    random.choice(materials.METAL_MATERIALS),
+                                                    adjusted_x, x_scale, True)
+            occluders.extend(occluder_pair)
+        for i in range(num_occluders - num_objects):
+            x_scale = random_real(MIN_OCCLUDER_SCALE, 1, 0.05)
+            limit = 3 - x_scale / 2.0
+            x_position = random_real(-limit, limit, 0.05)
```
ThomasSchellenbergNextCentury, line 884: We use the `0.05` (minimum random interval) in a lot of places throughout the scene generator.  We could probably make it a global constant.

scene_generator/goals.py
chrislong: Ok, that makes sense. I'll add that.

chrislong:
@ThomasSchellenbergNextCentury Thanks for the comments about the magic numbers. They've all been updated.
Still need to fix the occluder spacing.

ThomasSchellenbergNextCentury:
Whoops, almost forgot to post this...

Here's a comment that probably applies to most of the IntPhys scenes:

In each IntPhys scene containing occluders, the first 12 steps involve moving and rotating the occluders, so the action should start on step 13 at the earliest.  The objects-moving-across-behind-occluders scenes have 60 steps, and the objects-falling-down-behind-occluders scenes have 40.  The last 6 steps of the scene involve moving and rotating the occluders again.  For objects-falling-down-behind-occluders scenes, we reserve 8 steps for falling, and 8 steps for post-falling actions, meaning that the objects can appear and begin falling anytime between steps 13 and 20, inclusive.

scene_generator/goals.py
```diff
@@ -650,6 +650,18 @@ class IntPhysGoal(Goal, ABC):
...
+    # start on step 13 at the earliest. The
+    # objects-moving-across-behind-occluders scenes have 60 steps, and
+    # the objects-falling-down-behind-occluders scenes have 40. The
+    # last 6 steps of the scene involve moving and rotating the
+    # occluders again. For objects-falling-down-behind-occluders
+    # scenes, we reserve 8 steps for falling, and 8 steps for
+    # post-falling actions, meaning that the objects can appear and
+    # begin falling anytime between steps 13 and 20, inclusive.
+    EARLIEST_ACTION_START_STEP = 13
+    LATEST_ACTION_START_STEP = 20
```
ThomasSchellenbergNextCentury, line 664: This is the latest action start step for the objects-falling-down-behind-occluders scenes, but it may be different for other IntPhys scenes.  I'll go ahead and merge this branch though and you can address it in future commits.

PR #68, opened 2020-04-29
MCS-119: IntPhys Scene Generator: Ramp Creation Function

chrislong:


scene_generator/ramps.py
```diff
@@ -0,0 +1,212 @@
...
+        },
+        "scale": {
+            "x": 10,
+            "y": 2,
+            "z": 2.2
+        }
+    }]
+}]
+
+RAMP_30_90_TEMPLATE = [{
```
ThomasSchellenbergNextCentury, line 140: I think the 30/90 and 45/90 ramps are both missing parts to their templates.  They should each have 3 objects.

scene_generator/ramps.py
```diff
@@ -0,0 +1,212 @@
...
+
+RAMP_TEMPLATE_INFO = [(RAMP_30_TEMPLATE, 1), (RAMP_45_TEMPLATE, 3), (RAMP_90_TEMPLATE, 4),
+                      (RAMP_30_90_TEMPLATE, 2), (RAMP_45_90_TEMPLATE, 3)]
+
+def create_ramp(material_string, x_position_percent, left_to_right = False):
+    """Create a ramp of a random type. Returns a list of objects that make
+    up the ramp."""
+    template_info = random.choice(RAMP_TEMPLATE_INFO)
+    ramp = []
+    x_term = x_position_percent * template_info[1]
```
ThomasSchellenbergNextCentury, line 202: Do we want to assume that the given `x_position_percent` is always between 0 and 1, or should we do some validation here?

chrislong: Ok, sure, I'll add an assertion

scene_generator/ramps.py
chrislong: Weird, I thought I cut & pasted straight from the ticket. I'll double-check.

scene_generator/ramps.py
```diff
@@ -0,0 +1,212 @@
...
+        "scale": {
+            "x": 10,
+            "y": 2,
+            "z": 2.2
+        }
+    }]
+}]
+
+RAMP_TEMPLATE_INFO = [(RAMP_30_TEMPLATE, 1), (RAMP_45_TEMPLATE, 3), (RAMP_90_TEMPLATE, 4),
+                      (RAMP_30_90_TEMPLATE, 2), (RAMP_45_90_TEMPLATE, 3)]
```
ThomasSchellenbergNextCentury, line 295: You can add a comment here (if it makes sense):

Each type of ramp has a specific amount of variation allowed in its final X position based on keeping the ramp and some space at the bottom of the ramp within the camera's viewport.

scene_generator/ramps.py
```diff
@@ -0,0 +1,212 @@
...
+    "id": "ramp_part2_",
+    "type": "cube",
+    "materials": [],
+    "kinematic": True,
+    "structure": True,
+    "mass": 100,
+    "shows": [{
+        "stepBegin": 0,
+        "position": {
+            "x": 4,
```
ThomasSchellenbergNextCentury, line 44: I made a mistake on the 30-degree-ramp-part-2: the `shows.position.x` should be `6`, not `4`.

PR #69, opened 2020-04-29
Mcs 118: IntPhys Scene Generator: Occluders for Objects Moving Across

chrislong:
Probably want to merge PR #64 first.


scene_generator/goals.py
```diff
@@ -36,6 +36,50 @@
 WALL_PROBS = [60, 20, 10, 10]
 
 
+def random_real(a, b, step):
```
brianpippin, line 39: This function is also in MCS-117 PR watch for merge conflicts

chrislong: Yeah, I thought I'd test to see how smart git is. (Actually just needed the function for code in both tickets.)

scene_generator/goals.py
```diff
@@ -36,6 +36,50 @@
...
+    """Return a random real number N where a <= N <= b and N - a is divisible by step."""
+    steps = int((b - a) / step)
+    n = random.randint(0, steps)
+    return a + n * step
+
+
+def float_range(*args):
```
ThomasSchellenbergNextCentury, line 46: Where is the float_range function being used?

scene_generator/goals.py
```diff
@@ -36,6 +36,50 @@
...
+            raise TypeError(f'floatRange only accepts float and integer arguments. ({type(num)} : {num} given)')
+    count = 0
+    value = start
+    while value < stop:
+        yield value
+        count += 1
+        value = start + step*count
+    return
+
+def all_items_less_equal(a, b):
```
ThomasSchellenbergNextCentury, line 72: I don't see all_items_less_equal used anywhere either.

scene_generator/goals.py
chrislong: Oops, it's not any more. I'll take it out.

scene_generator/goals.py
chrislong: Yup, good catch, I'll nuke it, too.

scene_generator/goals.py
```diff
@@ -709,7 +709,68 @@ def compute_objects(self, wall_material_name):
...
+        non_wall_materials = [m for m in materials.CEILING_AND_WALL_MATERIALS
+                              if m[0] != wall_material_name]
+        occluder_list = []
+        for i in range(num_occluders):
+            occluder_fits = False
+            # try random position and scale until we find one that fits (or try too many times)
+            for occluder_try in range(IntPhysGoal.MAX_OCCLUDER_TRIES):
+                if i < num_paired_occluders:
+                    paired_obj = obj_list[i]
+                    min_scale = min(max(paired_obj['shows'][0]['scale']['x'], 0.25), 1)
```
ThomasSchellenbergNextCentury, line 733: I think `0.25` is the MIN_OCCLUDER_SCALE

ThomasSchellenbergNextCentury: And `1` is the MAX_OCCLUDER_SCALE

scene_generator/goals.py
```diff
@@ -709,7 +709,68 @@ def compute_objects(self, wall_material_name):
...
+                    position_by_step = paired_obj['intphys_options']['position_by_step']
+                    position_index = random.randrange(len(position_by_step))
+                    paired_x = position_by_step[position_index]
+                    paired_z = paired_obj['shows'][0]['position']['z']
+                    if paired_z == 1.6:
+                        occluder_x = paired_x * 0.9
+                    elif paired_z == 2.7:
+                        occluder_x = paired_x * 0.8
```
ThomasSchellenbergNextCentury, line 741: Global variables already exist for `1.6`, `2.7`, `0.9`, and `0.8` that you can use.

scene_generator/goals.py
```diff
@@ -709,7 +709,68 @@ def compute_objects(self, wall_material_name):
...
+                    else:
+                        logging.warning(f'Unsupported z for occluder target "{paired_obj["id"]}": {paired_z}')
+                        occluder_x = paired_x
+                else:
+                    min_scale = 0.25
+                    occluder_x = None
+                x_scale = random_real(min_scale, 1.0, 0.05)
+                if occluder_x is None:
+                    limit = 3.0 - x_scale / 2.0
```
ThomasSchellenbergNextCentury, line 750: This is a simple calculation but it's also used in `_get_objects_falling_down` so you might want to move it to its own function for better clarity.

scene_generator/goals.py
```diff
@@ -709,7 +709,68 @@ def compute_objects(self, wall_material_name):
...
+                        occluder_x = paired_x * 0.9
+                    elif paired_z == 2.7:
+                        occluder_x = paired_x * 0.8
+                    else:
+                        logging.warning(f'Unsupported z for occluder target "{paired_obj["id"]}": {paired_z}')
+                        occluder_x = paired_x
+                else:
+                    min_scale = 0.25
+                    occluder_x = None
+                x_scale = random_real(min_scale, 1.0, 0.05)
```
ThomasSchellenbergNextCentury, line 748: I think these are all now global variables as well.

scene_generator/goals.py
```diff
@@ -709,7 +709,68 @@ def compute_objects(self, wall_material_name):
...
+                if occluder_x is None:
+                    limit = 3.0 - x_scale / 2.0
+                    limit = int(limit / 0.05) * 0.05
```
ThomasSchellenbergNextCentury, line 751: Why do we need to do this to the limit?  And we here but not in `_get_objects_falling_down`?

scene_generator/goals.py
```diff
@@ -709,7 +709,68 @@ def compute_objects(self, wall_material_name):
...
+                    if geometry.occluders_too_close(other_occluder, occluder_x, x_scale):
+                        found_collision = True
+                        break
+                if not found_collision:
+                    occluder_fits = True
+                    break
+            if occluder_fits:
+                occluder_objs = objects.create_occluder(random.choice(non_wall_materials),
+                                                        random.choice(materials.METAL_MATERIALS),
+                                                        occluder_x, x_scale)
```
ThomasSchellenbergNextCentury, line 764: This logic checks the gap between occluders, right?  Could some of this be combined with the logic in `_get_objects_falling_down`?  They're doing the same thing, right?

scene_generator/goals.py
```diff
@@ -709,7 +709,68 @@ def compute_objects(self, wall_material_name):
...
+        """Get occluders to for objects in obj_list."""
+        num_occluders = self._get_num_occluders()
+        num_paired_occluders = self._get_num_paired_occluders()
+        non_wall_materials = [m for m in materials.CEILING_AND_WALL_MATERIALS
+                              if m[0] != wall_material_name]
+        occluder_list = []
+        for i in range(num_occluders):
+            occluder_fits = False
+            # try random position and scale until we find one that fits (or try too many times)
+            for occluder_try in range(IntPhysGoal.MAX_OCCLUDER_TRIES):
```
ThomasSchellenbergNextCentury, line 730: You can add a comment here:

We want to position each paired occluder at the same X position that its corresponding object will be at the end/start of a random step during its movement across the scene described by position_by_step.  This will let us add an implausible event (make the object disappear, teleport it, or replace it with another object) at that specific step.

scene_generator/goals.py
chrislong: Yeah, those constants didn't exist when I wrote this originally. I'll update it, though.

scene_generator/goals.py
chrislong: I'm doing it here because I thought we wanted the values to all align with 0.05. It's not in the other place because I didn't think of it.

scene_generator/goals.py
chrislong: With the occluders that aren't paired with a target, yes, they are. I'll see about factoring that out.

scene_generator/goals.py
ThomasSchellenbergNextCentury: Ah, OK.  Makes sense.

scene_generator/goals.py
ThomasSchellenbergNextCentury: Thanks.

chrislong: No problem. Definitely made it cleaner, and probably less error-prone.

PR #70, opened 2020-04-30
MCS-54: Save scene history to file

brianpippin:


python_api/machine_common_sense/mcs_controller_ai2thor.py
```diff
@@ -123,7 +133,9 @@ def start_scene(self, config_data):
 
         self.__scene_configuration = config_data
         self.__step_number = 0
+        self.__history_list = []
         self.__goal = self.retrieve_goal(self.__scene_configuration)
+        self.__scene_file = self.HISTORY_DIRECTORY + self.__scene_configuration['name'].replace('.json','') + "-" + datetime.datetime.now().strftime("%Y%m%d-%H%M%S") + ".txt"
```
chrislong, line 138: It's better to build path names with `os.path.join`. (Also, you don't need to worry about including `/`s in directory names, then.)

python_api/machine_common_sense/mcs_controller_ai2thor.py
```diff
@@ -111,8 +114,15 @@ def on_init(self, debug=False, enable_noise=False):
...
         self.__goal = None
 
+        if not os.path.exists(self.HISTORY_DIRECTORY):
+            os.makedirs(self.HISTORY_DIRECTORY)
+
     # Override
     def end_scene(self, classification, confidence):
+        self.__history_list.append({"classification": classification, "confidence": confidence})
+        historyFile = open(self.__scene_file, "w")
+        historyFile.write('\n'.join(json.dumps(history) for history in self.__history_list))
```
chrislong, line 124: It's better to use `with` because then you can't forget to close the file. Like
```with open(...) as historyFile:
    historyFile.write(...)
```

python_api/machine_common_sense/mcs_controller_ai2thor.py
deanwetherby: Alternatively, take a look at pathlib

brianpippin: I will look at both, I appreciate how code reviews help me learn new things about python.

python_api/machine_common_sense/mcs_controller_ai2thor.py
brianpippin: Oh that's cool, let me read up on that too, I will look into changing that.

python_api/machine_common_sense/mcs_controller_ai2thor.py
```diff
@@ -294,6 +306,11 @@ def step(self, action, **kwargs):
         # Only call mcs_action_to_ai2thor_action AFTER calling validate_and_convert_params
         action = self.mcs_action_to_ai2thor_action(action)
 
+        self.__history_list.append({"step": self.__step_number, "action": action, "args": kwargs, "params": params})
+
+        historyFile = open(self.__scene_file, "w")
+        historyFile.write('\n'.join(json.dumps(history) for history in self.__history_list))
```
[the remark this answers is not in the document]
brianpippin: I was actually thinking about moving that to its own file since I'm using it twice.

python_api/machine_common_sense/mcs_controller_ai2thor.py
```diff
@@ -123,7 +133,9 @@ def start_scene(self, config_data):
 
         self.__scene_configuration = config_data
         self.__step_number = 0
+        self.__history_list = []
```
ThomasSchellenbergNextCentury, line 136: We should add a check here in case they've called `start_scene` without calling `end_scene` on the previous scene so we don't lose the previous scene's history.

python_api/machine_common_sense/mcs_controller_ai2thor.py
```diff
@@ -111,8 +114,15 @@ def on_init(self, debug=False, enable_noise=False):
         self.__step_number = 0
         self.__goal = None
 
+        if not os.path.exists(self.HISTORY_DIRECTORY):
+            os.makedirs(self.HISTORY_DIRECTORY)
+
     # Override
     def end_scene(self, classification, confidence):
+        self.__history_list.append({"classification": classification, "confidence": confidence})
```
ThomasSchellenbergNextCentury, line 122: Would it make sense to move the `history_list` and the act of saving the `historyFile` into the `mcs_controller` superclass?  The exact structure of the elements that we append to the `history_list` may be specific to AI2-THOR, but the act of saving the history seems more generic to me.

python_api/machine_common_sense/mcs_controller_ai2thor.py
ThomasSchellenbergNextCentury: Also, this is nitpicky, but your naming scheme is inconsistent (snake case vs. camel case).  Snake case is usually the default for Python code.

python_api/machine_common_sense/mcs_controller_ai2thor.py
ThomasSchellenbergNextCentury: You're always so positive, Brian!  :)

python_api/machine_common_sense/mcs_controller_ai2thor.py
brianpippin: It doesn't matter if they call start_scene without calling end_scene because I write to the file each time they call step.  And if they called start scene with the same file, I create a new timestamp on the end so it would just create a new file.

python_api/machine_common_sense/mcs_controller_ai2thor.py
brianpippin: My problem is that I'm so used to using camel case but when I see the python conventions I usually try to follow it, but the came case keeps seeping back in.

python_api/machine_common_sense/mcs_controller_ai2thor.py
```diff
@@ -117,11 +117,16 @@ def on_init(self, debug=False, enable_noise=False):
         if not os.path.exists(self.HISTORY_DIRECTORY):
             os.makedirs(self.HISTORY_DIRECTORY)
 
+    # Write the history file
+    def write_history_file(self, history_item):
+        with open(self.__scene_file, "a+") as history_file:
```
chrislong, line 122: I think this could be just `"a"` instead of `"a+"` since you're not reading the file, only appending.

python_api/machine_common_sense/mcs_controller_ai2thor.py
```diff
@@ -306,10 +311,9 @@ def step(self, action, **kwargs):
         # Only call mcs_action_to_ai2thor_action AFTER calling validate_and_convert_params
         action = self.mcs_action_to_ai2thor_action(action)
 
-        self.__history_list.append({"step": self.__step_number, "action": action, "args": kwargs, "params": params})
-
-        historyFile = open(self.__scene_file, "w")
-        historyFile.write('\n'.join(json.dumps(history) for history in self.__history_list))
+        history_item = {"step": self.__step_number, "action": action, "args": kwargs, "params": params}
+        self.__history_list.append(history_item)
+        self.write_history_file(history_item)
```
chrislong, line 316: I'm not sure if performance is a concern, but if it is, you might want to refactor to not open and close the file every step, but keep it open, instead. That does risk losing data if the process is killed, though, and requires more work to make sure you close it when you're done. So maybe not worth it, I just thought I'd point it out in case you need more performance here.

PR #71, opened 2020-04-30
Mcs 120: IntPhys Scene Generator: Scenery

chrislong:


scene_generator/goals.py
```diff
@@ -698,15 +705,41 @@ def get_config(self, goal_objects):
...
+
         return goal
 
     def generate_walls(self, material, performer_position, bounding_rects):
         """IntPhys goals have no walls."""
         return []
 
+    def _compute_scenery(self):
+        def random_x():
+            return random_real(-6.5, 6.5, 0.05)
```
ThomasSchellenbergNextCentury, line 719: The `-6.5` and `6.5` are the max X positions in each direction that the scenery objects should still be visible in the camera's viewport.  You could make `6.5` a global variable and reuse the one you already have for `0.05`

scene_generator/goals.py
```diff
@@ -698,15 +705,41 @@ def get_config(self, goal_objects):
...
+
+        def random_z():
+            return random_real(3.25, 4.95, 0.05)
```
ThomasSchellenbergNextCentury, line 722: The `3.25` and `4.95` will ensure that the scenery objects will be positioned between the moving IntPhys objects and the room's wall.

chrislong:
Addressed comments, ready for re-review.

PR #74, opened 2020-05-05
Mcs 164: [Pathfinding] Target Objects with locationParent Properties

chrislong:
Also includes a bunch of type hinting.


scene_generator/goals.py
```diff
@@ -205,10 +205,75 @@ def find_image_for_object(object_def):
...
+                "action": "MoveAhead",
+                "params": {}
+                }]*int(whole))
+    actions.append({
+            "action": "MoveAhead",
+            "params": {
+                "amount": round(frac,POSITION_DIGITS)
+                }
+        })
+    return actions, current_heading
```
ThomasSchellenbergNextCentury, line 248: Some of the dict indentation here is a bit weird.

scene_generator/goals.py
```diff
@@ -205,10 +205,75 @@ def find_image_for_object(object_def):
...
+    current_heading. Returns a tuple: (list of actions, new heading)"""
+    index = 1
+    actions = []
+    dx = path_section[1][0]-path_section[0][0]
+    dz = path_section[1][1]-path_section[0][1]
+    theta = math.degrees(math.atan2(dx,dz))
+
+        #IF my calculations are correct, this should be right no matter what
+        # I'm assuming a positive angle is a clockwise rotation- so this should work
+        #I think
```
ThomasSchellenbergNextCentury, line 223: Yes, positive angle is clockwise rotation.

scene_generator/goals.py
chrislong: So it is. Fixed now.

scene_generator/goals.py
```diff
@@ -411,6 +411,21 @@ def compute_objects(self):
...
+        """Get the action sequence for going from performer start to the goal_object."""
+        performer = (self._performer_start['position']['x'], self._performer_start['position']['z'])
+        goal = (goal_object['shows'][0]['position']['x'], goal_object['shows'][0]['position']['z'])
+        hole_rects = []
+        hole_rects.extend(object['shows'][0]['bounding_box'] for object in all_objects if object['id'] != goal_object['id'])
+        path = generatepath(performer, goal, hole_rects)
+
+        actions = []
+        current_heading = self._performer_start['rotation']['y']
+        for indx in range(len(path)-1):
```
deanwetherby: It'd be cool if 'path' was iterable.

scene_generator/goals.py
```diff
@@ -200,6 +200,69 @@ def find_image_name(target):
...
+                }
+        })
+    return actions
+
+
+def get_navigation_actions(start_location, goal_object, all_objects):
+    """Get the action sequence for going from performer start to the goal_object."""
+    performer = (start_location['position']['x'], start_location['position']['z'])
+    if 'locationParent' in goal_object:
+        parent = next((obj for obj in all_objects if obj['id'] == goal_object['locationParent']))
```
deanwetherby, line 262: I think you need to provide a default for next.
        parent = next((obj for obj in all_objects if obj['id'] == goal_object['locationParent'])**, None**)

scene_generator/goals.py
chrislong: I didn't write this, but I'm not sure how that would help, since it needs to call parse_path_section with a range of points.

scene_generator/goals.py
chrislong: Good idea.

scene_generator/goals.py
deanwetherby: You are right. Any time I see range(len(thing)) it makes me think that it should be iterable but you are using ranges of that thing below.

brianpippin:
Everything looks good on this I think, but I think we should consider adding another ticket for breaking up the goals.py file, it is getting quite large and can be difficult to track everything in it, perhaps we need some sub class files for specific things, and I know we could use a utility file as well.

PR #75, opened 2020-05-05
MCS-188: IntPhys Scene Generator: Collision

chrislong:
use acceleration to check for possible collision between objects


deanwetherby:
Looks fine. Are there any corresponding unit test(s)?

chrislong:
No, but that's a good idea. I'll write one

PR #77, opened 2020-05-06
MCS-178 Playroom docker file

deanwetherby:
Draft PR to discuss Dockerization approach. We can copy in the MCS repo which would allow for local changes or we can pull directly from git. There are advantages to both.


Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+#
+# Run playroom with preloaded scene
+# docker run --gpus all -e DISPLAY -v /tmp/.X11-unix:/tmp/.X11-unix:rw mcs-playroom:0.0.6 \
+#  python3.6 /mcs/python_api/machine_common_sense/run_mcs_just_pass.py /mcs/MCS-AI2-THOR-Unity-App-v0.0.6.x86_64 /mcs/python_api/scenes/playroom.json
+#
+# Run with scenes provided through volume mapping
+# docker run --gpus all -e DISPLAY -v /home/HQ/dwetherby/Downloads/interaction_scenes:/scenes -v /tmp/.X11-unix:/tmp/.X11-unix:rw mcs-playroom:0.0.6 \
+#  python3.6 /mcs/python_api/machine_common_sense/run_mcs_just_rotate.py /mcs/MCS-AI2-THOR-Unity-App-v0.0.6.x86_64 /scenes/retrieval/retrieval_goal-0002.json
+
+FROM nvidia/opengl:1.0-glvnd-runtime-ubuntu16.04
```
chrislong, line 19: Why not 18.04? Then I don't think you'd need extra ppas for python.

deanwetherby: Easy enough to update which would make the python installation trivial. Wasn't sure we wanted an 18.04 docker container if no one is running 18.04. I'm still on 16.04. Maybe I should do an OS poll.

Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+# Run interactive bash shell
+# docker run -it --rm --gpus all -e DISPLAY -v /tmp/.X11-unix:/tmp/.X11-unix:rw mcs-playroom:0.0.6
+#
+# Run playroom with preloaded scene
+# docker run --gpus all -e DISPLAY -v /tmp/.X11-unix:/tmp/.X11-unix:rw mcs-playroom:0.0.6 \
+#  python3.6 /mcs/python_api/machine_common_sense/run_mcs_just_pass.py /mcs/MCS-AI2-THOR-Unity-App-v0.0.6.x86_64 /mcs/python_api/scenes/playroom.json
+#
+# Run with scenes provided through volume mapping
+# docker run --gpus all -e DISPLAY -v /home/HQ/dwetherby/Downloads/interaction_scenes:/scenes -v /tmp/.X11-unix:/tmp/.X11-unix:rw mcs-playroom:0.0.6 \
+#  python3.6 /mcs/python_api/machine_common_sense/run_mcs_just_rotate.py /mcs/MCS-AI2-THOR-Unity-App-v0.0.6.x86_64 /scenes/retrieval/retrieval_goal-0002.json
```
ThomasSchellenbergNextCentury, line 17: We'll want to add a section in the `python_api/README.md` as well.

Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+
+WORKDIR /mcs
+
+RUN apt-get update -y&& \
+    apt-get install -y software-properties-common && \
+    add-apt-repository ppa:deadsnakes/ppa
+
+RUN apt-get update -y && \
+    apt-get install -y xvfb mesa-utils python3.6 python3-pip && \
+    python3.6 -m pip install --upgrade pip setuptools wheel
```
ThomasSchellenbergNextCentury, line 31: We should note in our documentation that we're assuming everyone's using Python 3.6

Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+COPY . /mcs
+RUN python3.6 -m pip install .
+# or we could install from MCS tagged branch rather than copying into the container
+# RUN pip install git+https://github.com/NextCenturyCorporation/MCS@latest
+
+# add ai2thor/unity resources
+ADD https://github.com/NextCenturyCorporation/MCS/releases/download/0.0.6/MCS-AI2-THOR-Unity-App-v0.0.6.x86_64 /mcs
+ADD https://github.com/NextCenturyCorporation/MCS/releases/download/0.0.6/MCS-AI2-THOR-Unity-App-v0.0.6_Data.tar.gz /mcs
+RUN tar -xzvf /mcs/MCS-AI2-THOR-Unity-App-v0.0.6_Data.tar.gz -C /mcs && \
+    chmod a+x /mcs/MCS-AI2-THOR-Unity-App-v0.0.6.x86_64 && \
```
ThomasSchellenbergNextCentury, line 43: Can we make the version into a variable so we only need to update it in one place?  Also, I'm not sure if it's worth noting in our documentation that this container will require internet access to download the resources.

ThomasSchellenbergNextCentury:
@deanwetherby So do the TA1s just copy this Dockerfile into their code repository?  How does this work?

deanwetherby:
@ThomasSchellenbergNextCentury Great questions. Short answer is I don't know yet. We'll have to iterate towards perfection.

Dockerfile
deanwetherby: We mention 3.6 in python_api/machine_common_sense/README.md and python_api/test/README.md already. But yes, we should be explicit. We are also using f-strings in a few places so we ARE requiring 3.6+.

We should add "python_requires='>=3.6'" to the setup.py now that I think about it.

Dockerfile
deanwetherby: Agree this would be a good idea. I'll work on a docker variable and string substitution for those lines. Alternatively, we could pass that value in as an argument. We'll see which is better.

deanwetherby: So now do we pull the tagged branch from git instead of copying in the MCS folder? It sounds like the right thing to do but then we lose the run_mcs*.py scripts and other utilities.

deanwetherby: I was thinking of doing pip install from git in the last message. If we pull the whole tagged repo, we can still get everything.

Dockerfile
ThomasSchellenbergNextCentury: Sounds good.

Dockerfile
ThomasSchellenbergNextCentury: I don't think the utility scripts are needed for the Docker container.  But whatever you think is best.

deanwetherby:
The good news is that with the latest changes, the docker image dropped from 8.5 GB to 5.75 GB.

deanwetherby:
How would we determine the maximum number of steps between the intphys and interaction tests? intphys is fixed to 50 right? Then we have to determine a reasonable max for interaction and we can use the test type to set this limit before exiting.

template.Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+# Running the playroom using volume mapping scene files into /scenes
+# $ docker run -t --gpus all -e DISPLAY -v /tmp/.X11-unix:/tmp/.X11-unix:rw \
+#   -v /Downloads/interaction_scenes:/scenes mcs-submission:0.0.1 \
+#   python3.8 /mcs/playroom.py /mcs/MCS-AI2-THOR-Unity-App-v0.0.6.x86_64 /scenes/retrieval/retrieval_goal-0001.json
+
+# This is a multi-stage build based upon the MCS playroom base image
+FROM mcs-playroom:0.0.6
+
+# pathlib is used in the playroom file to find the unity executable
+RUN python3.8 -m pip install pathlib
```
chrislong, line 19: There's a pathlib module that's part of the standard Python library. Is this different but named the same?

template_playroom.py
```diff
@@ -0,0 +1,63 @@
...
+        # TEMPLATE: break out of loop if goal achieved
+
+
+def main():
+    args = parse_args()
+
+    unity_exe_path = find_unity_executable()
+    if unity_exe_path is None:
+        print("Unity executable not found in /mcs")
+        exit()
```
chrislong, line 51: Ideally you'd print error messages to stderr and exit with non-0 exit code on error.

template.Dockerfile
deanwetherby: Hmm. I didn't realize that it was part of the standard library.

ThomasSchellenbergNextCentury:
@deanwetherby The IntPhys scenes end at either 40, 50, or 60 steps, depending on the goal type and setup.  These numbers were chosen based on how long it took objects to roll across the scene at their slowest speed.  I think we should choose much higher numbers for the interaction scenes (maybe something like 150) since they'll usually require exploration of the scene which could take a long time.

template_playroom.py
```diff
@@ -28,14 +30,14 @@ def determine_max_steps(scene):
     goal = scene['goal']
     max_steps = 0
     if 'interaction' in goal['type_list']:
-        max_steps = 200
+        max_steps = MAX_ALLOWED_INTERACTION_STEPS
     elif 'intphys' in type_list:
```
chrislong, line 34: Should `type_list` be `goal['type_list']`?

deanwetherby: yes. thank you

PR #78, opened 2020-05-06
MCS-198: Delete Some Object Properties from Scene Generator Output

chrislong:


scene_generator/scene_generator.py
```diff
@@ -39,13 +39,13 @@
...
-def load_object_file(object_file_name):
-    with open(object_file_name) as object_file:
-        objects = json.load(object_file)
-    return objects
+def clean_object(obj):
+    """Remove properties we do not want TA1s to have access to."""
+    obj.pop('dimensions', None)
+    obj.pop('intphys_option', None)
+    if 'shows' in obj:
+        obj['shows'].pop('bounding_box', None)
```
ThomasSchellenbergNextCentury, line 48: I didn't know about the pop function.  Neat.

scene_generator/scene_generator.py
```diff
@@ -62,16 +62,15 @@ def generate_file(name, goal_type, find_path):
 
     # Use PrettyJsonNoIndent on some of the lists and dicts in the output body because the indentation from the normal
     # Python JSON module spaces them out far too much.
-    if 'goal' in body:
```
ThomasSchellenbergNextCentury, line 65: Do all scenes have goals now?  I think that will stay true moving forward.

chrislong: It's in the template, so I thought it would always be there.

PR #81, opened 2020-05-07
Mcs 209: [Critical] Minor IntPhys Goal Scene Generator Updates

chrislong:


scene_generator/goals.py
```diff
@@ -81,10 +90,15 @@ def instantiate_object(object_def, object_location):
     for attribute in object_def['attributes']:
         new_object[attribute] = True
 
+    new_object['torques'] = [DEFAULT_TORQUE]
```
ThomasSchellenbergNextCentury, line 93: @chrislong This should be only for IntPhys goal objects.

chrislong: Ok, I'll fix it.

PR #82, opened 2020-05-07
Mcs 200: Add Visible to Goal Type List

chrislong:
Also fixes a bug in scene_generator


scene_generator/set_target_visibility.py
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env python3
+#
+import argparse
+import os
+import sys
+import logging
+
+from machine_common_sense.mcs import MCS
```
deanwetherby, line 8: It feels like we've gone a little overboard on the MCS marketing/namespacing. There's nothing to be done right now, but we might have to rethink this at some point. Maybe something numpy-like:
mcs.Goal, mcs.Reward, mcs.Controller

brianpippin:
Nothing stood out to me.

scene_generator/set_target_visibility.py
ThomasSchellenbergNextCentury: Yeah, that's fine.  Rachel had also requested that our C# classes be renamed from MachineCommonSense* to MCS*.

scene_generator/set_target_visibility.py
```diff
@@ -0,0 +1,85 @@
...
+    target_ids = [metadata[key]['id'] for key in ('target', 'target_1', 'target_2') if key in metadata]
+    logging.debug(f'target_ids: {target_ids}')
+    logging.debug(f'num objects = {len(output.object_list)}')
+    
+    targets_visible = False
+    targets_not_visible = False
+    for obj in output.object_list:
+        if obj.uuid in target_ids:
+            # If it's in the object list to begin with, it should be visible, but just to make sure...
+            if obj.visible:
```
ThomasSchellenbergNextCentury, line 37: It's good that you check here, in case we change the behavior for some reason in the future.

scene_generator/set_target_visibility.py
```diff
@@ -0,0 +1,85 @@
...
+
+    metadata = config_data['goal']['metadata']
+    target_ids = [metadata[key]['id'] for key in ('target', 'target_1', 'target_2') if key in metadata]
+    logging.debug(f'target_ids: {target_ids}')
+    logging.debug(f'num objects = {len(output.object_list)}')
+    
+    targets_visible = False
+    targets_not_visible = False
+    for obj in output.object_list:
+        if obj.uuid in target_ids:
```
ThomasSchellenbergNextCentury, line 35: Don't we need to add an else here that also sets targets_not_visible to True?

scene_generator/set_target_visibility.py
```diff
@@ -0,0 +1,85 @@
...
+        os.remove(backup_name)
+
+
+def main(argv):
+    parser = argparse.ArgumentParser(description='Update scene descriptions with target visibility info')
+    parser.add_argument('--app', metavar='UNITY_MCS_APP', required=True, help='Path to Unity MCS application')
+    parser.add_argument('--debug', action='store_true', default=False, help='Enable debugging output')
+    parser.add_argument('--backup', action='store_true', default=False, help='Keep backup files')
+    parser.add_argument('--loglevel', choices=LOG_LEVELS, help='set logging level')
+    parser.add_argument('scene_files', metavar='SCENE_FILE', nargs='+', help='scene file to update')
```
ThomasSchellenbergNextCentury, line 69: Out of curiosity, why does scene_files not start with a double-dash?  I'm assuming it's a Python standard with which I'm not familiar.

chrislong: It's an argparse thing. Those are the positional arguments, which don't have any '--foo' argument. Basically, all the leftover stuff is put in there. If you run it with --help that should clarify the usage.

scene_generator/set_target_visibility.py
chrislong: An `else` for this would be for objects that aren't targets. Are we supposed to do something with those?

ThomasSchellenbergNextCentury: Oh sorry I misread.

I meant to say that we need to check and see if there are any target_ids that are NOT in the output.object_list.

chrislong: That is checked, but I guess it's kind of hidden. L39 removes visible targets from `target_ids` and then on L48 it checks if there are any targets left (i.e., that weren't found in the `object_list`.

ThomasSchellenbergNextCentury: Oh, you're right, my bad.  Must need some more coffee.

PR #83, opened 2020-05-07
MCS-210 Improved reward calculations

deanwetherby:
Retrieval has object in hand.
Traversal uses object['distanceXZ'] < MAX_REACH_DISTANCE.
Transferral makes an XZ planar assumption. Goal object bounds are projected to the XZ plane and a convex hull is used to create polygons for both goal objects. Nearness is calculated using shapely polygon distance [poly1.distance(poly2)]. Stacking uses polygon intersections on the XZ plane [poly1.intersects(poly2)] along with the original objects y positioning for stack order.


deanwetherby:
Rerunning scene timing to ensure we didn't slow down drastically.

python_api/machine_common_sense/mcs_reward.py
```diff
@@ -30,54 +30,22 @@ def __get_object_from_list(objects: List[Dict],
         return next((o for o in objects if o['objectId'] == target_id), None)
 
     @staticmethod
-    def _convert_bounds_to_polygons(goal_object: Dict) -> Tuple[geometry.Polygon, geometry.Polygon]:
+    def _convert_object_to_planar_polygon(goal_object: Dict) -> geometry.Polygon:
         '''
-        Converts goal object bounds to an upper and lower planar polygon.
+        Project goal object bounds (x,y,z) to an XZ planar polygon.
```
chrislong, line 35: It's a convex polygon, right?

deanwetherby: Yes. From shapely docs:
object.convex_hull
  Returns a representation of the smallest convex Polygon containing all the points in the object ...

python_api/test/test_mcs_reward.py
```diff
@@ -85,78 +85,31 @@ def test_convert_object_bounds_to_polygons(self):
         goal_object['objectBounds']['objectBoundsCorners'].append({'x':1.0, 'y': 1.0, 'z': 1.0})
         goal_object['objectBounds']['objectBoundsCorners'].append({'x':0.0, 'y': 1.0, 'z': 1.0})
 
-        lower, upper = MCS_Reward._convert_bounds_to_polygons(goal_object)
+        polygon = MCS_Reward._convert_object_to_planar_polygon(goal_object)
+        
+        self.assertIsInstance(polygon, geometry.Polygon)
+        self.assertEqual(len(list(polygon.exterior.coords)), 5)
```
chrislong, line 91: Does this mean the polygon will always have 5 vertices? Why?

deanwetherby:
Phew.
Transferral
RAN 1080 SCENES WITH 21600 TOTAL STEPS IN **4861.9295** SECONDS
Average single step took 0.2191 seconds
Longest single step took 0.4214 seconds
Shortest single step took 0.1619 seconds
Average single scene took 4.5018 seconds
Longest single scene took 5.6126 seconds
Shortest single scene took 3.4417 seconds

PR #84, opened 2020-05-08
Mcs 202: Generate Two Separate Scene Configuration Files

chrislong:


scene_generator/scene_generator.py
```diff
@@ -65,7 +66,8 @@ def generate_file(name, goal_type, find_path):
     goal_obj = goals.choose_goal(goal_type)
     goal_obj.update_body(body, find_path)
 
-    write_scene(name + '-debug', body)
+    debug_name = name[:-5] + '-debug.json'
```
deanwetherby, line 83: Could use pathlib stem  or os.path.splitext here to get the file basename.

chrislong: Yeah, normally I would do something like that, but in this case it's required to end in `.json`.

scene_generator/scene_generator.py
```diff
@@ -40,12 +40,26 @@
...
+
+
 def clean_object(obj):
     """Remove properties we do not want TA1s to have access to."""
+    obj.pop('info', None)
     obj.pop('dimensions', None)
     obj.pop('intphys_option', None)
     if 'shows' in obj:
-        obj['shows'].pop('bounding_box', None)
+        obj['shows'][0].pop('bounding_box', None)
```
brianpippin, line 62: This change is another code review I think?

chrislong: Yeah, it is. I wanted the tests to pass.

PR #89, opened 2020-05-13
MCS-162: [Pathfinding] Detect Impossible Paths

chrislong:
* accounts for the width of the performer
* catches impossible path errors and tries generating the scene afresh if no path can be found
* works around a serious performance bug in the path finding library

Scene generation with path finding takes about 0.1 seconds per scene.


scene_generator/optimal_path.py
```diff
@@ -1,44 +1,96 @@
...
+    poly_coords = [list(poly.exterior.coords) for poly in unified_polygons]
+    # The polygons we get back from unary_union can have the same
+    # first and last point, which the shortest path algorithm doesn't
+    # like.
+    num_unified_points = 0
+    for coords in poly_coords:
+        num_unified_points += len(coords)
+        if coords[0] == coords[-1]:
+            del coords[-1]
+    logging.debug(f'unified polygons: {len(poly_coords)}\tunified points: {num_unified_points}')
```
deanwetherby, line 65: Function is pretty long but there are some clearly defined functions in here like unifying and dilating that can be pulled out into separate functions. Then we can write docstrings and unit tests against each of those functions which will help with clarity.

scene_generator/optimal_path.py
```diff
@@ -1,44 +1,96 @@
...
-        environment = Environment()
-        environment.store(boundary_coordinates, holes, validate = False)
+        environment.store(boundary_coordinates, poly_coords, validate=True)
         environment.prepare()
-        path,length = environment.find_shortest_path(source_loc, target_loc)
+        path, length = environment.find_shortest_path(source_loc, target_loc)
+        # Work around a bug in the DirectedHeuristicGraph class
+        # defined by the extremitypathfinder module. It causes the
+        # graph to keep growing across successive Environments.
+        environment.graph.all_nodes.clear()
```
deanwetherby, line 75: Just wondering if this is worth a bug report or even a PR with the original repo.

chrislong: Good idea. I filed a bug report and put a reference to it in this comment.

scene_generator/goals.py
```diff
@@ -606,7 +610,11 @@ def find_optimal_path(self, goal_objects: List[Dict[str, Any]], all_objects: Lis
         else:
             target = (goal_objects[0]['shows'][0]['position']['x'], goal_objects[0]['shows'][0]['position']['z'])
         goal = (goal_objects[1]['shows'][0]['position']['x'], goal_objects[1]['shows'][0]['position']['z'])
+        logging.debug(f'TransGoal.f_o_p: target = {target}\tgoal = {goal}\tholes = {hole_rects}')
         path = generatepath(target, goal, hole_rects)
+        if path is None:
+            raise GoalException('could not find path from target object to goal')
+        logging.debug(f'TransGoal.f_o_p: got path = {path}')
```
ThomasSchellenbergNextCentury, line 617: Should probably say "TransferralGoal" rather than "TransGoal" in the debug statements here.

scene_generator/optimal_path.py
```diff
@@ -1,44 +1,96 @@
...
+
+    # dilate all the polygons based on the agent width
+    source_point = Point(*source_loc)
+    target_point = Point(*target_loc)
+    num_boundary_points = 0
+    polygons = []
+    for hole in holes:
+        # Warning: Increasing resolution can dramatically increase
+        # time required for path finding.
+        poly = Polygon(hole).buffer(dilation, resolution=4)
```
ThomasSchellenbergNextCentury: What does the resolution number do?

chrislong: It controls how many vertices get made to approximate a circular arc when growing the polygon. The default is 16, but I figured we didn't need that level of detail for obstacles for our path planning.

scene_generator/goals.py
chrislong: Ok.

chrislong:
I think all the comments have been addressed, ready for re-review.

scene_generator/geometry.py
```diff
@@ -6,9 +6,10 @@
 from separating_axis_theorem import sat_entry
 
 MAX_TRIES = 100
+PERFORMER_WIDTH = 0.1
 # the following mins and maxes are inclusive
-MIN_PERFORMER_POSITION = -4.8
-MAX_PERFORMER_POSITION = 4.8
+MIN_PERFORMER_POSITION = -4.8 + PERFORMER_WIDTH / 2.0
```
brianpippin, line 11: It seems that whenever you are using PERFORMER_WIDTH you are always using it divided by 2.  Have you considered changing the variable, or having another variable that does this, so you don't need to always redo this calculation.  I know dividing by 2 isn't expensive, but it might be cleaner?

chrislong: Sounds good

PR #90, opened 2020-05-13
Mcs 143: IntPhys Scene Generator: Create Scene Quartets

chrislong:
This does not complete the ticket, but in the interests of bite-sized PRs I'm doing this one as the start of a framework for quartets. It also includes a refactor of goals.py to break it into smaller pieces, and some more type hinting.


scene_generator/goal.py
```diff
@@ -0,0 +1,181 @@
+import logging
+import uuid
+from abc import ABC, abstractmethod
+import random
+from typing import Dict, Any, Tuple, List, Optional
+
+import geometry
+from geometry import random_position, random_rotation, calc_obj_pos, POSITION_DIGITS
```
deanwetherby, line 8: Why not just import geometry and then everywhere you use random_position or any of those other imports you prepend with geometry dot? My preference is to namespace these functions so I know where they came from. If typing geometry over and over is problematic, we could also import geometry as geom or something like that.

chrislong: I don't remember why, it might be whatever pycharm did by default or maybe it used to be in the same file. I think you're right, though, that it's usually better to use the namespace. I'll fix it.

PR #91, opened 2020-05-14
Mcs 139: IntPhys Scene Generator: Generate Object Permanence Scene Quartets

chrislong:
Skeleton code for quartets for object permanence scenes. Merge after https://github.com/NextCenturyCorporation/MCS/pull/90

### DIFF
--- a/scene_generator/scene_generator.py
+++ b/scene_generator/scene_generator.py
@@ -172,7 +172,8 @@ def generate_quartets(prefix: str, count: int, goal_type: str,
                 break
             index += 1
         generate_quartet(prefix, index, goal_type, find_path, stop_on_error)
-    
+
+
 def generate_fileset(prefix: str, count: int, goal_type: str, find_path: bool, stop_on_error: bool,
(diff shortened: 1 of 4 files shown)


scene_generator/scene_generator.py
```diff
@@ -162,7 +207,20 @@ def main(argv):
     if args.loglevel:
         logging.getLogger().setLevel(args.loglevel)
 
-    generate_fileset(args.prefix, args.count, args.goal, args.find_path, args.stop_on_error)
+    if args.goal is not None and args.quartet is not None:
+        parser.error('Cannot specify body goal and quartet')
```
deanwetherby, line 214: An option here is to use a mutually exclusive group and let the parser handle it. I prefer this method because the of the built-in help documentation makes it clear how to use. (python scene_generator.py -h)
group = parser.add_mutually_exclusive_group()
group.add_argument('--goal')
group.add_argument('--quartet')

chrislong: Good idea, thanks!

PR #92, opened 2020-05-14
Mcs 123: IntPhys Scene Generator: Implausible Object Permanence with Objects Moving Across

chrislong:


chrislong:
I accidentally pushed changes to this for MCS 130: IntPhys Scene Generator: Implausible Object Permanence with Objects Falling Down. If you want I can try to undo it, but I'm not positive how, so I'd rather leave them here and have this PR be fore both 123 and 130. (They're pretty closely related, at least.) Sorry.

chrislong:
I just realized that I was doing quartets wrong (each one as a standalone scene rather than each one a different variant of the same scene). Converting this to a draft while I fix it.

scene_generator/quartets.py
```diff
@@ -0,0 +1,100 @@
...
+        self._template = template
+        self._find_path = find_path
+        self._scenes: List[Optional[Dict[str, Any]]] = [None]*4
+
+    @abstractmethod
+    def get_scene(self, q: int) -> Dict[str, Any]:
+        pass
+
+
+class ObjectPermanenceQuartet(Quartet):
```
ThomasSchellenbergNextCentury, line 28: The quartet functions could use some unit tests.

scene_generator/intphys_goals.py
```diff
@@ -104,6 +107,7 @@ def get_config(self, goal_objects: List[Dict[str, Any]], all_objects: List[Dict[
         goal = copy.deepcopy(self.TEMPLATE)
         goal['last_step'] = self._last_step
         goal['action_list'] = [['Pass']] * goal['last_step']
+        goal['metadata']['objects'] = [obj['id'] for obj in self._goal_objects]
```
ThomasSchellenbergNextCentury, line 110: What is this for?

chrislong: It's so the quartets can find out which things in the scene are target objects and not occluders or walls.

ThomasSchellenbergNextCentury: Ah, OK.  We should make sure to remove this property from the non-debug output file.

chrislong: Oh, good point. Just pushed a fix for this.

PR #93, opened 2020-05-15
Mcs 140:  IntPhys Scene Generator: Generate Shape Constancy Scene Quartets

chrislong:
First phase of quartet for shape constancy. Will be completed in subsequent tickets.

### DIFF
--- a/scene_generator/quartets.py
+++ b/scene_generator/quartets.py
@@ -87,6 +87,71 @@ class ObjectPermanenceQuartet(Quartet):
         return scene
 
 
+class ShapeConstancyQuartet(Quartet):
+    def __init__(self, template: Dict[str, Any], find_path: bool):
+        super(ShapeConstancyQuartet, self).__init__(template, find_path)
+        self._goal = intphys_goals.ShapeConstancyGoal()
+        self._scenes[0] = copy.deepcopy(self._template)
+        self._goal.update_body(self._scenes[0], self._find_path)
+        # we need the b object for 3/4 of the scenes, so generate it now
+        self._b = self._create_b()
+
+    def _create_b(self) -> Dict[str, Any]:
+        a = self._scenes[0]['objects'][0]
+        while True:
+            b_def = random.choice(objects.OBJECTS_INTPHYS)
+            if b_def['type'] != a['type']:
+                break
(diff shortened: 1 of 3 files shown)


scene_generator/quartets.py
```diff
@@ -87,6 +87,71 @@ def get_scene(self, q: int) -> Dict[str, Any]:
...
 
 
+class ShapeConstancyQuartet(Quartet):
+    def __init__(self, template: Dict[str, Any], find_path: bool):
+        super(ShapeConstancyQuartet, self).__init__(template, find_path)
+        self._goal = intphys_goals.ShapeConstancyGoal()
+        self._scenes[0] = copy.deepcopy(self._template)
+        self._goal.update_body(self._scenes[0], self._find_path)
+        # we need the b object for 3/4 of the scenes, so generate it now
+        self._b = self._create_b()
```
ThomasSchellenbergNextCentury, line 97: I feel like we should be using terminology that's a little more descriptive than "a" and "b", but I don't have any great ideas.  Maybe "target_a" and "target_b"?  Or just leave it as-is.  Not a big deal.

scene_generator/quartets.py
```diff
@@ -87,6 +87,71 @@ def get_scene(self, q: int) -> Dict[str, Any]:
...
+
+    def _create_b(self) -> Dict[str, Any]:
+        a = self._scenes[0]['objects'][0]
+        while True:
+            b_def = random.choice(objects.OBJECTS_INTPHYS)
+            if b_def['type'] != a['type']:
+                break
+        b_def = util.finalize_object_definition(b_def)
+        b = util.instantiate_object(b_def, a['original_location'], a['material_list'])
+        return b
```
ThomasSchellenbergNextCentury, line 107: Isn't object B supposed to have the same scale as object A?

chrislong: I think maybe that requirement came after I did this branch? The scale stuff is in MCS-124.

ThomasSchellenbergNextCentury: Understood.  Thanks.

scene_generator/quartets.py
chrislong: I added a docstring that I think makes it clear enough what's going on.

PR #94, opened 2020-05-18
Mcs 124: IntPhys Scene Generator: Implausible Shape Constancy with Objects Moving Across

chrislong:


scene_generator/quartets.py
```diff
@@ -98,35 +102,77 @@ def __init__(self, template: Dict[str, Any], find_path: bool):
...
     def _create_b(self) -> Dict[str, Any]:
         a = self._scenes[0]['objects'][0]
-        while True:
-            b_def = random.choice(objects.OBJECTS_INTPHYS)
-            if b_def['type'] != a['type']:
-                break
+        possible_defs = []
+        normalized_scale = a['shows'][0]['scale']
+        # cylinders have "special" scaling: scale.y is half what it is
+        # for other objects
```
ThomasSchellenbergNextCentury, line 108: You can add something like this to the end of the comment:  `because Unity automatically doubles the scale.y of cylinders`

scene_generator/quartets.py
```diff
@@ -98,35 +102,77 @@ def __init__(self, template: Dict[str, Any], find_path: bool):
...
+        if a['type'] == 'cylinder':
+            normalized_scale = normalized_scale.copy()
+            normalized_scale['y'] *= 2
+        for obj_def in objects.OBJECTS_INTPHYS:
+            if obj_def['type'] != a['type']:
+                def_scale = obj_def['scale']
+                if obj_def['type'] == 'cylinder':
+                    def_scale = def_scale.copy()
+                    def_scale['y'] *= 2
```
ThomasSchellenbergNextCentury, line 117: This is relatively minor, but you could extract the special cylinder behavior into a separate function, since it's used in two places.

scene_generator/quartets.py
```diff
@@ -98,35 +102,77 @@ def __init__(self, template: Dict[str, Any], find_path: bool):
...
+            implausible_event_index = a['intphys_option']['implausible_event_index']
+            implausible_event_step = implausible_event_index + a['forces'][0]['stepBegin']
+            implausible_event_x = a['intphys_option']['position_by_step'][implausible_event_index]
+            a['hides'] = [{
+                'stepBegin': implausible_event_step
+            }]
+            b = copy.deepcopy(self._b)
+            b['shows'][0]['stepBegin'] = implausible_event_step
+            b['shows'][0]['position']['x'] = implausible_event_x
+            b['shows'][0]['position']['z'] = a['shows'][0]['position']['z']
```
ThomasSchellenbergNextCentury, line 140: `b['shows'][0]['position']['y']` is already set (in `self._b`) before it gets to here, right?  Just checking.

scene_generator/quartets.py
```diff
@@ -98,35 +102,77 @@ def __init__(self, template: Dict[str, Any], find_path: bool):
...
+    def _turn_b_into_a(self, scene: Dict[str, Any]) -> None:
+        scene['answer']['choice'] = 'implausible'
+        a = self._scenes[0]['objects'][0]
+        if self._goal._object_creator == intphys_goals.IntPhysGoal._get_objects_and_occluders_moving_across:
+            implausible_event_index = a['intphys_option']['implausible_event_index']
+            implausible_event_step = implausible_event_index + a['forces'][0]['stepBegin']
+            implausible_event_x = a['intphys_option']['position_by_step'][implausible_event_index]
+            b = copy.deepcopy(self._b)
+            b['shows'][0]['position']['x'] = a['shows'][0]['position']['x']
+            b['shows'][0]['position']['z'] = a['shows'][0]['position']['z']
```
ThomasSchellenbergNextCentury, line 159: You might want to add a comment here about the need to replace A with B first, so that you can then turn B into A.

scene_generator/quartets.py
```diff
@@ -98,35 +102,77 @@ def __init__(self, template: Dict[str, Any], find_path: bool):
...
             pass
-        elif self._goal._object_creator == IntPhysGoal._get_objects_falling_down:
+        elif self._goal._object_creator == intphys_goals.IntPhysGoal._get_objects_falling_down:
             # TODO: In MCS-131
             pass
         else:
             raise ValueError('unknown object creation function, cannot update scene')
 
-    def _b_replaces_a(self, body: Dict[str, Any], q: int) -> None:
+    def _b_replaces_a(self, body: Dict[str, Any]) -> None:
```
ThomasSchellenbergNextCentury, line 175: There's a part where we take the `info` from all the `objects` and append them to the `goal.info_list`, right?  It happens after this point, right?  Otherwise we'd need to update the `info_list`.

scene_generator/quartets.py
chrislong: Ok

scene_generator/quartets.py
chrislong: No, that's already happened. Good catch. Is `goal.info_list` supposed to contain info about all objects, or just the "goal" objects (i.e., not occluders or walls)?

scene_generator/quartets.py
chrislong: Yeah, it gets set when b in instantiated (to the original location of a)

scene_generator/quartets.py
```diff
@@ -98,35 +102,77 @@ def __init__(self, template: Dict[str, Any], find_path: bool):
...
+        b_def = random.choice(possible_defs)
         b_def = util.finalize_object_definition(b_def)
-        b = util.instantiate_object(b_def, a['original_location'], a['material_list'])
+        b = util.instantiate_object(b_def, a['original_location'], a['materials_list'])
         return b
 
-    def _turn_a_into_b(self, body: Dict[str, Any], q: int) -> None:
-        if self._goal._object_creator == IntPhysGoal._get_objects_and_occluders_moving_across:
-            # TODO: In MCS-124
+    def _turn_a_into_b(self, scene: Dict[str, Any]) -> None:
```
ThomasSchellenbergNextCentury, line 127: All these functions could use some unit tests.

scene_generator/quartets.py
chrislong: Considering how small it is, I think it's easier this way. I think it would either be more lines or less efficient or both to factor it out.

scene_generator/quartets.py
chrislong: Huh. Ok. That's still weird.

ThomasSchellenbergNextCentury: Agreed.  You can get a glimpse by looking at the screenshots here:  https://docs.unity3d.com/520/Documentation/Manual/PrimitiveObjects.html

scene_generator/quartets.py
ThomasSchellenbergNextCentury: Fair enough.

scene_generator/quartets.py
ThomasSchellenbergNextCentury: Awesome.  Thanks.

scene_generator/quartets.py
ThomasSchellenbergNextCentury: Yes, all the objects.  You did that already!  :)  https://nextcentury.atlassian.net/browse/MCS-203

chrislong: Ok, that's easy then. I'll fix it.

scene_generator/quartets_test.py
```diff
@@ -0,0 +1,43 @@
...
+    assert a['shows'][0]['scale']['z'] == quartet._b['shows'][0]['scale']['z']
+    assert a['materials'] == quartet._b['materials']
+
+
+def test_ShapeConstancyQuartet_get_scene_2():
+    # tests _turn_a_into_b
+    template = {'wallMaterial': 'dummy'}
+    quartet = ShapeConstancyQuartet(template, False)
+    scene = quartet.get_scene(2)
+    if quartet._goal._object_creator == intphys_goals.IntPhysGoal._get_objects_and_occluders_moving_across:
```
ThomasSchellenbergNextCentury, line 21: Do we need this `if` here because the scene setup (moving across vs falling down) is randomly chosen each time the test is run?

chrislong: Yep, that's the reason.

PR #97, opened 2020-05-20
Mcs 141:  IntPhys Scene Generator: Generate Spatio-Temporal Continuity Scene Quartets

chrislong:
Skeleton for Spatio-Temporal Continuity Quartet. Will be fleshed out in MCS-125 and MCS-132.


scene_generator/intphys_goals.py
```diff
@@ -133,6 +133,49 @@ def _get_num_paired_occluders(self) -> int:
...
+                paired_obj['intphys_option']['occluder_indices'] = occluder_indices
+                occluder_fits = True
+                break
+        if occluder_fits:
+            occluder_objs = objects.create_occluder(random.choice(non_wall_materials)[0],
+                                                    random.choice(pole_materials)[0],
+                                                    occluder_x, x_scale)
+        else:
+            occluder_objs = None
+        return occluder_objs
```
ThomasSchellenbergNextCentury, line 177: Is this mostly the same code as below?

scene_generator/intphys_goals.py
```diff
@@ -133,6 +133,49 @@ def _get_num_paired_occluders(self) -> int:
...
+        occluder_fits = False
+        for _ in range(IntPhysGoal.MAX_OCCLUDER_TRIES):
+            min_scale = min(max(paired_obj['shows'][0]['scale']['x'],
+                                IntPhysGoal.MIN_OCCLUDER_SCALE),
+                            IntPhysGoal.MAX_OCCLUDER_SCALE)
+            x_scale = random_real(min_scale, IntPhysGoal.MAX_OCCLUDER_SCALE, MIN_RANDOM_INTERVAL)
+            position_by_step = paired_obj['intphys_option']['position_by_step']
+            paired_z = paired_obj['shows'][0]['position']['z']
+            min_paired_x = -3 + x_scale / 2
+            max_paired_x = 3 - x_scale / 2
```
ThomasSchellenbergNextCentury, line 146: These variables are used to set the occluder's X position (as opposed to its X scale or X rotation), right?

scene_generator/intphys_goals.py
```diff
@@ -133,6 +133,49 @@ def _get_num_paired_occluders(self) -> int:
...
+            found_collision = False
+            for other_occluder in occluder_list:
+                if geometry.occluders_too_close(other_occluder, occluder_x, x_scale):
+                    found_collision = True
+                    break
+            if not found_collision:
+                # occluder_indices are needed by quartets
+                occluder_indices = paired_obj['intphys_option'].get('occluder_indices', [])
+                occluder_indices.append(position_index)
+                paired_obj['intphys_option']['occluder_indices'] = occluder_indices
```
ThomasSchellenbergNextCentury, line 168: We delete the `intphys_option` from the non-debug output file, right?

scene_generator/quartets.py
```diff
@@ -63,7 +63,7 @@ def _disappear_behind_occluder(self, body: Dict[str, Any]) -> None:
 
     def get_scene(self, q: int) -> Dict[str, Any]:
         if q < 1 or q > 4:
-            raise ValueError(f'q must be between 1 and 4 (exclusive), not {q}')
+            raise ValueError(f'q must be between 1 and 4 (inclusive), not {q}')
```
ThomasSchellenbergNextCentury, line 66: Did this get fixed in a different commit?

scene_generator/intphys_goals.py
chrislong: Yes. I'll rename the variables to make it clearer.

scene_generator/intphys_goals.py
chrislong: I'm guessing you mean similar to `_add_occluders`? It is very similar. The key difference is how the `x` value for the occluder position is computed. Paired occluders also need extra info stored somewhere about which `position_index` was used for them. I guess the computation of occluder `x` could be factored out into a function and that function passed as a parameter. I'm not sure exactly how the `position_index` would be handled, then, though.

scene_generator/intphys_goals.py
chrislong: Yes.

scene_generator/quartets.py
chrislong: Could be, I don't remember.

PR #101, opened 2020-05-22
Mcs 158: Don't Position Objects So They Automatically Meet the Goal Condition

chrislong:


scene_generator/interaction_goals_test.py
```diff
@@ -238,3 +239,32 @@ def test__generate_transferral_goal_with_nonstackable_goal():
     with pytest.raises(ValueError) as excinfo:
         goal = goal_obj.get_config([pickupable_obj, other_obj], [])
     assert "second object must be" in str(excinfo.value)
+
+
+def test_MCS_158_performerStart():
```
ThomasSchellenbergNextCentury, line 244: This is a nitpick, but since some people won't immediately recognize the ticket name, I think it would be better to rename these two unit tests.  Maybe `test_traversal_performer_start_not_close_to_target` and `test_transferral_targets_not_close_to_each_other`?

chrislong: Fine, I'll change it to accomodate those losers. ;-)

ThomasSchellenbergNextCentury: /s/losers/teammates/

PR #103, opened 2020-05-27
Mcs 161: [Pathfinding] Add RotateLook to Action List Before PickupObject or PutObject

chrislong:


scene_generator/interaction_goals.py
```diff
@@ -7,7 +7,9 @@
 import math
 from typing import Dict, Any, AnyStr, List, Tuple
 
-from machine_common_sense.mcs_controller_ai2thor import MAX_MOVE_DISTANCE
+from sympy import Segment, intersection
```
deanwetherby, line 10: Are there any speed implications here? After the reward issues, I prefer any other package over sympy.

chrislong: That's a fair question. Billy did the sympy parts, and I'm pretty sure it was before you discovered the performance issues with sympy. I expect I could change it to use shapely (or something else).

PR #107, opened 2020-05-29
Mcs 122: IntPhys Scene Generator: Implausible Gravity with Ramps (30 and 45 Degree)

chrislong:


scene_generator/quartets.py
```diff
@@ -31,6 +32,25 @@ def find_targets(scene: Dict[str, Any], num_targets: Optional[int] = None) -> Li
             for target_id in target_ids]
 
 
+def get_position_step(target: Dict[str, Any], x_position: float,
+                      left_to_right: bool, forwards: bool) -> int:
```
ThomasSchellenbergNextCentury, line 36: What does `forward` mean here?

scene_generator/quartets.py
```diff
@@ -473,11 +495,70 @@ def _get_steep_scene(self, q: int) -> Dict[str, Any]:
...
+        ramp_type = self._goal.get_ramp_type()
+        for obj in scene['objects']:
+            if obj.get('intphys_option', {}).get('moving_object', False):
+                obj['shows'][0]['position']['x'] *= -1
+                obj['forces'][0]['vector']['x'] *= -1
+                # adjust height to be on top of ramp
+                obj['shows'][0]['position']['y'] += ramps.RAMP_OBJECT_HEIGHTS[ramp_type]
+                # Add a downward force to all objects moving down the
+                # ramps so that they will move more realistically.
+                obj['forces'][0]['vector']['y'] = obj['mass'] * intphys_goals.IntPhysGoal.RAMP_DOWNWARD_FORCE
```
ThomasSchellenbergNextCentury, line 520: Would it make sense to combine some of this code with the duplicate parts of the `_get_ramp_and_objects` function in `intphys_goals`?

scene_generator/quartets.py
```diff
@@ -473,11 +495,70 @@ def _get_steep_scene(self, q: int) -> Dict[str, Any]:
...
+
+    def _get_ramp_offsets(self) -> Tuple[float, float]:
+        bottom_offset = -1.55 if self._goal.get_ramp_type() == ramps.Ramp.RAMP_30 else -2.4
+        top_offset = 1 if self._goal.get_ramp_type() == ramps.Ramp.RAMP_30 else -1
```
ThomasSchellenbergNextCentury, line 524: Probably want to add global variables for these magic numbers.

scene_generator/quartets.py
```diff
@@ -473,11 +495,70 @@ def _get_steep_scene(self, q: int) -> Dict[str, Any]:
         pass
 
     def _get_gentle_scene(self, q: int) -> Dict[str, Any]:
-        # TODO: in MCS-122
+        scene = copy.deepcopy(self._scenes[0])
+        if q == 2:
+            self._make_roll_down(scene)
```
ThomasSchellenbergNextCentury, line 500: It looks like this assumes that the object in the 1st scene is always rolling UP the ramp.  But is that a safe assumption?  Because it wasn't that way before.  Did we change that somewhere?  Was that part of MCS-142 and I don't remember?

scene_generator/quartets.py
chrislong: Forwards in the sense of progressing in the direction of the 'positions_by_step' as opposed to the opposite.

scene_generator/quartets.py
chrislong: I guess those two lines (for position.y and vector.y) could get factored out into another function that get called from both places.

scene_generator/quartets.py
chrislong: Ok, will do.

scene_generator/quartets.py
chrislong: In the description of MCS-122, it says scene 1 should be the object rolling up, so I added code to the GravityGoal to make it possible to force that to happen (for gentle ramps only; can't roll up steep ones).

ThomasSchellenbergNextCentury: Oh, is that the change you made to `_get_ramp_and_objects`?  I see now.  Not sure how I missed that before.  Sorry/thanks.

scene_generator/quartets.py
ThomasSchellenbergNextCentury: Ah, OK.

scene_generator/quartets.py
ThomasSchellenbergNextCentury: Hmm... Maybe you're right, and it's an extraneous/unnecessary bit of extraction.  If we ever need to change the behavior in the future, we can use the global constants to locate the relevant code.

scene_generator/quartets.py
chrislong: No problem. It's good to check assumptions.

PR #108, opened 2020-06-03
Mcs 133: IntPhys Scene Generator: Implausible Gravity with Ramps (90 Degree, 30/90, and 45/90 Combinations)

chrislong:


scene_generator/intphys_goals.py
```diff
@@ -484,7 +490,7 @@ class GravityGoal(IntPhysGoal):
         }
     }
 
-    def __init__(self, ramp_type: Optional[ramps.Ramp] = None, roll_down: Optional[bool] = None):
+    def __init__(self, ramp_type: Optional[ramps.Ramp] = None, roll_down: Optional[bool] = None, use_fastest: bool = False):
```
ThomasSchellenbergNextCentury, line 493: Isn't this constructor only called once?  Do we need the `use_fastest` argument here?

scene_generator/intphys_goals.py
```diff
@@ -596,7 +617,6 @@ def _get_ramp_and_objects(self, room_wall_material_name: str) -> \
             valid_positions = set(IntPhysGoal.Position)
         positions = []
 
-        valid_defs = objects.get_intphys_objects_y_0()
```
ThomasSchellenbergNextCentury, line 599: Is the `get_intphys_objects_y_0` function used elsewhere?  If not, you can delete it.

scene_generator/quartets.py
```diff
@@ -496,8 +501,39 @@ def get_scene(self, q: int) -> Dict[str, Any]:
...
+            target['forces'][0]['vector']['x'] = new_intphys_option['force']['x']
+            if not self._goal.is_left_to_right():
+                target['forces'][0]['vector']['x'] *= -1
+            if q == 4:
+                # Doing nothing here causes the downward force to
+                # start before the object reaches the top of the ramp,
+                # but it shouldn't affect it so we won't bother
+                # recomputing all the 'position_by_step' values for
+                # the slow speed to get the top-of-ramp value.
+                pass
```
ThomasSchellenbergNextCentury, line 520: Is this `if` statement still useful?

scene_generator/intphys_goals.py
```diff
@@ -332,15 +334,18 @@ def _get_objects_moving_across(self, room_wall_material_name: str, last_action_e
...
+                    'x': IntPhysGoal.OBJECT_POSITIONS[location][0],
                     'y': intphys_option['y'] + obj_def['position_y'],
-                    'z': object_positions[location][1]
+                    'z': IntPhysGoal.OBJECT_POSITIONS[location][1]
                 }
             }
             obj = instantiate_object(obj_def, object_location)
+            if 'saved_intphys_options' in obj_def:
+                # needed by GravityQuartet for steep ramps
+                intphys_option['saved_options'] = obj_def['saved_intphys_options']
```
ThomasSchellenbergNextCentury, line 345: I might not completely understand the workflow here, but is there a reason why we need both `saved_intphys_options` and `saved_options`?  Couldn't we just save the data in `saved_options` directly (in `_get_ramp_and_objects`), and not need the `saved_intphys_options` middleman?  Maybe I'm missing something.

scene_generator/intphys_goals.py
chrislong: It's only called explicitly in one place, but it can also be called (with no arguments) if the user uses the `--goal` argument on the command line.

scene_generator/intphys_goals.py
chrislong: Good catch, I'll remove it.

scene_generator/quartets.py
chrislong: I kept it because of the comment, but I could remove the actual `if`. I'll do that.

scene_generator/intphys_goals.py
chrislong: I don't think so. `saved_options` needs to be set on an object instance for use later by `GravityQuartet`. At the time `saved_intphys_options` is captured, we don't have an object instance yet, just object definitions. So I think we need to stuff the info in the object definition somewhere and then pull it out and put it in the object once it's instantiated.

ThomasSchellenbergNextCentury: Ah, OK.  Great explanation.  Thank you!

scene_generator/intphys_goals.py
ThomasSchellenbergNextCentury: Ah, right.  OK, that's fine then.

PR #109, opened 2020-06-04
MCS-240:  Add way for performers to submit for eval

brianpippin:


homepage/js/upload_s3_button.js
```diff
@@ -0,0 +1,97 @@
...
+        const url = "https://60lrpeo443.execute-api.us-east-1.amazonaws.com/version1/uploadEvaluationSubmission";
+
+        xhr.open("POST", url, true);
+        xhr.setRequestHeader("Content-Type", "application/json");
+        xhr.send(
+            JSON.stringify({
+                key: selectedFile.name
+            })
+        );
+        xhr.onload = function () {
```
chrislong, line 19: Shouldn't you set `onload` before doing the `send`? Seems like there could be a race if you don't.

homepage/js/upload_s3_button.js
```diff
@@ -0,0 +1,97 @@
...
+        super(props);
+        this.state = { file: undefined, fileName: ""}
+
+        this.onFormSubmit = this.onFormSubmit.bind(this)
+        this.handleChange = this.handleChange.bind(this)
+    }
+
+    onFormSubmit(e) {
+        e.preventDefault() 
+        const that = this;
```
chrislong, line 60: Just for my own edification: what's this for?

brianpippin: That looks like a relic I should have removed, good catch.

homepage/js/upload_s3_button.js
brianpippin: Probably for safety they should be inversed.  Done.

homepage/js/upload_s3_button.js
```diff
@@ -0,0 +1,96 @@
...
+                .then( response => {
+                    uploadFileToS3(response, this.state.file)
+                        .then (response => {
+                            console.log("Success!");
+                            $("#fileUploadStatus").text("File Upload Success!");
+                        })
+                })
+        } catch (e) {
+            console.log("An error occurred!", e.message);
+            $("#fileUploadStatus").text("Error uploading the file.");
```
ThomasSchellenbergNextCentury, line 73: We might want to print out the error message/stack trace here so the user can copy-and-paste it when they email us about it.

homepage/js/upload_s3_button.js
```diff
@@ -0,0 +1,96 @@
...
+        const formData = new FormData();
+
+        formData.append("file", file);
+
+        const xhr = new XMLHttpRequest();
+        xhr.open("PUT", presignedPostData.url, true);
+        xhr.setRequestHeader("Content-Type", "multipart/form-data");
+        xhr.onload = function () {
+            this.status === 200 ? resolve() : reject(this.responseText);
+        };
```
ThomasSchellenbergNextCentury, line 42: We might want to add an `xhr.onerror` here, and in the other request, that rejects and shows the error message in the `fileUploadStatus`.

PR #110, opened 2020-06-04
Mcs 235: Interaction Scenes: Generate Pairs of Scenes

chrislong:
Most implementation will be in subsequent tickets/branches. This is common stuff that I expect will be used for a bunch of it.

### DIFF
--- a/scene_generator/pairs.py
+++ b/scene_generator/pairs.py
@@ -1,0 +1,39 @@
+import copy
+from abc import ABC, abstractmethod
+import random
+from typing import Tuple, Dict, Any
+
+import geometry
+
+
+class InteractionPair(ABC):
+    def __init__(self, template: Dict[str, Any], find_path: bool):
+        self._template = template
+        self._find_path = find_path
+        self._compute_performer_start()
+    
+    @abstractmethod
+    def get_scenes(self) -> Tuple[Dict[str, Any], Dict[str, Any]]:
+        pass
+
+    def _compute_performer_start(self) -> Dict[str, Dict[str, float]]:
+        """Set the starting location (position & rotation) for the performer
+        (_performer_start). This default implementation chooses a
+        random location.
+        """
+        if getattr(self, '_performer_start', None) is None:
+            self._performer_start = {
+                'position': {
+                    'x': geometry.random_position(),
+                    'y': 0,
+                    'z': geometry.random_position()
+                },
+                'rotation': {
+                    'y': geometry.random_rotation()
+                }
+            }
+
+
+def get_pair_class():
+    # TODO in future tickets
+    return None
--- a/scene_generator/scene_generator.py
+++ b/scene_generator/scene_generator.py
@@ -8,8 +8,10 @@ import os.path
 import json
 import copy
 import random
-from typing import Dict, Any, List
+from enum import Enum, auto
+from typing import Dict, Any, List, Tuple
 
+import pairs
 import quartets
 from materials import *
 from pretty_json.pretty_json import PrettyJsonEncoder, PrettyJsonNoIndent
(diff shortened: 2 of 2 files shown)


scene_generator/pairs.py
```diff
@@ -0,0 +1,39 @@
...
+        if getattr(self, '_performer_start', None) is None:
+            self._performer_start = {
+                'position': {
+                    'x': geometry.random_position(),
+                    'y': 0,
+                    'z': geometry.random_position()
+                },
+                'rotation': {
+                    'y': geometry.random_rotation()
+                }
```
ThomasSchellenbergNextCentury, line 33: Would it make sense to combine this with Goal.compute_performer_start?

scene_generator/scene_generator.py
```diff
@@ -185,15 +187,66 @@ def generate_quartets(prefix: str, count: int, quartet_name: str,
...
+    scene = scenes[index - 1]
+    scene['name'] = name
+    scene_copy = copy.deepcopy(scene)
+    write_file(name, scene_copy)
+
+
+def generate_pair(prefix: str, count: int,
+                  find_path: bool, stop_on_error: bool) -> None:
+    template = generate_body_template('')
+    name_stem = f'{prefix}-{count:04}-'
```
ThomasSchellenbergNextCentury, line 202: Does the count need to be out to 4 digits?

chrislong: Not necessarily. The files list better in a file browser if they're 0-padded, but if you think 3 is enough, that would be fine. I wasn't sure how many scenario files we'd be making so I didn't know if 3 would be enough.

scene_generator/pairs.py
chrislong: Maybe. I'm not sure yet exactly how Goals and InteractionPairs are going to interact.

scene_generator/scene_generator.py
```diff
@@ -185,15 +187,66 @@ def generate_quartets(prefix: str, count: int, quartet_name: str,
...
+            logging.warning(f'failed to create a pair: {e}')
+    logging.debug(f'end generation of scene pair #{count}')
+
+
+def generate_pair_fileset(prefix: str, count: int,
+                          find_path: bool, stop_on_error: bool) -> None:
+    index = 1
+    while count > 0:
+        while True:
+            name = f'{prefix}-{index:04}-1.json'
```
ThomasSchellenbergNextCentury, line 224: What's the `-1` mean here?

scene_generator/scene_generator.py
```diff
@@ -185,15 +187,66 @@ def generate_quartets(prefix: str, count: int, quartet_name: str,
...
+            file_exists = os.path.exists(name)
+            if not file_exists:
+                break
+            index += 1
+        generate_pair(prefix, index, find_path, stop_on_error)
+        count -= 1
+
+
+class FilesetType(Enum):
+    NORMAL = auto()
```
ThomasSchellenbergNextCentury, line 234: Maybe we should rename NORMAL to SINGLE so it would better align with PAIR and QUARTET?

scene_generator/scene_generator.py
ThomasSchellenbergNextCentury: Oh, my bad, I thought the `count` was either 0 or 1.  You can keep it as 4.

scene_generator/pairs.py
ThomasSchellenbergNextCentury: Fair enough.

scene_generator/scene_generator.py
chrislong: That's the file name of the first scene in the pair. It's just used here to check if we've already made a scene pair with that `count`.

scene_generator/scene_generator.py
chrislong: Good idea.

PR #111, opened 2020-06-04
Mcs 230: Interaction Scenes: Function to Generate Multiple Similar Objects

chrislong:


scene_generator/util.py
```diff
@@ -92,3 +94,81 @@ def instantiate_object(object_def: Dict[str, Any],
...
+            else:
+                valid_defs.append(obj_def)
+    return valid_defs
+
+
+def get_def_with_new_type(obj: Dict[str, Any]) -> Dict[str, Any]:
+    valid_defs = get_similar_defs(obj, ('materialCategory',
+                                        'similarityScale'),
+                                  ('type',))
+    obj_def = random.choice(valid_defs)
```
ThomasSchellenbergNextCentury, line 157: Theoretically we should be smart enough to ensure that each object has at least one valid option for an object with similar type/material/scale, but, in case we make a mistake, we should probably handle the case in which `valid_defs` is empty.

chrislong: Good point. What do you think the right way to handle it is? Raise an exception (that will presumably bubble up until the main loop catches it and restarts the pair from scratch)?

scene_generator/util.py
```diff
@@ -92,3 +94,81 @@ def instantiate_object(object_def: Dict[str, Any],
     info.append(' '.join(info))
     new_object['info'] = info
     return new_object
+
+
+def get_similar_definition(obj: Dict[str, Any]) -> Dict[str, Any]:
```
ThomasSchellenbergNextCentury, line 99: Should this function instantiate the similar object before returning it?  (I'm not sure myself.)

scene_generator/util.py
ThomasSchellenbergNextCentury: Yeah, and include the object definition.

scene_generator/util.py
chrislong: Yeah, maybe. We'd have to pass in a location for it to do that, so I decided to keep it simpler.

ThomasSchellenbergNextCentury: Oh, right.  That makes sense.  Thanks.

PR #115, opened 2020-06-08
MCS-232: Interaction Scenes Pair 1: Position the Target Object Either Immediately Visible in the Player Camera or Not Visible

chrislong:


scene_generator/geometry.py
```diff
@@ -191,3 +202,64 @@ def occluders_too_close(occluder: Dict[str, Any], x_position: float, x_scale: fl
 def position_distance(a: Dict[str, float], b: Dict[str, float]) -> float:
     """Compute the distance between two positions."""
     return math.sqrt((a['x'] - b['x'])**2 + (a['y'] - b['y'])**2 + (a['z'] - b['z'])**2)
+
+
+def get_visible_segment(performer_start: Dict[str, Dict[str, float]]) \
+        -> shapely.geometry.Point:
+    logging.debug(f'>>>get_visible_segment: {performer_start}')
```
ThomasSchellenbergNextCentury, line 209: Not sure if you want to keep the debug logs in here.

scene_generator/geometry.py
```diff
@@ -191,3 +202,64 @@ def occluders_too_close(occluder: Dict[str, Any], x_position: float, x_scale: fl
...
+    target_segment = room.intersection(view_segment)
+    if target_segment.is_empty:
+        raise exceptions.SceneException(f'performer too close to the wall, cannot place object in front of it (performer location={performer_start})')
+    logging.debug(f'<<<get_visible_segment: {target_segment}')
+    return target_segment
+
+
+def get_location_in_front_of_performer(performer_start: Dict[str, Dict[str, float]],
+                                       target_def: Dict[str, Any]) -> Optional[Dict[str, Any]]:
+    visible_segment = get_visible_segment(performer_start)
```
ThomasSchellenbergNextCentury, line 229: We might want to add a comment here that we're assuming any object positioned along the line directly in front of the performer (at least MIN_START_DISTANCE_AWAY) will always be immediately visible to the performer.

scene_generator/geometry.py
```diff
@@ -191,3 +202,64 @@ def occluders_too_close(occluder: Dict[str, Any], x_position: float, x_scale: fl
...
+                        ROOM_DIMENSIONS[1][1] - ROOM_DIMENSIONS[1][0])
+    # if the performer were at the origin facing 0, this would be behind it
+    base_rear = shapely.geometry.box(-max_dimension*2, -max_dimension*2,
+                                     max_dimension*2, 0)
+    performer_point = shapely.geometry.Point(performer_start['position']['x'],
+                                             performer_start['position']['z'])
+    translated_rear = affinity.translate(base_rear, performer_point.x, performer_point.y)
+    performer_rear = affinity.rotate(translated_rear, -performer_start['rotation']['y'],
+                                     origin=performer_point)
+    bounds = performer_rear.bounds
```
ThomasSchellenbergNextCentury, line 252: This calculates the 180 degree arc behind the performer, correct?  We might want to add a comment here explaining that.

scene_generator/pairs.py
```diff
@@ -1,39 +1,107 @@
...
+        # choose target object
+        target_def = util.finalize_object_definition(random.choice(objects.get_all_object_defs()))
+        scene1 = self._get_empty_scene()
+        # place target object in scene 1 right in front of the performer
+        for _ in range(MAX_PLACEMENT_TRIES):
+            in_front_location = geometry.get_location_in_front_of_performer(self._performer_start, target_def)
+            if in_front_location is not None:
+                break
+        if in_front_location is None:
+            raise exceptions.SceneException('could not place object in front of performer')
```
ThomasSchellenbergNextCentury, line 84: I think `if in_front_location is None` or `if behind_location is None`, rather than raise an exception, we should just regenerate the random performerStart position and rotation, and do the whole thing again.  Would that be hard?

scene_generator/pairs.py
```diff
@@ -1,39 +1,107 @@
...
-from typing import Tuple, Dict, Any
+from typing import Tuple, Dict, Any, Type
 
+import exceptions
 import geometry
+import objects
+import util
+from geometry import ROOM_DIMENSIONS, MIN_START_DISTANCE_AWAY
+
+MAX_PLACEMENT_TRIES = 100
```
ThomasSchellenbergNextCentury, line 13: Do we have multiple `MAX_*_TRIES` now?  Would it make more sense to combine them?

ThomasSchellenbergNextCentury:
@chrislong We're still adding some other random objects to the scene after all of the Pair-specific objects are positioned, correct?  (That should be Yes.)

Is there a chance that a random object will be positioned in front of the performer, between the performer and the object that's supposed to be "immediately visible", therefore hiding the object?  I think the solution is to check if a random object's dimensions overlap the `visible_segment` and, if so, choose a new random position for it.  What do you think?

scene_generator/geometry.py
chrislong: Ok

scene_generator/geometry.py
chrislong: I thought I had a comment about that but I'll elaborate.

scene_generator/pairs.py
chrislong: > @chrislong We're still adding some other random objects to the scene after all of the Pair-specific objects are positioned, correct? (That should be Yes.)

No, not currently. I'll add it. How many more objects do you want?


> Is there a chance that a random object will be positioned in front of the performer, between the performer and the object that's supposed to be "immediately visible", therefore hiding the object? I think the solution is to check if a random object's dimensions overlap the `visible_segment` and, if so, choose a new random position for it. What do you think?

Yeah, I think that'll work. (Actually I think we want the segment between the performer and the immediate visible object, but I think the concept will work.)

chrislong:
> @chrislong We're still adding some other random objects to the scene after all of the Pair-specific objects are positioned, correct? (That should be Yes.)

No, not currently. I'll add it. How many more objects do you want?


> Is there a chance that a random object will be positioned in front of the performer, between the performer and the object that's supposed to be "immediately visible", therefore hiding the object? I think the solution is to check if a random object's dimensions overlap the `visible_segment` and, if so, choose a new random position for it. What do you think?

Yeah, I think that'll work. (Actually I think we want the segment between the performer and the immediate visible object, but I think the concept will work.)

scene_generator/pairs.py
chrislong: It's not that bad, I'll change it

scene_generator/pairs.py
chrislong: Yeah, there are 5. I guess we could combine them.

chrislong:
Comments addressed (5 new objects added currently), ready for re-review.

PR #116, opened 2020-06-08
Remove top drawer of changing table from enclosed areas list (MCS-213).

rartiss55:
This was blocked by MCS-227, since I needed to confirm which drawers its actually possible to see objects in without getting stuck in them. 

I think this fix was simple and straightforward, but PR-ing just in case.

### DIFF
--- a/scene_generator/objects.py
+++ b/scene_generator/objects.py
@@ -1326,17 +1326,6 @@ OBJECTS_IMMOBILE = [{
     "enclosed_areas": [{
         "position": {
             "x": 0.165,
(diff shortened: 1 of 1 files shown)


rartiss55:
Meant to post a screenshot yesterday -- technically I had to move to the side to see the object in the bottom drawer, since the front is so high. Is this acceptable?
Screen Shot 2020-06-08 at 5 56 17 PM

ThomasSchellenbergNextCentury:
@rartiss55 Can't you just look down a little to see the object in the drawer?

PR #119, opened 2020-06-10
MCS-238: Interaction Scenes Pair 8: Swap Target Object and Similar Object Inside Container And Adjacent

chrislong:


scene_generator/scene_generator.py
```diff
@@ -65,6 +65,7 @@ def clean_object(obj: Dict[str, Any]) -> None:
     obj.pop('dimensions', None)
     obj.pop('intphys_option', None)
     obj.pop('materials_list', None)
+    obj.pop('materialCategory', None)
```
deanwetherby, line 68: Any reason this string is camelCase while the others are snake?

scene_generator/util_test.py
```diff
@@ -1,8 +1,10 @@
 import uuid
 
+import geometry
 import materials
+import objects
 from goals import *
-from util import finalize_object_definition, instantiate_object
+from util import finalize_object_definition, instantiate_object, check_same_and_different, get_similar_defs
```
deanwetherby, line 7: I ran across this article recently about the use of util. I am guilty of this same practice. Just sharing awareness.
https://breadcrumbscollector.tech/stop-naming-your-python-modules-utils/

scene_generator/util.py
```diff
@@ -105,3 +107,81 @@ def instantiate_object(object_def: Dict[str, Any],
...
+    else:
+        return get_def_with_new_scale(obj)
+
+
+def check_same_and_different(a: Dict[str, Any], b: Dict[str, Any],
+                            same: Iterable[str], different: Iterable[str]) -> bool:
+    """Return true iff for all properties in same that are in a, they
+    exist in b and have the same value, and for all properties in
+    different that are in a, they exist in b and are different.
+    """
```
deanwetherby, line 149: Thank you for having a unit test for this which helped explain this function along with the type hinting. I kept wondering why you weren't using dict == dict or intersection/union on the dictionary keys. So this function compares the dictionaries a and b ensuring that the keys represented by 'same' are equal between the two dictionaries while the key strings in 'different' contain nonequal values. If all those conditions are met, then it returns True.

scene_generator/scene_generator.py
chrislong: I don't know. @ThomasSchellenbergNextCentury might know.

scene_generator/util_test.py
chrislong: Good points, thanks for the link.

brianpippin:
Nothing jumped out at me.

scene_generator/scene_generator.py
ThomasSchellenbergNextCentury: I guess I forgot which case I was using at the time.

scene_generator/objects.py
```diff
@@ -3934,7 +3934,12 @@ def get_enclosed_containers() -> List[Dict[str, Any]]:
     global _ENCLOSED_CONTAINERS
     if _ENCLOSED_CONTAINERS is None:
         all_defs = get_all_object_defs()
-        _ENCLOSED_CONTAINERS = [obj_def for obj_def in all_defs if 'enclosed_areas' in obj_def and len(obj_def['enclosed_areas']) > 0]
+        _ENCLOSED_CONTAINERS = [
+            obj_def for obj_def in all_defs
+            if 'enclosed_areas' in obj_def and len(obj_def['enclosed_areas']) > 0
+            or 'choose' in obj_def and 'enclosed_areas' in obj_def['choose'][0]
+            and len(obj_def['choose'][0]['enclosed_areas']) > 0
```
ThomasSchellenbergNextCentury, line 3941: There are a lot of ands and ors together here, and I'm confused about the order of operations.  Maybe some parentheses would help with the clarity?

chrislong: Ok, I'll add parens.

scene_generator/interaction_goals.py
```diff
@@ -7,9 +7,14 @@
...
 from typing import Dict, Any, AnyStr, List, Tuple, Sequence
 
+<<<<<<< HEAD
+import util
+from machine_common_sense.mcs_controller_ai2thor import MAX_MOVE_DISTANCE
+=======
 from sympy import Segment, intersection
 
 from machine_common_sense.mcs_controller_ai2thor import MAX_MOVE_DISTANCE, PERFORMER_CAMERA_Y
+>>>>>>> master
```
ThomasSchellenbergNextCentury, line 17: Merge chaff

chrislong: Oops, sorry, fixing...

scene_generator/geometry.py
```diff
@@ -166,19 +179,58 @@ def can_enclose(objectA: Dict[str, Any], objectB: Dict[str, Any]) -> bool:
...
+                                  -> List[Dict[str, Any]]:
+    """Return a list of object definitions for containers that can enclose
+    all the pass objects objs. If container_defs is None, use
+    objects.get_enclosed_containers().
+    """
+    if container_defs is None:
+        container_defs = objects.get_enclosed_containers()
+    valid_container_defs = []
+    for container_def in container_defs:
+        index = can_contain(container_def, *objs)
```
ThomasSchellenbergNextCentury, line 214: I forget if we talked about this before, but I think we should try rotating the object to see if it would fit inside the container at a different angle.  Rotation along just the Y axis is fine.  I'm not sure what increments we should use; maybe 45 degree?

With a single object, this seems like a simple loop; however, with multiple objects, you not only want to try rotating each object (and each combination across the objects), but you may also want to reposition one object in relation to another object (for example, if object A is to the right of object B, and they don't fit, maybe swap them so that object B is to the right of object A).  You can split the repositioning out into a separate ticket.

chrislong: This is addressed somewhat for 2 objects in another branch. Ironically, the latest code now tests for 90 degree rotations for 2 objects together (and side-by-side and front-to-back), but not for rotations of a single object.

scene_generator/geometry.py
```diff
@@ -191,3 +243,120 @@ def occluders_too_close(occluder: Dict[str, Any], x_position: float, x_scale: fl
...
+
+def get_adjacent_location(obj_def: Dict[str, Any],
+                          target: Dict[str, Any],
+                          performer_start: Dict[str, float]) -> Dict[str, Any]:
+    """Find a location such that, if obj_def is instantiated there, it
+    will be next to target. Ensures that the object at the new
+    location will not overlap the performer start, if necessary trying
+    to put the new object on each cardinal side of the target.
+    """
+    for side in range(4):
```
ThomasSchellenbergNextCentury, line 317: It looks like this function always tries using side 0 first, then side 1, then 2, then 3?  I think it would be better to try the sides in a random order.

scene_generator/geometry.py
ThomasSchellenbergNextCentury: Ah, OK.  So should we address different rotations (like I described above) in that other branch?

chrislong: I'd rather make a new branch for it, to handle different rotations of a single object in a container.

scene_generator/geometry.py
chrislong: Yes, that's what it does now. That's easy, I'll change it to be random.

scene_generator/geometry.py
ThomasSchellenbergNextCentury: Sure.  I made a ticket:  https://nextcentury.atlassian.net/browse/MCS-286

scene_generator/util.py
ThomasSchellenbergNextCentury: Yeah, the idea here is that, given a specific attribute (currently either color, shape, or size), we want the two objects to be exactly the same, except for that attribute.  So you have a blue ball next to a red ball, or a blue ball next to a blue block, or a blue ball next to a larger blue ball.

PR #120, opened 2020-06-11
MCS-258 Return Object Shape and Color in Object Metadata

ThomasSchellenbergNextCentury:
Corresponding Unity PR https://github.com/NextCenturyCorporation/ai2thor/pull/43

### DIFF
--- a/scene_generator/objects.py
+++ b/scene_generator/objects.py
@@ -283,8 +283,8 @@ OBJECTS_PICKUPABLE_BLOCKS = [{
         "z": 2
     }
 }, {
-    # Readers, please ignore the "yellow number 1" in the type: the object's chosen material will change this design.
-    "type": "block_yellow_number_1",
+    # Readers, please ignore the "blue letter c" in the type: the object's chosen material will change this design.
+    "type": "block_blue_letter_c",
     "info": ["tiny", "letter block", "cube"],
     "mass": 0.66,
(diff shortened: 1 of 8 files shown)


rartiss55:
Built the associated unity branch and ran this with the playroom...noticed the changing table shelves didn't have colors listed:
Screen Shot 2020-06-11 at 1 07 17 PM

scene_generator/objects.py
```diff
@@ -283,8 +283,8 @@
         "z": 2
     }
 }, {
-    # Readers, please ignore the "yellow number 1" in the type: the object's chosen material will change this design.
-    "type": "block_yellow_number_1",
+    # Readers, please ignore the "blue letter c" in the type: the object's chosen material will change this design.
+    "type": "block_blue_letter_c",
```
rartiss55, line 287: Silly question but: why did this change?

ThomasSchellenbergNextCentury:
Re changing table shelves: Yeah, that's more complicated, but fortunately the changing table's shelves are never going to be the Target Object in a Goal, so I'm not in as much of a rush on that.

scene_generator/objects.py
ThomasSchellenbergNextCentury: An object's "shape" word is dependent on its specific type, so the type here needs to be a "number block" to match its description for the scene generator.

PR #121, opened 2020-06-11
MCS-239: Interaction Scenes Pair 2: Position the Target Object Either Immediately Visible in the Player Camera or Behind an Object

chrislong:


brianpippin:
Everything looks good, do you think we should write some tests for any of this?

chrislong:
> Everything looks good, do you think we should write some tests for any of this?

Yeah, probably a good idea. I'll do that.

chrislong:
Ready for re-review.

scene_generator/objects.py
```diff
@@ -1226,14 +1226,14 @@
...
+            "y": 0.656,
+            "z": 0.62
         },
         "offset": {
-            "x": -0.024 / 0.5,
-            "y": 0.32 / 0.5,
-            "z": -0.018 / 0.5
+            "x": -0.024,
+            "y": 0.32,
+            "z": -0.018
```
ThomasSchellenbergNextCentury, line 1236: Hmm... The fact that these were wrong makes me wonder about the properties for the other plants.  Are you sure this matches what you saw in the Unity Editor?  Note that these properties are divided by `0.5` because the scale is set to `0.5` on each axis.

scene_generator/pairs.py
```diff
@@ -146,7 +205,8 @@ def get_scenes(self) -> Tuple[Dict[str, Any], Dict[str, Any]]:
         return scene1, scene2
 
 
-_INTERACTION_PAIR_CLASSES = [ImmediatelyVisiblePair, SimilarAdjacentContainedPair]
+#_INTERACTION_PAIR_CLASSES = [ImmediatelyVisiblePair, SimilarAdjacentContainedPair]
```
ThomasSchellenbergNextCentury, line 208: What's up with the commented-out code here?

scene_generator/pairs.py
```diff
@@ -101,6 +102,64 @@ def get_scenes(self) -> Tuple[Dict[str, Any], Dict[str, Any]]:
...
+
+    def get_scenes(self) -> Tuple[Dict[str, Any], Dict[str, Any]]:
+        for _ in range(MAX_PLACEMENT_TRIES):
+            target_def = util.finalize_object_definition(random.choice(objects.get_all_object_defs()))
+            occluder_defs = geometry.get_wider_and_taller_defs(target_def)
+            if occluder_defs is not None:
+                break
+        if occluder_defs is None:
+            raise exceptions.SceneException('could not get a target and occluder')
+        occluder_def = util.finalize_object_definition(random.choice(occluder_defs))
```
ThomasSchellenbergNextCentury, line 117: Using the word `occluder` here might be confusing since we use the same word in a different context for the IntPhys goals.  I don't have a better word suggestion though, and maybe I'm worrying too much.

scene_generator/pairs.py
```diff
@@ -101,6 +102,64 @@ def get_scenes(self) -> Tuple[Dict[str, Any], Dict[str, Any]]:
...
+                break
+        if in_front_location is None:
+            raise exceptions.SceneException('could not place occluder in front of performer')
+        # Rotate occluder to be facing the performer. There is a
+        # chance that this rotation could cause the occluder to
+        # intersect the wall of the room, because it's different from
+        # the rotation returned by get_location_in_front_of_performer
+        # (which does check for that). But it seems pretty unlikely.
+        dx = self._performer_start['position']['x'] - in_front_location['position']['x']
+        dz = self._performer_start['position']['z'] - in_front_location['position']['z']
```
ThomasSchellenbergNextCentury, line 144: Remind me why we're doing this rotation again?  Was this just to make the math easier?  Not a deal-breaker, just curious.

scene_generator/pairs.py
chrislong: Oops, testing artifact, I'll fix it.

scene_generator/pairs.py
chrislong: I see what you mean. I don't think it's a big deal, but maybe could be improved. Do you think `blocker` or `hider` would be better in this context?

scene_generator/pairs.py
chrislong: It's to make the occlusion calculation easier. Otherwise we'd have to compute how wide the occluder appears to be from the performer's perspective.

scene_generator/geometry.py
```diff
@@ -360,3 +375,44 @@ def _adjacent_location(obj_def: Dict[str, Any],
...
+
+def get_wider_and_taller_defs(obj_def: Dict[str, Any]) \
+        -> List[Dict[str, Any]]:
+    dims = obj_def['dimensions']
+    bigger_defs = []
+    for new_def in objects.get_all_object_defs():
+        if 'dimensions' in new_def:
+            if new_def['dimensions']['x'] >= dims['x'] and \
+               new_def['dimensions']['y'] >= dims['y']:
+                bigger_defs.append(new_def)
```
ThomasSchellenbergNextCentury, line 388: So this function doesn't check the Z dimensions to see if rotating the object 90 degrees would cause it to be wider?  I guess that's fine, though it would be nice to handle that case as well, if it's not too hard.  (And maybe this answers my question about the rotation in `HiddenBehindPair.get_scenes` too?)

scene_generator/pairs.py
```diff
@@ -101,6 +102,64 @@ def get_scenes(self) -> Tuple[Dict[str, Any], Dict[str, Any]]:
...
+        angle = math.degrees(math.atan2(dz, dx))
+        # negative because we do clockwise rotation
+        in_front_location['rotation']['y'] = -angle
+        occluder = util.instantiate_object(occluder_def, in_front_location)
+        occluded_location = geometry.get_adjacent_location_on_side(target_def,
+                                                                   occluder,
+                                                                   self._performer_start['position'],
+                                                                   2)
```
ThomasSchellenbergNextCentury, line 152: Maybe the specific side numbers should be an enum or global variables?

scene_generator/pairs.py
ThomasSchellenbergNextCentury: I think either of those would work.  Maybe `obstruction` too.

scene_generator/objects.py
chrislong: Maybe I'm not reading the settings correctly in the Unity editor. I was basing it on what I saw in the BoundingBox for the object.

scene_generator/geometry.py
chrislong: Yeah, it only checks the unrotated object. It wouldn't be trivial because it would have to remember which way to orient it for later when it's instantiated. I'd prefer a new ticket for this feature.

scene_generator/pairs.py
ThomasSchellenbergNextCentury: That makes sense.  Thanks.

scene_generator/geometry.py
ThomasSchellenbergNextCentury: Sounds good.  The ticket:  https://nextcentury.atlassian.net/browse/MCS-287

scene_generator/objects.py
ThomasSchellenbergNextCentury: Yeah, I just double-checked, and I'm confident that the original values were correct.  In the Unity Editor, displayed values don't always correspond to their final computed values.  So the numbers that you saw on the bounding box component were probably being modified under-the-hood by a `scale` in the `Transform` of a parent component.  I'll revert the commit.

scene_generator/pairs.py
chrislong: Darn you, that's a good idea. Fixed.

PR #122, opened 2020-06-12
MCS-266 Traversal reward fix

deanwetherby:
Initially, the traversal reward was relying on the ai2thor object metadata field distanceXZ. We were misinformed what the distance measure represented which turned out to be the distance from the agent to the object center. For oblong objects like a couch, the agent could be near the arm of the couch but nowhere near the center. The fix was to use the object extents and shapely distance from agent center to polygon edge like we do for the transferral reward calculation. Added tests for oblong objects on long and short sides as well as within.


deanwetherby:
The only question remains if this gets merged before or after the 0.0.9 release. Only Saurabh has chimed in about them having a work-around.

PR #123, opened 2020-06-12
MCS-170

arashBehpour:


scene_generator/goal.py
```diff
@@ -20,23 +21,33 @@
...
     while tries < MAX_TRIES:
         rotation = random.choice((0, 90, 180, 270))
         new_x = geometry.random_position()
         new_z = geometry.random_position()
         new_x_size = round(random.uniform(MIN_WALL_WIDTH, MAX_WALL_WIDTH), geometry.POSITION_DIGITS)
+        
+        if ((rotation == 0 or rotation == 180) and (new_z < -SAFE_DIST_FROM_ROOM_WALL or new_z > SAFE_DIST_FROM_ROOM_WALL)) or \
+            ((rotation == 90 or rotation == 270) and (new_x < -SAFE_DIST_FROM_ROOM_WALL or new_x > SAFE_DIST_FROM_ROOM_WALL)): #This is to make sure the wall is not too close to the rooms 4 walls
+            tries += 1
+            continue
```
chrislong, line 43: I would put the rest of the loop in an `else` clause, instead. Then it's easier to see, for example, that you are always incrementing `tries` and aren't doing that in two places.

scene_generator/goal.py
```diff
@@ -20,23 +21,33 @@
...
+
         rect = geometry.calc_obj_coords(new_x, new_z, new_x_size, WALL_DEPTH, 0, 0, rotation)
+        test_rect = geometry.calc_obj_coords(new_x, new_z, new_x_size + DIST_WALL_APART, WALL_DEPTH + DIST_WALL_APART, 0, 0, rotation)
```
chrislong, line 46: Shouldn't the `DIST_WALL_APART` be multiplied by 2 here (for both x and z) to account for buffers on both sides of the wall?

scene_generator/goal.py
```diff
@@ -146,12 +157,15 @@ def get_config(self, goal_objects: List[Dict[str, Any]], all_objects: List[Dict[
     def generate_walls(self, material: str, performer_position: Dict[str, Any],
                        bounding_rects: List[List[Dict[str, float]]]) -> List[Dict[str, Any]]:
         wall_count = random.choices(WALL_COUNTS, weights=WALL_PROBS, k=1)[0]
-
-        walls = []
+        
+        walls = [] 
+        all_bounding_rects = [copy.deepcopy(bounding_rect) for bounding_rect in bounding_rects] # Add bounding rects to walls
```
chrislong, line 162: Wouldn't it be sufficient to use `bounding_rects.copy()`?

scene_generator/goal.py
arashBehpour: I did not multiply by 2 because I saw the calc_obj_coords() takes the value new_x_size + DIST_WALL_APART as the delta_x which will be used for the x_plus and x_minus.

scene_generator/goal.py
arashBehpour: Sounds good I will fix that!

scene_generator/goal.py
arashBehpour: Yeah, I forgot about list having a copy method. Thank you!

scene_generator/goal.py
brianpippin: I would also put your comment above the if block with your logic and reasoning, it is hard to read at the end of the if statement and usually it is better to put the comment right above the line so people don't miss it.

scene_generator/goal.py
```diff
@@ -20,23 +21,33 @@
...
+
         rect = geometry.calc_obj_coords(new_x, new_z, new_x_size, WALL_DEPTH, 0, 0, rotation)
+        test_rect = geometry.calc_obj_coords(new_x, new_z, new_x_size + DIST_WALL_APART, WALL_DEPTH + DIST_WALL_APART, 0, 0, rotation)
         if not geometry.collision(rect, performer_position) and \
-                all(geometry.point_within_room(point) for point in rect) and \
-                (len(other_rects) == 0 or not any(
-                    separating_axis_theorem.sat_entry(rect, other_rect) for other_rect in other_rects)):
+                geometry.rect_within_room(rect) and \
+                (len(other_rects) == 0 or not any(separating_axis_theorem.sat_entry(rect, other_rect) for other_rect in other_rects)) and \
+                (len(other_rects) == 0 or not any(separating_axis_theorem.sat_entry(test_rect, other_rect) for other_rect in other_rects)): #Parallel walls should be at least 1 apart on the appropriate axis
```
brianpippin, line 50: See my comment about comments from above.

scene_generator/goal.py
brianpippin: If possible put your comments on the line above the line, I know it's probably not what you are used to doing it, but that is the convention we use, and sorry I keep leaving the same comment :)

scene_generator/goal.py
arashBehpour: Sounds good!

scene_generator/goal.py
arashBehpour: Good with me!

scene_generator/goal.py
arashBehpour: Should be updated

scene_generator/goal.py
chrislong: Ok, I think I didn't fully understand `test_rect` before.

scene_generator/goal.py
```diff
@@ -20,24 +21,36 @@
...
+            continue
+        else:
+
+            rect = geometry.calc_obj_coords(new_x, new_z, new_x_size, WALL_DEPTH, 0, 0, rotation)
+            # test_rect is to allow parallel walls to be at least 1(DIST_WALL_APART) apart on the appropriate axis
+            test_rect = geometry.calc_obj_coords(new_x, new_z, new_x_size + DIST_WALL_APART, WALL_DEPTH + DIST_WALL_APART, 0, 0, rotation)
+            if not geometry.collision(rect, performer_position) and \
+                    geometry.rect_within_room(rect) and \
+                    (len(other_rects) == 0 or not any(separating_axis_theorem.sat_entry(rect, other_rect) for other_rect in other_rects)) and \
+                    (len(other_rects) == 0 or not any(separating_axis_theorem.sat_entry(test_rect, other_rect) for other_rect in other_rects)):
```
chrislong, line 52: `test_rect` is in the same place as `rect`, but bigger, right? So you should be able to test only `test_rect` and not bother testing `rect`, right?

brianpippin:
You might want to see if you can add a couple of tests for this use case.

scene_generator/goal.py
arashBehpour: Yeah good point! Because if they intersect by the first statement making it false, the second statement is automatically false.

PR #125, opened 2020-06-16
Rotation bug fix

rartiss55:
Making rotation property on MCS_Object a dict containing x, y, and z rotation angles, instead of just a single float value.

I think I updated all the tests/documentation, please let me know if I missed anything.


deanwetherby:
There's another ['rotation']['y'] in retrieve_rotation function in the controller.

rartiss55:
@deanwetherby I think that's for the agent, who can only rotate around y, so I felt like I should leave that one alone.

python_api/test/test_mcs_object.py
```diff
@@ -79,8 +79,8 @@ def test_position(self):
         self.assertIsInstance(self.mcs_object.position, dict)
 
     def test_rotation(self):
-        self.assertAlmostEqual(self.mcs_object.rotation, 0.0)
-        self.assertIsInstance(self.mcs_object.rotation, float)
+        self.assertFalse(self.mcs_object.rotation)
```
chrislong, line 82: I'd compare it with an empty dict, but that might be just a style preference.

deanwetherby: I struggle with this one. I would use assertFalse as well for empty lists and dicts but it is perhaps more readable your way.

brianpippin:
So looking through the code it's still totally okay that some places rotation takes a float like in rotate look and we check for rotation being a float in validate_and_convert_params right?  I'm just trying to reconcile how we now have rotation as a float in some places and a dict in other places now and trying to make sure we won't get into any error assigning one wrong object toe the other.  

Example:
RotateLook
Rotate your viewport left/right and/or up/down based on your current viewport.

Parameters
rotation : float
Rotation degrees around the Y axis to change your look angle (left/right).
horizon : float
Rotation degrees around the X axis to change your look angle (up/down). This affects your current "head tilt".

rartiss55:
@brianpippin I think its okay -- I think there's a (hopefully) understandable difference with the agent's rotation (which would only make sense around y currently + the camera rotating along x) versus object rotation.

I think we do want to take extra care to explain the update in the release notes so that nothing breaks for anyone at this point, but then again they probably will just stick with the version they have if they don't need the rotation values.

PR #126, opened 2020-06-16
MCS-243: Interaction Scenes Pair 3: Either With or Without a Similar Object Positioned Adjacent To

chrislong:
I think this is the best branch to compare against, but I'm not positive. Sorry if this duplicates some stuff that's already been reviewed.


scene_generator/containers.py
```diff
@@ -0,0 +1,219 @@
...
+import geometry
+import objects
+import util
+from geometry import Orientation
+
+
+def put_object_in_container(obj: Dict[str, Any],
+                            container: Dict[str, Any],
+                            container_def: Dict[str, Any],
+                            area_index: int) -> None:
```
deanwetherby, line 13: I hadn't thought about using None return type when the function doesn't return anything. I usually just leave the return blank but I will modify that practice now that I see it this way.

scene_generator/containers.py
```diff
@@ -0,0 +1,219 @@
...
+        shows_b['position']['z'] += height_b / 2.0
+    shows_a['rotation'] = { 'y': rot_a }
+    shows_b['rotation'] = { 'y': rot_b }
+
+
+def can_enclose(objectA: Dict[str, Any], objectB: Dict[str, Any]) -> bool:
+    """Return True iff each 'dimensions' of objectA is >= the corresponding dimension of objectB."""
+    return objectA['dimensions']['x'] >= objectB['dimensions']['x'] and \
+        objectA['dimensions']['y'] >= objectB['dimensions']['y'] and \
+        objectA['dimensions']['z'] >= objectB['dimensions']['z']
```
deanwetherby, line 88: Feels like if this was a class instead of a dict we do some dunder things (__gt__, __lt__, __eq__) here. Hmmm.

chrislong: Yeah. And if objects and object definitions were classes, that would make some things easier, too. I'm not sure if it's worth the overhead, though.

scene_generator/containers.py
```diff
@@ -0,0 +1,234 @@
...
+    bx = obj_b['dimensions']['x']
+    az = obj_a['dimensions']['z']
+    bz = obj_b['dimensions']['z']
+    cx = ea_def['dimensions']['x']
+    cz = ea_def['dimensions']['z']
+    # first try side-by-side
+    width = ax + bx
+    depth = max(az, bz)
+    if cx >= width and cz >= depth:
+        return Orientation.SIDE_BY_SIDE, 0, 0
```
ThomasSchellenbergNextCentury, line 125: I'm thinking it wouldn't matter in this calculation if we switched the positions of objects A and B?

scene_generator/pairs.py
```diff
@@ -205,8 +251,8 @@ def get_scenes(self) -> Tuple[Dict[str, Any], Dict[str, Any]]:
         return scene1, scene2
 
 
-#_INTERACTION_PAIR_CLASSES = [ImmediatelyVisiblePair, SimilarAdjacentContainedPair]
-_INTERACTION_PAIR_CLASSES = [HiddenBehindPair]
+#_INTERACTION_PAIR_CLASSES = [HiddenBehindPair, ImmediatelyVisiblePair, SimilarAdjacentContainedPair]
+_INTERACTION_PAIR_CLASSES = [SimilarAdjacentPair]
```
ThomasSchellenbergNextCentury, line 255: I think this is an artifact of testing.

scene_generator/util.py
```diff
@@ -4,12 +4,13 @@
 import random
 from typing import Dict, Any, Optional, List, Tuple, Iterable
 
-import geometry
 import materials
 import objects
 
 
 MIN_RANDOM_INTERVAL = 0.05
+TARGET_CONTAINED_CHANCE = 0.25
```
ThomasSchellenbergNextCentury, line 12: Let's increase this chance to 0.5

scene_generator/pairs.py
```diff
@@ -160,6 +160,52 @@ def get_scenes(self) -> Tuple[Dict[str, Any], Dict[str, Any]]:
         return scene1, scene2
     
 
+class SimilarAdjacentPair(InteractionPair):
```
ThomasSchellenbergNextCentury, line 163: Would you mind adding a description of the scenes to each Pair class?  You can use the one from the corresponding JIRA ticket.

scene_generator/util.py
chrislong: Ok, will do

scene_generator/pairs.py
chrislong: Yes, it is, sorry, I'll fix it.

scene_generator/pairs.py
chrislong: No problem, will do.

scene_generator/containers.py
chrislong: Yeah, I think that's the case.

PR #127, opened 2020-06-16
MCS-244: Interaction Scenes Pair 4: Either With or Without a Similar Object Positioned Far Away

chrislong:


scene_generator/geometry.py
```diff
@@ -358,3 +360,12 @@ def get_bounding_polygon(obj: Dict[str, Any]) -> shapely.geometry.Polygon:
         poly = shapely.geometry.box(x - dx, z - dz, x + dx, z + dz)
         poly = shapely.affinity.rotate(poly, -show['rotation']['y'])
     return poly
+
+
+def are_adjacent(obj_a: Dict[str, Any], obj_b: Dict[str, Any]) -> bool:
+    poly_a = get_bounding_polygon(obj_a)
+    poly_b = get_bounding_polygon(obj_b)
+    distance = poly_a.distance(poly_b)
```
ThomasSchellenbergNextCentury, line 368: I'm assuming this calculates the distance from edge to edge (not center to center)?  The docs I found weren't clear.

chrislong: The docs say it returns the "minimum distance", which if it's not the distance between the edges I'm not sure what it would be.

PR #128, opened 2020-06-17
MCS-270: Transferral goal optimal path has no pickup action for goal object

chrislong:


chrislong:
I think maybe we should make a ticket for the problem with test_IntPhysGoal__get_objects_moving_across_collisions. I don't want it to hold this fix up.

deanwetherby:
You are correct. It is unrelated to this ticket.

chrislong:
The "could not find path from target object to goal" is I think expected to happen sometimes, because objects could be in the way and prevent the agent from reaching the target. The test should probably do what the actual code does, which is retry with a new scene.

chrislong:
@deanwetherby You had a comment that said "unit tests". There is a `test_TransferralGoal_ensure_pickup_action` for this ticket. Were there other tests you had in mind?

deanwetherby:
I think I pulled my initial approval because that specific test was failing occasionally. I should have been more clear in the comment.

PR #130, opened 2020-06-18
MCS-111

arashBehpour:
Take a look, let me know if I am misunderstood anything for the ticket. Also I was not sure to include random.seed(self.__seed) in on_init() to initiate the the random number generator within the mcs_controller_ai2thor.py file

### DIFF
--- a/python_api/machine_common_sense/mcs_controller_ai2thor.py
+++ b/python_api/machine_common_sense/mcs_controller_ai2thor.py
@@ -105,9 +105,9 @@ class MCS_Controller_AI2THOR(MCS_Controller):
     CONFIG_METADATA_MODE_NO_VISION = 'no_vision' # No vision (image feature) metadata, except for the images
     CONFIG_METADATA_MODE_NONE = 'none' # No metadata, except for the images and haptic/audio feedback
 
-    def __init__(self, unity_app_file_path, debug=False, enable_noise=False, seed=None):
+    def __init__(self, unity_app_file_path, debug=False, enable_noise=False):#, seed=None):
         super().__init__()
-
+        print("HEREE22222")
         self.__controller = ai2thor.controller.Controller(
             quality='Medium',
             fullscreen=False,
@@ -124,15 +124,16 @@ class MCS_Controller_AI2THOR(MCS_Controller):
                 "objects": []
             }
         )
+        print("HEREE333333")
+        self.on_init(debug, enable_noise)#, seed)
 
-        self.on_init(debug, enable_noise, seed)
-
-    def on_init(self, debug=False, enable_noise=False, seed=None):
+    def on_init(self, debug=False, enable_noise=False):#, seed=None):
+        print("HEREE44444")
         self.__debug_to_file = True if (debug is True or debug is 'file') else False
         self.__debug_to_terminal = True if (debug is True or debug is 'terminal') else False
 
         self.__enable_noise = enable_noise
-        self.__seed = seed
+        #self.__seed = seed
         """FIXME
         if not self.__seed == None:
             random.seed(self.__seed)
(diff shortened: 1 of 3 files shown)


python_api/machine_common_sense/mcs_controller_ai2thor.py
```diff
@@ -124,14 +124,19 @@ def __init__(self, unity_app_file_path, debug=False, enable_noise=False):
...
+    def on_init(self, debug=False, enable_noise=False, seed=None):
+        
         self.__debug_to_file = True if (debug is True or debug is 'file') else False
         self.__debug_to_terminal = True if (debug is True or debug is 'terminal') else False
 
         self.__enable_noise = enable_noise
+        self.__seed = seed
+        
+        #if not self.__seed == None:
+            #random.seed(self.__seed)
```
chrislong, line 139: It seems weird to have an optional argument that causes an error if it's omitted. If it's actually required, it would be better to make it a required argument.

chrislong, line 139: Why not delete these two lines?

arashBehpour: I will remove these comments! I originally thought to initialize the seed within the controller class, so that the random number generator would already be set. However the user can just call the controllers get_seed() to grab it.

arashBehpour:
Yeah I can do that, however I think random.seed is called every time before using the random function to generate a value. Like shown here https://www.geeksforgeeks.org/random-seed-in-python/ I can still include it though!

ThomasSchellenbergNextCentury:
Hmm, it appears that `random.seed` doesn't work as I expected.  In addition to needing to call it repeatedly, it also only affects the `random` library, and not other libraries like those from `numpy` or `tensorflow`.  These complications kind of defeat the purpose of this ticket.  Any ideas, maybe @deanwetherby or @chrislong ?

chrislong:
I'm not sure what you mean about calling `random.seed` repeatedly. Normally you'd call it once at the start and then you'd get the same sequence of random numbers every time.
As for numpy and tensorflow, I haven't used messed with their random number generation at all. It looks like you can set the seed values for them, though (https://numpy.org/doc/stable/reference/random/generated/numpy.random.RandomState.seed.html?highlight=numpy%20random%20seed#numpy.random.RandomState.seed and https://www.tensorflow.org/api_docs/python/tf/random/set_seed). It looks kinda complicated for tensorflow, unfortunately, but I think if you're looking for repeatable, deterministic results, I think it should be doable.

ThomasSchellenbergNextCentury:
@chrislong According to the 2nd example on this page, it looks like you need to call `random.seed` before each random call (unless I'm misunderstanding it?) https://pynative.com/python-random-seed/

Also, my other concern was that, for every library X that the TA1s want to use, we'd need to call X's "set random seed" function in our code.  So we'd need to know all possible values of X, and hope that each has a "set random seed" function.

chrislong:
> @chrislong According to the 2nd example on this page, it looks like you need to call `random.seed` before each random call (unless I'm misunderstanding it?) https://pynative.com/python-random-seed/

@ThomasSchellenbergNextCentury It depends on what you want. I thought what we wanted was a deterministic sequence of random numbers, which you can get by calling `seed` once, at the start. If you want your random number generator to return the exact same value every time you call it, you can call `seed` before every call, but I don't see why you'd want or need that.
 
> Also, my other concern was that, for every library X that the TA1s want to use, we'd need to call X's "set random seed" function in our code. So we'd need to know all possible values of X, and hope that each has a "set random seed" function.

I didn't realize we were trying to make their code deterministic, too. I wouldn't even guarantee that's possible, for example if they use threads internally. When I worked on a simulation for a DARPA evaluation before, we had to make our simulator (and the virtual environment) deterministic, but not the agents operating in the environment. They were not our problem.

ThomasSchellenbergNextCentury:
@chrislong  Thanks for clarifying that first point.  It makes sense now.

Yeah, my hope was to make the whole run deterministic, in case we wanted to run everything again.  Seems like it was just a pipe dream.  But what Arash has done is good enough for now.

Thanks for your input.

PR #134, opened 2020-06-22
MCS-231: Interaction Scenes Pair 5: Position Similar Object Either Adjacent or Not Adjacent

chrislong:


scene_generator/pairs.py
```diff
@@ -264,6 +264,63 @@ def get_scenes(self) -> Tuple[Dict[str, Any], Dict[str, Any]]:
...
+        else:
+            target_container_loc = geometry.calc_obj_pos(performer_position, [], container_def)
+            target_container = util.instantiate_object(container_def, target_container_loc)
+            target2 = util.instantiate_object(target_def, geometry.ORIGIN_LOCATION)
+            containers.put_object_in_container(target2, target_container, container_def, index, rot_a)
+
+            similar_container = instantiate_away_from(container_def, performer_position, target_container)
+            similar2 = util.instantiate_object(target_def, geometry.ORIGIN_LOCATION)
+            containers.put_object_in_container(similar2, similar_container, container_def, index, rot_b)
+            scene2['objects'] = [target2, similar2, target_container, similar_container]
```
ThomasSchellenbergNextCentury, line 320: Couldn't we move this code chunk to the `if` block starting on line 287?  Then we could also combine the two `if placement is None` blocks.

scene_generator/pairs.py
```diff
@@ -264,6 +264,63 @@ def get_scenes(self) -> Tuple[Dict[str, Any], Dict[str, Any]]:
         return scene1, scene2
 
 
+class SimilarAdjacentFarPair(InteractionPair):
```
ThomasSchellenbergNextCentury, line 267: If you're going to be making some changes anyway, would you mind adding the pair's scene descriptions here too please?

scene_generator/pairs.py
chrislong: No, because the `if` on 287 is nested inside the `if` on L279 (`if random.random() <= util.TARGET_CONTAINED_CHANCE`). There might be a way to restructure it in spite of that, but it seems complicated to me.

scene_generator/pairs.py
chrislong: Sure, no problem.

scene_generator/pairs.py
ThomasSchellenbergNextCentury: Can't `placement` only be `not None` if it's assigned within the `if random.random()` block?  Or am I just being dense?

chrislong: No, you're right. I don't know what I was thinking just now. I blame it on it being Friday afternoon. I'll move the code and combine the `if placement is None` blocks.

scene_generator/pairs_test.py
```diff
@@ -1,8 +1,14 @@
...
+import containers
+from geometry_test import are_adjacent
+from pairs import SimilarAdjacentPair, SimilarFarPair, SimilarAdjacentFarPair
+=======
 import shapely
 
 import geometry
 from geometry_test import are_adjacent
 from pairs import SimilarAdjacentPair, SimilarFarPair, ImmediatelyVisiblePair, HiddenBehindPair
+>>>>>>> master
```
ThomasSchellenbergNextCentury, line 11: You left some merge stuff here, and below.

chrislong: Nuts. Ok, thanks for letting me know. I'll fix it.

PR #135, opened 2020-06-22
MCS-281 Shape Constancy Scene Generator Use Same Object ID

ThomasSchellenbergNextCentury:
Use the same object ID for the target object in implausible shape constancy scenes from the scene generator.

### DIFF
--- a/scene_generator/quartets.py
+++ b/scene_generator/quartets.py
@@ -381,6 +381,7 @@ class ShapeConstancyQuartet(Quartet):
         b_def = random.choice(possible_defs)
         b_def = util.finalize_object_definition(b_def)
         b = util.instantiate_object(b_def, a['original_location'], a['materials_list'])
+        b['id'] = a['id']
         logging.debug(f'a type: {a["type"]}\tb type: {b["type"]}')
         return b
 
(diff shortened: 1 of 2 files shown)


deanwetherby:
@chrislong excellent suggestion!

ThomasSchellenbergNextCentury:
@deanwetherby @brianpippin So I just realized that, for this evaluation, we're actually going to need to write a little script.  We don't want to generate a dataset with different IDs, then generate a dataset with matching IDs, since the scenes will all be different, and we won't really know if the change in performance is due to the matching IDs or the differences in the scenes.  I think we need to generate a dataset, then have a script iterate over the dataset and create a second version of each scene.  In that case, I think we should merge this PR before we generate the dataset.

deanwetherby:
@ThomasSchellenbergNextCentury @brianpippin That's an excellent point about keeping the scene the same. I wonder if at some the scene generator couldn't build a base scene with objects and positions but no goal. Then the goal could be added in from the list of scene objects as a second stage. That doesn't help this particular situation though.

ThomasSchellenbergNextCentury:
I'll merge this branch now then.  I've created a new ticket to write the script.  It has some pseudocode if you want to take a look.  https://nextcentury.atlassian.net/browse/MCS-295

PR #136, opened 2020-06-23
MCS-274 Fix TA1 See-Through-Wall Bug

ThomasSchellenbergNextCentury:
Don't accidentally position the player to start inside a dynamically generated wall in the scene generator.

We needed to check for collision using the player's polygon bounding box, not just the player's center point.


scene_generator/util.py
```diff
@@ -216,3 +218,17 @@ def get_def_with_new_scale(obj: Dict[str, Any]) -> Dict[str, Any]:
     else:
         obj_def = None
     return obj_def
+
+
+def find_performer_rect(performer_position: Dict[str, float]) -> List[Dict[str, float]]:
```
chrislong, line 223: I think geometry would be a better module for this function

ThomasSchellenbergNextCentury: Sure.

PR #137, opened 2020-06-23
Mcs 285: Optimal path planner should account for agent reach

chrislong:


scene_generator/interaction_goals.py
```diff
@@ -150,12 +151,43 @@ def get_navigation_actions(start_location: Dict[str, Any],
...
+    step_dist = 0
+    dist_moved = 0
+    while total_distance < MAX_REACH_DISTANCE and i > 0:
+        i -= 1
+        action = actions[i]
+        if action['action'] != 'MoveAhead':
+            break
+        step_dist = action['params']['amount'] * MAX_MOVE_DISTANCE
+        total_distance += step_dist
+        dist_moved += step_dist
```
ThomasSchellenbergNextCentury, line 184: In what scenarios will we need to remove more than one Move action from the end of the action list?

scene_generator/interaction_goals.py
```diff
@@ -150,12 +151,43 @@ def get_navigation_actions(start_location: Dict[str, Any],
...
+
+def trim_actions_to_reach(actions: List[Dict[str, Any]],
+                          performer: Tuple[float, float],
+                          heading: float,
+                          goal_obj: Dict[str, Any]) -> \
+                          Tuple[List[Dict[str, Any]], Tuple[float, float]]:
+    """Trim the action list from the end so that the performer doesn't
+    take additonal steps toward the goal object once they're within
+    MAX_REACH_DISTANCE.
+    """
```
ThomasSchellenbergNextCentury, line 170: This doesn't appear to reduce the `amount` of the last Move action, like I'd have expected.  Is that not necessary?

scene_generator/interaction_goals.py
chrislong: Reach is greater than movement, and the last Move action might be for a small amount, so removing it and the previous action may sometimes still leave the agent within reach.

scene_generator/interaction_goals.py
chrislong: I thought it was sufficient to minimize the number of actions, not necessarily the distance traversed, but I could change it.

deanwetherby: Number of actions is what we are trying to reduce so reducing the amount isn't necessary although there's a fun optimization to consider the following action in order to be really optimal!

chrislong: I think I see what you mean. At least for some value of "fun".

scene_generator/interaction_goals.py
ThomasSchellenbergNextCentury: Oh, reach is greater than movement.  I had forgotten that.  Thanks for the explanation.

scene_generator/interaction_goals.py
ThomasSchellenbergNextCentury: That makes sense.  Thanks.

PR #141, opened 2020-06-24
MCS-236 Scene Generator Quartet/Pair ID Tag

ThomasSchellenbergNextCentury:
Added a unique ID shared between scenes in quartets or pairs so they can be filtered in the UI.

### DIFF
--- a/scene_generator/scene_generator.py
+++ b/scene_generator/scene_generator.py
@@ -162,6 +162,7 @@ def generate_quartet(prefix: str, index: int, quartet_name: str,
     quartet_class = quartets.get_quartet_class(quartet_name)
     quartet = quartet_class(template, find_path)
     quartet_id = str(uuid.uuid4())
+    quartet_name = quartet.__class__.__name__.replace('Quartet', '').lower()
     for q in range(1, 5):
         name = f'{prefix}-{index:04}-{q}.json'
         logging.debug(f'starting generation of\t{name}')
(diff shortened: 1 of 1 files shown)


scene_generator/scene_generator.py
```diff
@@ -206,10 +209,13 @@ def generate_pair(prefix: str, count: int,
     name_stem = f'{prefix}-{count:04}-'
     pair_class = pairs.get_pair_class()
     pair = pair_class(template, find_path)
+    pair_id = str(uuid.uuid4())
```
deanwetherby, line 212: Does a pair have to be universally unique?

ThomasSchellenbergNextCentury: You mean, opposed to counting them from 1 to 1000, or whatever?  I guess that's fine, though I was concerned that we may want to compare pairs or quartets across different evaluations, in which case I think a unique ID would be more useful.

chrislong: It occurs to me that it might also be useful to save which type of quartet or pair it was (e.g., SpatioTemporalContinuity). If so, I guess it could be this ticket or another one, whichever.

deanwetherby: Disregard. Interaction pairs absolutely need a uuid. I was thinking of something else looking at just the changes.

ThomasSchellenbergNextCentury: @chrislong  That IS a good idea!  I guess you're allowed to have those once in a while.  I'll add it here.

PR #142, opened 2020-06-24
MCS-289: Fix file format for upload, gzips unable to be downloaded properly

brianpippin:
gzip files were being wrapped with extra formatting, fix so that nothing wraps the files that are being uploaded.

### DIFF
--- a/homepage/js/upload_s3_button.js
+++ b/homepage/js/upload_s3_button.js
@@ -32,7 +32,7 @@ const uploadFileToS3 = (presignedPostData, file) => {
     return new Promise((resolve, reject) => {
         const xhr = new XMLHttpRequest();
         xhr.open("PUT", presignedPostData.url, true);
-        xhr.setRequestHeader("Content-Type", "multipart/form-data");
+        xhr.setRequestHeader("Content-Type", "application/octet-stream");
         xhr.onload = function () {
             this.status === 200 ? resolve() : reject(this.responseText);
         };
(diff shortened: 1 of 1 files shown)


homepage/js/upload_s3_button.js
```diff
@@ -30,10 +30,6 @@ const getPresignedPostData = selectedFile => {
  */
 const uploadFileToS3 = (presignedPostData, file) => {
     return new Promise((resolve, reject) => {
-        const formData = new FormData();
-
-        formData.append("file", file);
-
         const xhr = new XMLHttpRequest();
         xhr.open("PUT", presignedPostData.url, true);
         xhr.setRequestHeader("Content-Type", "multipart/form-data");
```
chrislong, line 35: Is this still the right content type? It seems like it should be some kind of binary type.

brianpippin: That seems like the generic one that all the AWS examples I was looking at, recommended, and it worked for all tar, gzip, and zip files I uploaded to S3.

brianpippin: I could try it with "application/octet-stream" if you think that would be a more generic one?

chrislong: It does seem to me like it _should_ be "application/octet-stream", but I'm not sure if it matters. According to this stackoverflow answer, S3 itself doesn't care: https://stackoverflow.com/questions/36301483/what-does-amazon-s3-use-the-content-type-header-for

brianpippin: Yeah I set it to octet-stream before I merged to be closer to what it should be, but you are right with what I saw about S3 not seeming to care.

PR #145, opened 2020-06-26
MCS-201 MCS-290 MCS-294 Tags for Eval 2 IntPhys Dataset

ThomasSchellenbergNextCentury:
Accomplishments:
- Scene generator tags will now identify novel colors, shapes, and combinations in object definitions.
- Scene generator object tags now differentiate between target, distractor, and other objects.
- Scene generator now lists all targets, distractors, occluders, walls, and background objects in scenes.
- Fixed the scene generator so that all IntPhys goal types may add background objects (a.k.a. "scenery"), not just gravity goals.  (This probably should have been in its own commit, but oh well.)

Apologies for the long PR.

### DIFF
--- a/scene_generator/goal.py
+++ b/scene_generator/goal.py
@@ -24,7 +24,7 @@ DIST_WALL_APART = 1
 SAFE_DIST_FROM_ROOM_WALL = 3.5
     
 
-def generate_wall(wall_mat_choice: str, performer_position: Dict[str, float],
+def generate_wall(wall_material: str, wall_colors: List[str], performer_position: Dict[str, float],
                   other_rects: List[List[Dict[str, float]]]) -> Optional[Dict[str, Any]]:
     # Wanted to reuse written functions, but this is a bit more of a special snowflake
     # Generates obstacle walls placed in the scene.
@@ -54,11 +54,13 @@ def generate_wall(wall_mat_choice: str, performer_position: Dict[str, float],
     if tries < util.MAX_TRIES:
         new_object = {
             'id': 'wall_' + str(uuid.uuid4()),
-            'materials': [wall_mat_choice],
+            'materials': [wall_material],
             'type': 'cube',
             'kinematic': 'true',
             'structure': 'true',
-            'mass': 100
+            'mass': 100,
+            'info': wall_colors,
+            'info_string': ' '.join(wall_colors)
         }
         shows_object = {
             'stepBegin': 0,
@@ -74,6 +76,7 @@ def generate_wall(wall_mat_choice: str, performer_position: Dict[str, float],
     return None
 
 
+
 class Goal(ABC):
     """An abstract Goal. Subclasses must implement compute_objects and
     get_config. Users of a goal object should normally only need to call 
@@ -82,20 +85,27 @@ class Goal(ABC):
     def __init__(self):
         self._performer_start = None
         self._targets = []
-        self._goal_objects = []
+        self._object_dict = []
 
     def update_body(self, body: Dict[str, Any], find_path: bool) -> Dict[str, Any]:
         """Helper method that calls other Goal methods to set performerStart, objects, and goal. Returns the goal body
         object."""
         body['performerStart'] = self.compute_performer_start()
-        goal_objects, all_objects, bounding_rects = self.compute_objects(body['wallMaterial'])
-        self._goal_objects = goal_objects
-        walls = self.generate_walls(body['wallMaterial'], body['performerStart']['position'],
-                                    bounding_rects)
-        body['objects'] = all_objects + walls
-        body['goal'] = self.get_config(goal_objects, all_objects + walls)
+
+        object_dict, bounding_rects = self.compute_objects(body['wallMaterial'])
+        self._object_dict = object_dict
+        self._targets = object_dict['target']
+
+        walls = self.generate_walls(body['wallMaterial'], body['wallColors'], body['performerStart']['position'], \
+                bounding_rects)
+        if walls is not None:
+            object_dict['wall'] = walls
+
+        body['objects'] = [element for key, value in object_dict.items() for element in value]
+        body['goal'] = self.get_config(self._targets, object_dict)
+
         if find_path:
-            body['answer']['actions'] = self.find_optimal_path(goal_objects, all_objects + walls)
+            body['answer']['actions'] = self.find_optimal_path(self._targets, body['objects'])
 
         return body
 
@@ -123,8 +133,8 @@ class Goal(ABC):
         return util.finalize_object_definition(random.choice(object_def_list))
 
     @abstractmethod
-    def compute_objects(self, wall_material_name: str) \
-        -> Tuple[List[Dict[str, Any]], List[Dict[str, Any]], List[List[Dict[str, float]]]]:
+    def compute_objects(self, wall_material_name: str) -> \
+            Tuple[Dict[str, List[Dict[str, Any]]], List[List[Dict[str, float]]]]:
         """Compute object instances for the scene. Returns a tuple:
         (objects required for the goal, all objects in the scene including objects required for the goal, bounding rectangles)"""
         pass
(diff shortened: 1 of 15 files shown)


scene_generator/goal.py
```diff
@@ -82,20 +85,27 @@ class Goal(ABC):
     def __init__(self):
         self._performer_start = None
         self._targets = []
-        self._goal_objects = []
+        self._object_dict = []
```
deanwetherby, line 88: If this is a dict, you should initialize with curly brackets {}. Do we need '_dict' in the variable name? It could just be 'self._object'.

scene_generator/goal.py
```diff
@@ -143,26 +153,65 @@ def add_objects(self, object_list: List[Dict[str, Any]], rectangles: List[List[D
...
         info_set = set(goal.get('info_list', []))
-        for obj in all_objects:
-            info_set |= frozenset(obj.get('info', []))
+
+        for key, value in object_dict.items():
+            for obj in value:
+                info_list = obj.get('info', []).copy()
+                if 'info_string' in obj:
+                    info_list.append(obj['info_string'])
+                info_set |= frozenset([(key + ' ' + info) for info in info_list])
```
deanwetherby, line 164: Is there a reason for using frozenset over set here that I'm not seeing? Using a frozenset implies you are doing something special with it like using it as a key into a dictionary. If we're just finding the union of two sets, the set type will work fine. There are no performance implications that I am aware of but maybe you can enlighten me.

scene_generator/goal.py
ThomasSchellenbergNextCentury: You're right concerning the curly brackets.  I wanted the name to indicate that it's a collection of objects.  Originally I called it `object_collection` but that seemed long.  Maybe just `object_map`?  (It maps specific tags like `target` or `distractor` to lists of objects.)

scene_generator/goal.py
ThomasSchellenbergNextCentury: This was copied from elsewhere so I'd ask @chrislong

deanwetherby: Thanks.

scene_generator/interaction_goals.py
```diff
@@ -328,7 +318,7 @@ class RelationshipType(Enum):
     def __init__(self):
         super(TransferralGoal, self).__init__()
 
-    def _set_goal_objects(self) -> None:
+    def _generate_additional_target_objs(self) -> None:
```
deanwetherby, line 321: Does the return type hint need to be a List now?

scene_generator/intphys_goals.py
```diff
@@ -115,27 +115,71 @@ def find_optimal_path(self, goal_objects: List[Dict[str, Any]], all_objects: Lis
...
+            # Choose values so the scenery is placed between the
+            # moving IntPhys objects and the room's wall.
+            return util.random_real(MIN_Z, MAX_Z, util.MIN_RANDOM_INTERVAL)
+
+        scenery_count = random.choices((0, 1, 2, 3, 4, 5),
+                                             (50, 10, 10, 10, 10, 10))[0]
+        scenery_list = []
+        scenery_rects = []
+        scenery_defs = objects.OBJECTS_MOVEABLE + objects.OBJECTS_IMMOBILE
+        for i in range(scenery_count):
```
deanwetherby, line 164: If 'i' isn't used to index anything, should we replace with an underscore to show the variable's insignificance?

scene_generator/intphys_goals.py
```diff
@@ -183,8 +227,8 @@ def _get_paired_occluder(self, paired_obj, occluder_list, non_room_wall_material
                 occluder_fits = True
                 break
         if occluder_fits:
-            occluder_objs = objects.create_occluder(random.choice(non_room_wall_materials)[0],
-                                                    random.choice(pole_materials)[0],
+            occluder_objs = objects.create_occluder(random.choice(non_room_wall_materials),
+                                                    random.choice(pole_materials),
```
deanwetherby, line 231: Not a review per se but a thank you for getting rid of those square brackets.

scene_generator/goal.py
chrislong: I don't remember any more why it's a frozenset specifically. As you pointed out, I think it doesn't matter and could be either set or frozenset.

scene_generator/goal.py
deanwetherby: I see. Yes a better name would be good here. object_map is fine with me.

scene_generator/goal.py
```diff
@@ -82,20 +85,27 @@ class Goal(ABC):
...
-        goal_objects, all_objects, bounding_rects = self.compute_objects(body['wallMaterial'])
-        self._goal_objects = goal_objects
-        walls = self.generate_walls(body['wallMaterial'], body['performerStart']['position'],
-                                    bounding_rects)
-        body['objects'] = all_objects + walls
-        body['goal'] = self.get_config(goal_objects, all_objects + walls)
+
+        object_dict, bounding_rects = self.compute_objects(body['wallMaterial'])
+        self._object_dict = object_dict
+        self._targets = object_dict['target']
```
chrislong, line 97: Nitpick: its it one target or two? Seems like the names should match

scene_generator/goal.py
```diff
@@ -82,20 +85,27 @@ class Goal(ABC):
...
+
+        walls = self.generate_walls(body['wallMaterial'], body['wallColors'], body['performerStart']['position'], \
+                bounding_rects)
+        if walls is not None:
+            object_dict['wall'] = walls
+
+        body['objects'] = [element for key, value in object_dict.items() for element in value]
```
chrislong, line 104: You don't care about the keys, right? So I think you could use `values()` instead of `items()` and simplify slightly.

scene_generator/goal.py
```diff
@@ -82,20 +85,27 @@ class Goal(ABC):
...
-        body['goal'] = self.get_config(goal_objects, all_objects + walls)
+
+        object_dict, bounding_rects = self.compute_objects(body['wallMaterial'])
+        self._object_dict = object_dict
+        self._targets = object_dict['target']
+
+        walls = self.generate_walls(body['wallMaterial'], body['wallColors'], body['performerStart']['position'], \
+                bounding_rects)
+        if walls is not None:
+            object_dict['wall'] = walls
```
chrislong, line 102: If `object_dict['wall']` is a list, I think `walls` would be better.

scene_generator/goal.py
```diff
@@ -123,8 +133,8 @@ def choose_object_def(self) -> Dict[str, Any]:
         return util.finalize_object_definition(random.choice(object_def_list))
 
     @abstractmethod
-    def compute_objects(self, wall_material_name: str) \
-        -> Tuple[List[Dict[str, Any]], List[Dict[str, Any]], List[List[Dict[str, float]]]]:
+    def compute_objects(self, wall_material_name: str) -> \
+            Tuple[Dict[str, List[Dict[str, Any]]], List[List[Dict[str, float]]]]:
         """Compute object instances for the scene. Returns a tuple:
         (objects required for the goal, all objects in the scene including objects required for the goal, bounding rectangles)"""
```
chrislong, line 139: Update the documentation to match the new return value. (The only thing worse than no documentation is documentation that lies. ;-) )

scene_generator/goal.py
```diff
@@ -143,26 +153,65 @@ def add_objects(self, object_list: List[Dict[str, Any]], rectangles: List[List[D
                 obj = util.instantiate_object(object_def, obj_location)
                 object_list.append(obj)
 
-    def _update_goal_info_list(self, goal: Dict[str, Any], all_objects: List[Dict[str, Any]]):
+    def _update_goal_info_list(self, goal: Dict[str, Any], object_dict: Dict[str, List[Dict[str, Any]]]) -> None:
         info_set = set(goal.get('info_list', []))
-        for obj in all_objects:
-            info_set |= frozenset(obj.get('info', []))
+
+        for key, value in object_dict.items():
```
chrislong, line 159: Again, could simplify slightly with `values()`

scene_generator/goal.py
ThomasSchellenbergNextCentury: Sounds good.

scene_generator/goal.py
ThomasSchellenbergNextCentury: I can change that.

scene_generator/interaction_goals.py
ThomasSchellenbergNextCentury: Ah, good catch.  Thanks.

scene_generator/intphys_goals.py
ThomasSchellenbergNextCentury: Sounds good.

scene_generator/goal.py
ThomasSchellenbergNextCentury: Yes, it's a list.  The keys in the object_dict are used to create new tags, so keeping it singular is necessary.

scene_generator/goal.py
ThomasSchellenbergNextCentury: Good idea.

scene_generator/goal.py
ThomasSchellenbergNextCentury: Oh, didn't notice that before.  Thanks.

scene_generator/goal.py
ThomasSchellenbergNextCentury: Sounds good.

ThomasSchellenbergNextCentury: The key is used below on line 164

scene_generator/goal.py
ThomasSchellenbergNextCentury: Due to the comments from @chrislong concerning the keys matching the values in the dict, I'm going to rename this variable to `tag_to_objects` which will hopefully emphasize that the values are indeed lists.

scene_generator/intphys_goals.py
```diff
@@ -506,60 +550,20 @@ def is_left_to_right(self) -> bool:
...
         self._ramp_type = ramp_type
         self._left_to_right = left_to_right
-        scenery = self._compute_scenery()
-        goal_objs = objs + ramp_objs
-        return goal_objs, goal_objs + scenery, []
+        background_objs = self._compute_scenery()
+
+        return {
+            'target': [moving_objs[0]],
+            'distractor': moving_objs[1:] if len(moving_objs) > 1 else [],
```
chrislong, line 563: FYI, you don't need the `if`...`else` here. If `moving_objs` has only one member, `moving_objs[1:]` will be an empty list.

scene_generator/quartets.py
```diff
@@ -11,20 +11,20 @@
...
     """Find a 'target' object in the scene. (IntPhys goals don't really
     have them, but they do have objects that may behave plausibly or
     implausibly.)
     """
-    return find_targets(scene, 1)[0]
+    return find_targets(scene, goal, 1)[0]
 
 
-def find_targets(scene: Dict[str, Any], num_targets: Optional[int] = None) -> List[Dict[str, Any]]:
+def find_targets(scene: Dict[str, Any], goal: goal.Goal, num_targets: Optional[int] = None) -> List[Dict[str, Any]]:
```
chrislong, line 22: At this point, in a perfect world, it seems like `find_targets` (and `find_target`) would be methods of `Goal`. I'm not sure if it's worth refactoring, though.

scene_generator/quartets.py
```diff
@@ -116,7 +116,7 @@ def get_scene(self, q: int) -> Dict[str, Any]:
                 self._appear_behind_occluder(scene)
             elif q == 4:
                 # target not in the scene (plausible)
-                target_id = scene['goal']['metadata']['objects'][0]
+                target_id = self._goal._targets[0]
```
chrislong, line 119: Is `_target` a list of IDs or objects? I thought it was a list of objects but maybe I'm wrong.

scene_generator/util.py
```diff
@@ -74,39 +77,67 @@ def instantiate_object(object_def: Dict[str, Any],
...
-    info = info[:1] + list(colors) + info[1:]
+    new_object['info'] = new_object['info'][:1] + list(colors) + new_object['info'][1:]
 
     if 'pickupable' in object_def['attributes']:
         size = 'light'
     elif 'moveable' in object_def['attributes']:
         size = 'heavy'
     else:
         size = 'massive'
-    info = info[:1] + [size] + info[1:]
```
chrislong, line 106: I think i't easier to read with a temporary variable `info` rather than `new_object['info']` everywhere, but YMMV.

scene_generator/intphys_goals.py
ThomasSchellenbergNextCentury: Ah, I was wondering about that but failed to find an answer with google.  Thanks for the FYI!

scene_generator/quartets.py
ThomasSchellenbergNextCentury: `_targets` is definitely always a list.  I'm not as sure about `_target`, but I think it's always a single object.

scene_generator/quartets.py
ThomasSchellenbergNextCentury: Agreed on both points.

scene_generator/quartets.py
chrislong: Oops, I said `_target` when I meant `_targets`. The issue is that `target_id` is being used as a string (e.g,. compared to `obj['id']` on L122). But I _think_ `self._goal._targets[0]` resolves to an object, not a string. Right?

ThomasSchellenbergNextCentury: Oh, I see what you're talking about now.  I'll fix that.  Thanks.

ThomasSchellenbergNextCentury:
Whoops, I totally forgot that I had branch MCS-293 merging into this branch for its PR.  Should I revert it?  I don't want to make this PR any bigger than it already is.

deanwetherby:
@ThomasSchellenbergNextCentury It shouldn't be a problem merging right? I wouldn't bother reverting if that's the case.

scene_generator/quartets.py
brianpippin: Agree as well, should we create a ticket to refactor that in the future after this eval or is it not worth refactoring at all?

deanwetherby: Ideally it would be best to tackle these as we encounter them but our current time constraint doesn't give us a lot of leeway for that. Most likely we will revisit our code base and tackle some of this tech debt after the evaluation. That's my take anywa.

ThomasSchellenbergNextCentury: https://nextcentury.atlassian.net/browse/MCS-300

PR #146, opened 2020-06-26
Mcs 93

arashBehpour:
Did not think the enum structure needed to changed. Was thinking about compressing it all into a class or have aenum(multiple attributes). But I think its fine since the commands are a set of finite values. Let me know If you all think it would be better to have MCS_Action as a class with self.shortcut_name and self.description instead.


python_api/machine_common_sense/getch.py
```diff
@@ -0,0 +1,40 @@
...
+            self.impl = _GetchWindows()
+        except ImportError:
+            self.impl = _GetchUnix()
+
+    def __call__(self): return self.impl()
+
+
+class _GetchUnix:
+    def __init__(self):
+        import tty, sys
```
chrislong, line 17: Looks like this is not used here.

python_api/machine_common_sense/run_mcs_human_input.py
```diff
@@ -16,6 +21,122 @@ def __init__(self, name, key, desc):
         self.key = key
         self.desc = desc
 
+class HumanInputShell(cmd.Cmd):
+
+    prompt = '(command)->'
+
+    controller = None
+    previous_output = None
+    config_data = None
```
chrislong, line 30: These are set in `__init__`, so why set them here?

python_api/machine_common_sense/getch.py
```diff
@@ -0,0 +1,40 @@
+# Source: https://stackoverflow.com/questions/510357/python-read-a-single-character-from-the-user
```
chrislong, line 1: Wouldn't it be simpler to use the build-in curses library? Then you wouldn't need to special case different OSs.

arashBehpour: I spent some time trying to learn and use the curses library to do the same functionality, but it kept interrupting the print output that would be shown after every executed command. There is probably a way to use curses library, but for me was not simpler than using the getch() class. With playing around with curses I believe I have messed up outputting on my terminal. But let me know if you try to get it to work with curses!

python_api/machine_common_sense/run_mcs_human_input.py
arashBehpour: You are right! Fixed it.

python_api/machine_common_sense/getch.py
arashBehpour: Its for using getch() on a UNIX system.

python_api/machine_common_sense/run_mcs_human_input.py
```diff
@@ -19,6 +19,79 @@ def __init__(self, name, key, desc):
...
+    config_data = None 
+
+    userInput = None #Will be re-initalized everytime input is read in
+
+    def __init__(self, input_controller, input_previous_output, input_config_data):
+        super(HumanInputShell,self).__init__()
+        
+        controller = input_controller
+        previous_output = input_previous_output
+        config_data = input_config_data
```
deanwetherby: Are these class variables? Don't you need to prefix with 'self'?

python_api/machine_common_sense/getchHelper.py
```diff
@@ -0,0 +1,73 @@
+# Source: https://stackoverflow.com/questions/510357/python-read-a-single-character-from-the-user
```
ThomasSchellenbergNextCentury, line 1: Thanks for linking the source here!

python_api/machine_common_sense/run_mcs_human_input.py
```diff
@@ -18,6 +19,115 @@ def __init__(self, name, key, desc):
...
+                print('  ' + action)
+        else:
+            print('All actions are available during this step.')
+
+        """
+        if self.previous_output.action_list is not None and len(self.previous_output.action_list) == 1:
+            print('Automatically selecting the only available action...')
+            userInput = self.previous_output.action_list[1]
+        else:
+        """
```
ThomasSchellenbergNextCentury, line 55: Is there a reason why this code is commented out?

python_api/machine_common_sense/run_mcs_human_input.py
```diff
@@ -18,6 +19,115 @@ def __init__(self, name, key, desc):
...
+        print("Press key 'e' to exit\n")
+        list_of_mcs_action_keys = [mcs_action_key.value for mcs_action_key in MCS_Action_Keys]
+
+        while True:
+            char = getch.__call__()
+            print('\n(shortcut-command)->', char)
+            if char == 'e': # exit shortcut key mode
+                break
+            elif char in list_of_mcs_action_keys:
+                self.default(char)
```
ThomasSchellenbergNextCentury, line 126: Is "shortcut mode" currently used anywhere?

python_api/machine_common_sense/run_mcs_human_input.py
```diff
@@ -18,6 +19,115 @@ def __init__(self, name, key, desc):
...
+
+        if params is None:
+            print("ERROR: Parameters should be separated by commas, and look like this example: rotation=45")
+            return
+
+        output = self.controller.step(action, **params)
+        self.previous_output = output
+    
+    #def emptyline(self):
+        #pass
```
ThomasSchellenbergNextCentury, line 91: You can remove this unused code.

python_api/machine_common_sense/run_mcs_human_input.py
```diff
@@ -40,79 +150,26 @@ def print_commands():
     print(" ")
     print("----------------- Other Commands -----------------")
     print(" ")
-    print("Enter 'help' to print the commands again.")
+    print("Enter 'print' to print the commands again.")
     print("Enter 'reset' to reset the scene.")
     print("Enter 'exit' to exit the program.")
```
ThomasSchellenbergNextCentury, line 155: Please add some sentences about the new changes, like about using the up/down arrow keys to access previous actions.

ThomasSchellenbergNextCentury:
@arashBehpour  The main change in this PR is being able to use the up/down arrow keys, right?  That functionality looks great, but I just wanted to know if there's anything else I should be testing.

Also, I have some thoughts about the Action enums, but I think they deserve their own ticket, so don't worry about them right now.

PR #147, opened 2020-06-27
MCS-293 Added novel objects to IntPhys falling-down scenes.

ThomasSchellenbergNextCentury:
- Added ducks and racecars as new options to IntPhys "falling down" scenes.
- Fixed a few minor bugs with IntPhys Shape Constancy quartets.

### DIFF
--- a/scene_generator/quartets_test.py
+++ b/scene_generator/quartets_test.py
@@ -1,5 +1,6 @@
 import intphys_goals
 import objects
+import pytest
 from quartets import ShapeConstancyQuartet, ObjectPermanenceQuartet, SpatioTemporalContinuityQuartet, find_targets, get_position_step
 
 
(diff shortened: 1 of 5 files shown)


scene_generator/quartets_test.py
```diff
@@ -69,8 +69,8 @@ def test_ShapeConstancyQuartet():
     assert quartet is not None
     a = quartet._scenes[0]['objects'][0]
     assert a['type'] != quartet._b['type']
-    assert a['shows'][0]['scale']['x'] == quartet._b['shows'][0]['scale']['x']
-    assert a['shows'][0]['scale']['z'] == quartet._b['shows'][0]['scale']['z']
+    assert (a['dimensions']['x'] > (quartet._b['dimensions']['x'] - 0.1) and
+            a['dimensions']['x'] < (quartet._b['dimensions']['x'] + 0.1))
```
chrislong, line 73: You could use the new constant here instead of 0.1, right? (Or, if I'm understanding the test correctly, you could use `pytest.approx` with its `abs` parameter., like `a['dimensions']['x'] == pytest.approx(quartet._b['dimensions']['x'], abs=0.1)`)

ThomasSchellenbergNextCentury: Ah, I didn't know about `approx`.  Thanks for the tip!  It's updated.

PR #148, opened 2020-06-29
MCS-302 Add tags to IntPhys quartets in scene generator.

ThomasSchellenbergNextCentury:


scene_generator/quartets.py
```diff
@@ -53,10 +53,16 @@ def get_position_step(target: Dict[str, Any], x_position: float,
...
+    def __init__(self, template: Dict[str, Any], find_path: bool, goal: intphys_goals.IntPhysGoal):
         self._template = template
         self._find_path = find_path
         self._scenes: List[Optional[Dict[str, Any]]] = [None]*4
+        self._scenes[0] = copy.deepcopy(self._template)
+        self._goal = goal
+        self._goal.update_body(self._scenes[0], self._find_path)
+        self._scenes[1] = copy.deepcopy(self._scenes[0])
+        self._scenes[2] = copy.deepcopy(self._scenes[0])
+        self._scenes[3] = copy.deepcopy(self._scenes[0])
```
ThomasSchellenbergNextCentury, line 65: Need to create all the scenes here on init rather than in `get_scene` so that we can add tags to `_scenes[0]` without affecting the other scenes.

scene_generator/intphys_goals.py
```diff
@@ -645,8 +664,8 @@ def _get_ramp_and_objects(self, room_wall_material_name: str) -> \
 class ObjectPermanenceGoal(IntPhysGoal):
     TEMPLATE = {
         'category': 'intphys',
-        'domain_list': ['objects', 'object_solidity', 'object_motion', 'object_permanence'],
-        'type_list': ['passive', 'action_none', 'intphys', 'object_permanence'],
+        'domain_list': ['objects', 'object solidity', 'object motion', 'object permanence'],
+        'type_list': ['passive', 'action none', 'intphys', 'object permanence'],
```
ThomasSchellenbergNextCentury, line 668: We don't need underscores for any technical reason, and I think the tags will look better without underscores in the UI.

scene_generator/quartets.py
```diff
@@ -174,7 +178,6 @@ def _find_other_occluder(self, target: Dict[str, Any], scene: Dict[str, Any]) ->
         return other_occluder, other_object
 
     def _teleport_forward(self, scene: Dict[str, Any]) -> None:
-        scene['answer']['choice'] = 'implausible'
```
ThomasSchellenbergNextCentury, line 177: All the changes to `scene['answer']['choice']` were moved into `get_scene`.

scene_generator/quartets.py
```diff
@@ -65,10 +71,7 @@ def get_scene(self, q: int) -> Dict[str, Any]:
 
 class ObjectPermanenceQuartet(Quartet):
     def __init__(self, template: Dict[str, Any], find_path: bool):
-        super(ObjectPermanenceQuartet, self).__init__(template, find_path)
-        self._goal = intphys_goals.ObjectPermanenceGoal()
-        self._scenes[0] = copy.deepcopy(self._template)
-        self._goal.update_body(self._scenes[0], self._find_path)
+        super(ObjectPermanenceQuartet, self).__init__(template, find_path, intphys_goals.ObjectPermanenceGoal())
```
ThomasSchellenbergNextCentury, line 74: Adding a new instantiation of the goal to the superclass constructor call here shouldn't cause any issues, right?  (I didn't see anything wrong in my testing.)

scene_generator/intphys_goals.py
```diff
@@ -545,6 +551,19 @@ def is_ramp_steep(self) -> bool:
...
+        elif self._ramp_type == ramps.Ramp.RAMP_45:
+            return '45-degree'
+        elif self._ramp_type == ramps.Ramp.RAMP_90:
+            return '90-degree'
+        elif self._ramp_type == ramps.Ramp.RAMP_30_90:
+            return '30-degree-90-degree'
+        elif self._ramp_type == ramps.Ramp.RAMP_45_90:
+            return '45-degree-90-degree'
+        return ''
+
```
chrislong, line 566: I'd put this in the Ramp enum, but I guess it doesn't really matter.

scene_generator/quartets.py
```diff
@@ -104,36 +107,37 @@ def get_scene(self, q: int) -> Dict[str, Any]:
...
+        elif q == 4:
+            # target not in the scene (plausible)
+            scene['goal']['type_list'].append('object permanence hide object')
+            target_id = self._goal._targets[0]['id']
+            for i in range(len(scene['objects'])):
+                obj = scene['objects'][i]
+                if obj['id'] == target_id:
+                    del scene['objects'][i]
+                    break
+        self._scenes[q - 1] = scene
```
chrislong, line 131: You don't need this assignment any more, right?

Note that this function should no longer be called more than once per index (whereas before that was ok to do), otherwise you'll get duplicate stuff on type_list, and maybe other side effects, too.

ThomasSchellenbergNextCentury: Ah, you're right, I missed that one.  And good note about calling the function multiple times.  I might rethink that.

ThomasSchellenbergNextCentury: @chrislong OK, I changed the implementation so that we save a `_scene_template` in the constructor and use that to create each new scene in `get_scene`.  Additionally, `get_scene` will return the previously created scene if it exists.

chrislong: Cool. It just seems safer to me if it's ok to call multiple times (esp. when it's named `get`*, implying that it just fetches a value, and getters usually return the same thing each time).

ThomasSchellenbergNextCentury: Yeah, it was a good idea!

scene_generator/intphys_goals_test.py
```diff
@@ -6,19 +7,143 @@
...
+  'objects': [],
+  'goal': {},
+  'answer': {}
+}
 
-def test_random_real():
-    n = random_real(0, 1, 0.1)
-    assert 0 <= n <= 1
-    # need to multiply by 10 and mod by 1 instead of 0.1 to avoid weird roundoff
-    assert n * 10 % 1 < 1e-8
```
ThomasSchellenbergNextCentury, line 14: Just moved this test to `util_test` since `random_real` was moved to `util` sometime in the past.

scene_generator/quartets_test.py
```diff
@@ -15,18 +18,48 @@ def test_get_position_step():
...
+
+def test_STCQ_get_scene_1():
+    quartet = SpatioTemporalContinuityQuartet(TEMPLATE, False)
+    scene = quartet.get_scene(1)
+    target = find_targets(scene, quartet._goal)[0]
+    assert scene['answer']['choice'] == 'plausible'
+    assert 'spatio temporal continuity move earlier' in scene['goal']['type_list']
+    assert 'hides' not in target
+    assert 'teleports' not in target
+    # TODO MCS-82 More asserts to test specific behavior
```
ThomasSchellenbergNextCentury, line 53: I limited what I put in these new tests based on the scope of my ticket.  Added TODOs as appropriate.

scene_generator/quartets.py
```diff
@@ -330,16 +330,24 @@ def _move_later(self, scene: Dict[str, Any]) -> None:
...
+        elif q == 2:
+            scene['answer']['choice'] = 'implausible'
+            scene['goal']['type_list'].append('spatio temporal continuity teleport forward')
+            self._teleport_forward(scene)
+        elif q == 3:
+            scene['answer']['choice'] = 'implausible'
+            scene['goal']['type_list'].append('spatio temporal continuity teleport backward')
+            self._teleport_backward(scene)
+        elif q == 4:
+            scene['goal']['type_list'].append('spatio temporal continuity move later')
```
brianpippin, line 347: I know this comment is super late as I'm half way through this code review and I looked at all the tags in the super long code review, but I feel like we should have an enum or file that has all the psychology tags in the future, that way if we need to change the wording or they want them to be slightly different, they would all be in the same file and we'd just update that file instead of hunting for the tags in the code.  Just a future improvement suggestion.

ThomasSchellenbergNextCentury: Hey, that's a super good idea.  I've created a new ticket to do that:  https://nextcentury.atlassian.net/browse/MCS-303

Thanks!

scene_generator/intphys_goals.py
deanwetherby: I think @chrislong is suggesting using the string as the value for the enum rather than using auto(). Then you can just return self._ramp_type.value instead of having all the ifs here.

ThomasSchellenbergNextCentury: Ah, OK.  I've made the change.

PR #149, opened 2020-07-02
MCS-296 Cleanup of Interactive Goals

ThomasSchellenbergNextCentury:
- MCS-296: All InteractionGoals now have the possibility of adding 0-10 distractors to the scene.
- Created ObjectRule classes that are used by InteractionGoals to describe the number of target and distractor objects needed by the goal type, as well as how those objects are selected and positioned.
  - This simplifies the code for the InteractiveGoals a lot.
  - This will be very helpful for MCS-306 when each InteractionPair will have its own InteractionGoal.  Each Pair will define specific Rules (the object definition must be similar to an existing object; the object must be positioned within specific parameters; the object must be positioned inside a container; etc.) and give them to its Goal.  Hopefully it will also be easy to mix-and-match Rules to create new Pairs in the future.

### DIFF
--- a/scene_generator/interaction_goals.py
+++ b/scene_generator/interaction_goals.py
@@ -198,76 +198,185 @@ def move_to_container(target: Dict[str, Any], bounding_rectangles: List[List[Dic
     return None
 
 
+class ObjectRule():
+    """Defines rules for how a specific object can be made and positioned as part of a specific goal."""
+
+    def __init__(self, position_inside_receptacle=False):
+        self._position_inside_receptacle = position_inside_receptacle
+
+    def choose_definition(self, target_list: List[Dict[str, Any]]) -> Dict[str, Any]:
+        """Choose and return an object definition."""
+        # Same chance to pick each list.
+        object_definition_list = random.choice(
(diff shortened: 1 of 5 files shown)


scene_generator/interaction_goals.py
```diff
@@ -178,86 +178,209 @@ def trim_actions_to_reach(actions: List[Dict[str, Any]],
...
+            raise ValueError(f'TransferToObjectRule cannot find any stack target object definition')
+
+        # Same chance to pick each list.
+        object_definition_list = random.choice(choice_list)
+        # Same chance to pick each object definition from the list.
+        return finalize_object_definition(random.choice(object_definition_list))
+
+    def choose_location(self, object_definition: Dict[str, Any], target_list: List[Dict[str, Any]], \
+            performer_start_position: Dict[str, float], bounding_rectangles: List[List[Dict[str, float]]]) -> \
+            Tuple[Dict[str, Any], List[List[Dict[str, float]]]]:
```
chrislong, line 278: Could use a doc string here to say at least what the return values are

PR #152, opened 2020-07-09
MCS-306 Add Interactive Goals to Scenes in Interactive Pairs

ThomasSchellenbergNextCentury:
This ticket still needs more testing but I wanted to open the PR so you all could start reviewing the changes and giving me feedback.

Task: MCS-306 (Each Interactive Pair should have an Interactive Goal)

Accomplishments:
- Reorganized the InteractionPair class and subclasses to generate an InteractionGoal and then use that Goal to generate the Pair's two scenes. Most of the behavior moved away from the subclasses into the InteractionPair class. The Pair will consider the Goal's requirements while generating its Target objects. The new structure should make it easier to add new Pairs in the future.
- Fixed some minor bugs with the existing Pairs. Some examples: if the Target and the Confuser were meant to switch being "immediately visible" in one scene and "not immediately visible" in another, they were incorrectly given new positions rather than just switching their positions; Pairs that should have a chance to containerize their Targets were never doing so.
- Each InteractionGoal subclass can customize its own Rules for generating Targets, Distractors, Confusors, and Obstructors. For example, in a TransferralGoal, the 1st Target must be pickupable, and the 2nd Target must have stackTarget.
- Minor cleanup in some functions relevant to the recent changes.

Outstanding Issues:
- Obstructors in the HiddenBehindPair aren't being positioned properly.

Solved Issues:
- Pair generation is still failing more often than it should.
- Some pair unit tests are broken because they rely on accessing specific objects at specific list indices and that behavior needs to be updated.
- The "cannot create enclosable receptacle" exceptions occur frequently due to the lack of similarityScale in object definitions (TODO MCS-261)
- Other misc. bug fixes.

### DIFF
--- a/scene_generator/geometry.py
+++ b/scene_generator/geometry.py
@@ -81,9 +81,9 @@ def collision(test_rect: List[Dict[str, float]], test_point: Dict[str, float]):
 
 
 def calc_obj_coords(position_x: float, position_z: float, delta_x: float, delta_z: float, offset_x: float,
-                    offset_z: float, rotation_y: float) -> List[Dict[str, float]]:
+                    offset_z: float, rotation: float) -> List[Dict[str, float]]:
     """Returns an array of points that are the coordinates of the rectangle """
-    radian_amount = math.pi * (2 - rotation_y / 180.0)
+    radian_amount = math.pi * (2 - rotation / 180.0)
 
     rotate_sin = math.sin(radian_amount)
     rotate_cos = math.cos(radian_amount)
@@ -157,7 +157,7 @@ def calc_obj_pos(performer_position: Dict[str, float],
             'rotation': {'x': rotation_x, 'y': rotation_y, 'z': rotation_z},
             'position':  {'x': new_x, 'y': obj_def.get('position_y', 0), 'z': new_z},
             'bounding_box': rect
-            }
+        }
         other_rects.append(rect)
         return new_object
 
@@ -189,7 +189,7 @@ def get_visible_segment(performer_start: Dict[str, Dict[str, float]]) \
         -> shapely.geometry.LineString:
     """Get a line segment that should be visible to the performer
     (straight ahead and at least MIN_START_DISTANCE_AWAY but within
-    the room).
+    the room). Return None if no visible segment is possible.
     """
     max_dimension = max(ROOM_DIMENSIONS[0][1] - ROOM_DIMENSIONS[0][0],
                         ROOM_DIMENSIONS[1][1] - ROOM_DIMENSIONS[1][0])
@@ -202,7 +202,8 @@ def get_visible_segment(performer_start: Dict[str, Dict[str, float]]) \
 
     target_segment = room.intersection(view_segment)
     if target_segment.is_empty:
-        raise exceptions.SceneException(f'performer too close to the wall, cannot place object in front of it (performer location={performer_start})')
+        logging.debug(f'performer too close to the wall, cannot place object in front of it (performer location={performer_start})')
+        return None
     return target_segment
 
 
@@ -211,6 +212,8 @@ def get_location_in_front_of_performer(performer_start: Dict[str, Dict[str, floa
                                        rotation_func: Callable[[], float] = random_rotation) \
                                        -> Optional[Dict[str, Any]]:
     visible_segment = get_visible_segment(performer_start)
+    if not visible_segment:
+        return None
 
     def segment_xz():
         fraction = random.random()
@@ -222,7 +225,7 @@ def get_location_in_front_of_performer(performer_start: Dict[str, Dict[str, floa
 
 
 def get_location_in_back_of_performer(performer_start: Dict[str, Dict[str, float]],
-                                  target_def: Dict[str, Any]) -> Optional[Dict[str, Any]]:
+                                      target_def: Dict[str, Any]) -> Optional[Dict[str, Any]]:
     # First, find the part of the the room that's behind the performer
     # (i.e., the 180 degree arc in the opposite direction from its
     # orientation)
@@ -311,6 +314,8 @@ def get_adjacent_location_on_side(obj_def: Dict[str, Any],
     performer = shapely.geometry.Point(performer_start['x'], performer_start['z'])
     room = get_room_box()
     if not bounding_box.intersects(performer) and room.contains(bounding_box):
+        offset_x = obj_def['offset']['x'] if 'offset' in obj_def else 0.0
+        offset_z = obj_def['offset']['z'] if 'offset' in obj_def else 0.0
         location = {
             'position': {
                 'x': x,
@@ -318,8 +323,11 @@ def get_adjacent_location_on_side(obj_def: Dict[str, Any],
                 'z': z
             },
             'rotation': {
-                'y': shows['rotation']['y']
-            }
+                'x': 0,
+                'y': shows['rotation']['y'],
+                'z': 0
+            },
+            'bounding_box': calc_obj_coords(x, z, dx, dz, offset_x, offset_z, shows['rotation']['y'])
         }
     else:
         location = None
--- a/scene_generator/geometry.py
+++ b/scene_generator/geometry.py
@@ -259,9 +259,8 @@ def get_location_in_back_of_performer(performer_start: Dict[str, Dict[str, float
     return calc_obj_pos(performer_start['position'], [], target_def, xz_func=compute_xz)
 
 
-def get_adjacent_location(obj_def: Dict[str, Any],
-                          target: Dict[str, Any],
-                          performer_start: Dict[str, float]) -> Optional[Dict[str, Any]]:
+def get_adjacent_location(obj_def: Dict[str, Any], target: Dict[str, Any], performer_start_position: Dict[str, float], \
+        behind: bool = False) -> Optional[Dict[str, Any]]:
     """Find a location such that, if obj_def is instantiated there, it
     will be next to target. Ensures that the object at the new
     location will not overlap the performer start, if necessary trying
@@ -270,9 +269,18 @@ def get_adjacent_location(obj_def: Dict[str, Any],
     sides = list(range(4))
     random.shuffle(sides)
     for side in sides:
-        location = get_adjacent_location_on_side(obj_def, target, performer_start, side)
-        if location is not None:
-            return location
+        location = get_adjacent_location_on_side(obj_def, target, performer_start_position, side)
+        if location:
+            # If behind, position the object so that the target is between it and the performer start.
+            if behind:
+                performer_start_coordinates = (performer_start_position['x'], performer_start_position['z'])
+                object_location_coordinates = (location['position']['x'], location['position']['z'])
+                line_to_object = shapely.geometry.LineString([performer_start_coordinates, object_location_coordinates])
+                target_poly = get_bounding_polygon(target)
+                if target_poly.intersects(line_to_object):
+                    return location
+            else:
+                return location
     return None
 
 
@@ -285,13 +293,13 @@ class Side(IntEnum):
 
 def get_adjacent_location_on_side(obj_def: Dict[str, Any],
                                   target: Dict[str, Any],
-                                  performer_start: Dict[str, float],
+                                  performer_start_position: Dict[str, float],
                                   side: Side) -> Optional[Dict[str, Any]]:
     """Get a location such that, if obj_def is instantiated there, it will
     be next to target. Side determines on which side of target to
     place it: 0 = right (positive x), 1 = behind (positive z), 2 =
     left (negative x) and 3 = in front (negative z). If the object
-    would overlap the performer_start or would be outside the room,
+    would overlap the performer_start_position or would be outside the room,
     None is returned."
     """
     GAP = 0.05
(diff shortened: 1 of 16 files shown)


scene_generator/pairs.py
```diff
@@ -37,152 +30,576 @@ def move_to_location(obj_def: Dict[str, Any], obj: Dict[str, Any],
...
+        print('target_1 ' + target_1['id'] + ' ' + target_1['type'])
+        print('target_2 ' + target_2['id'] + ' ' + target_2['type'])
+        print('target_receptacle_1 ' + ((target_receptacle_1['id'] + ' ' + target_receptacle_1['type']) if target_receptacle_1 else 'None'))
+        print('target_receptacle_2 ' + ((target_receptacle_2['id'] + ' ' + target_receptacle_2['type']) if target_receptacle_2 else 'None'))
+        print('confusor_1 ' + ((confusor_1['id'] + ' ' + confusor_1['type']) if confusor_1 else 'None'))
+        print('confusor_2 ' + ((confusor_2['id'] + ' ' + confusor_2['type']) if confusor_2 else 'None'))
+        print('confusor_receptacle_1 ' + ((confusor_receptacle_1['id'] + ' ' + confusor_receptacle_1['type']) if confusor_receptacle_1 else 'None'))
+        print('confusor_receptacle_2 ' + ((confusor_receptacle_2['id'] + ' ' + confusor_receptacle_2['type']) if confusor_receptacle_2 else 'None'))
+        print('obstructor_1 ' + ((obstructor_1['id'] + ' ' + obstructor_1['type']) if obstructor_1 else 'None'))
+        print('obstructor_2 ' + ((obstructor_2['id'] + ' ' + obstructor_2['type']) if obstructor_2 else 'None'))
```
ThomasSchellenbergNextCentury, line 286: Whoops, I accidentally left some test code here.  I'll make sure I delete it.

scene_generator/pairs.py
```diff
@@ -37,152 +30,576 @@ def move_to_location(obj_def: Dict[str, Any], obj: Dict[str, Any],
...
-                rects.pop()
+    def _initialize_each_goal(self) -> None:
+        """
+        Initialize the goals for the two scenes in this pair and all their objects:
+        - Create the objects specific to this pair (target, confusor, obstructor).
+        - Containerize this pair's objects if needed by this pair's setup or if randomly chosen.
+        - Move this pair's objects into locations specific to each scene.
+        - Create the rest of the objects (targets, distractors, walls) in the goal for the first scene.
+        - Copy the first scene's goal to make the second scene's goal and replace this pair's objects.
+        """
```
ThomasSchellenbergNextCentury, line 103: So this function is super long, and I'm not sure how I'd split it up into smaller chunks, without writing functions that have 15 inputs and 5 outputs.  Any suggestions are welcome.

scene_generator/geometry.py
```diff
@@ -48,8 +45,12 @@
...
 
 
-def random_position() -> float:
-    return round(random.uniform(MIN_PERFORMER_POSITION, MAX_PERFORMER_POSITION), POSITION_DIGITS)
+def random_position_x() -> float:
+    return round(random.uniform(ROOM_DIMENSIONS[0][0], ROOM_DIMENSIONS[0][1]), POSITION_DIGITS)
+
+
+def random_position_z() -> float:
+    return round(random.uniform(ROOM_DIMENSIONS[1][0], ROOM_DIMENSIONS[1][1]), POSITION_DIGITS)
```
chrislong, line 53: Do you need to account for the performer's width, like the old code did?

scene_generator/interaction_goals.py
```diff
@@ -201,55 +202,47 @@ def move_to_container(target: Dict[str, Any], bounding_rectangles: List[List[Dic
 class ObjectRule():
     """Defines rules for how a specific object can be made and positioned as part of a specific goal."""
 
-    def __init__(self, position_inside_receptacle=False):
-        self._position_inside_receptacle = position_inside_receptacle
+    def __init__(self, position_in_receptacle = False, position_on_receptacle = False):
+        self.position_in_receptacle = position_in_receptacle
+        self.position_on_receptacle = position_on_receptacle
```
chrislong, line 207: These seem like strange names for booleans to me. I'd expect something like `is_in_receptacle` or `in_receptacle` (and similarly for "on")

scene_generator/interaction_goals.py
```diff
@@ -266,122 +259,242 @@ def choose_definition(self, target_list: List[Dict[str, Any]]) -> Dict[str, Any]
...
+    def choose_definition(self) -> Dict[str, Any]:
+        if not self._target_definition:
+            raise exceptions.SceneException('cannot create an obstructor with no target definition')
+        for _ in range(util.MAX_TRIES):
+            if self._obstruct_vision:
+                obstructor_definition_list = geometry.get_wider_and_taller_defs(self._target_definition)
+            else:
+                # TODO MCS-262
+                pass
+            if not obstructor_definition_list:
```
chrislong, line 343: Isn't this backwards? Don't you want to stop if you have found a non-empty list? Also, why is this even in a loop? I would expect `get_wider_and_taller_defs` to return the same thing each time, but maybe it doesn't do what I think.
Also, if `_obstruction_vision` is False, this will raise an exception because `obstructor_definition_list` won't be defined.

scene_generator/interaction_goals.py
```diff
@@ -266,122 +259,242 @@ def choose_definition(self, target_list: List[Dict[str, Any]]) -> Dict[str, Any]
...
+        obstructor_definition, obstructor_angle = random.choice(obstructor_definition_list)
+        obstructor_definition = util.finalize_object_definition(obstructor_definition)
+        if not 'rotation' in obstructor_definition:
+            obstructor_definition['rotation'] = {
+                'x': 0,
+                'y': 0,
+                'z': 0
+            }
+        # Note that this rotation must be also modified with the final performer start Y.
+        obstructor_definition['rotation']['y'] = obstructor_definition['rotation']['y'] + obstructor_angle
```
chrislong, line 356: Could use `+=` here, right?

scene_generator/interaction_goals.py
```diff
@@ -266,122 +259,242 @@ def choose_definition(self, target_list: List[Dict[str, Any]]) -> Dict[str, Any]
...
-            'distractor': distractor_list
-        }, self._bounding_rects
+                distractor_rule_list.append(self.get_distractor_rule(position_in_receptacle = \
+                        (random.random() < InteractionGoal.OBJECT_RECEPTACLE_CHANCE)))
+            self.__generate_object_list(self._distractor_list, distractor_rule_list, self._target_list, \
+                    self._distractor_list)
+            self._is_distractor_list_done = True
+        return self._distractor_list
+
+    def generate_target_list(self, end_index: int = None) -> List[Dict[str, Any]]:
```
chrislong, line 398: Technically `end_index` should have type `Optional[int]`

scene_generator/interaction_goals.py
```diff
@@ -266,122 +259,242 @@ def choose_definition(self, target_list: List[Dict[str, Any]]) -> Dict[str, Any]
...
+        given distractor_list and self._bounds_list"""
+
+        for rule in rule_list:
+            object_definition = rule.choose_definition()
+            for _ in range(util.MAX_TRIES):
+                object_location, bounds_list_modified = rule.choose_location(object_definition, \
+                        self._performer_start, self._bounds_list)
+                if rule.validate_location(object_location, target_list, self._performer_start):
+                    break
+            if not object_location:
```
chrislong, line 466: Is this a sufficient test? Is it possible for `object_location` to be truthy but to not pass `rule.validate_location`?

scene_generator/pairs.py
```diff
@@ -37,152 +30,576 @@ def move_to_location(obj_def: Dict[str, Any], obj: Dict[str, Any],
...
+
+    def _generate_location_far_from_object(self, object_definition: Dict[str, Any], target_location: Dict[str, float], \
+            performer_start: Dict[str, Dict[str, float]]) -> Dict[str, float]:
+        """Generate and return a new location far away the given target location."""
+        for _ in range(util.MAX_TRIES):
+            # TODO Use existing target bounds?
+            location_far = geometry.calc_obj_pos(performer_start['position'], [], object_definition)
+            if not geometry.are_adjacent(target_location, location_far):
+                break
+        if not location_far:
```
chrislong, line 548: I don't think this is sufficient. Couldn't the last (the `MAX_TRIES`th) `location_far` that was computed be adjacent?

scene_generator/pairs_test.py
```diff
@@ -8,13 +8,31 @@
...
+BODY_TEMPLATE = {
+    'wallMaterial': 'test_wall_material',
+    'wallColors': ['test_wall_color'],
+    'performerStart': {
+        'position': {
+            'x': 0,
+            'y': 0,
+            'z': 0
+        },
+        "rotation": {
```
chrislong, line 20: Inconsistent quotes

scene_generator/geometry.py
ThomasSchellenbergNextCentury: I think we only need to account for the performer's width when we compute the performer's start position, and it's being included in the relevant function in Goal.  These functions are currently used to compute positions for other objects and dynamic walls, so I don't think the performer's width is needed here (but maybe I'm forgetting something important?).

scene_generator/interaction_goals.py
ThomasSchellenbergNextCentury: Oh, yeah, you're right, I should prefix their names with `is_` or `should_` or something similar.  Good idea.

scene_generator/interaction_goals.py
ThomasSchellenbergNextCentury: All excellent points.  You're correct on all counts.  I'll make the changes.  Thank you!

scene_generator/interaction_goals.py
ThomasSchellenbergNextCentury: Oh, I thought Python didn't support `+=`, but I must be confusing it with some other language.  Thanks for the tip.

scene_generator/interaction_goals.py
ThomasSchellenbergNextCentury: Good to know.  Are you supposed to wrap `Optional` around any function parameter that has a default argument?

scene_generator/interaction_goals.py
ThomasSchellenbergNextCentury: Ah, very good point.  I'll fix that here, and with the other blocks in which I use the same format.

scene_generator/pairs.py
ThomasSchellenbergNextCentury: Yeah, I think this is the same issue that you pointed out in your other comment.  I'll fix it and all the similar cases.

scene_generator/pairs_test.py
ThomasSchellenbergNextCentury: Whoops.  Good thing Rachel is working on that linter!  lol

scene_generator/interaction_goals.py
deanwetherby: Optional is used for value that can be None.

rartiss55:
I'm running the scene gen and going through some of the scenes generated to make sure things look okay (I made a test branch with these changes + MCS-296). I did notice this error with confuser objects when trying to generate scenes (besides the earlier mentioned issue with enclosable receptacles) -- so far I've only seen it for cribs and table_5 (although it might include more table types):

  File "/Users/rartiss/MCS/MCS-repo/scene_generator/interaction_goals.py", line 323, in choose_definition
    raise exceptions.SceneException(f'cannot find a confusor to create with target={self._target_definition}')
exceptions.SceneException: cannot find a confusor to create with target={'id': '31789918-7d6c-4ea3-86c6-6ee6c19def0d', 'type': 'crib', 'mass': 25, 'info': ['huge', 'massive', 'white', 'brown', 'crib'], 'novel_color': False, 'novel_combination': False, 'novel_shape': False, 'dimensions': {'x': 0.65, 'y': 0.9, 'z': 1.25}, 'original_location': {'position': {'x': 0.0, 'y': 0.0, 'z': 0.0}, 'rotation': {'x': 0.0, 'y': 0.0, 'z': 0.0}}, 'shows': [{'position': {'x': 0.0, 'y': 0.0, 'z': 0.0}, 'rotation': {'x': 0.0, 'y': 0.0, 'z': 0.0}, 'stepBegin': 0, 'scale': {'x': 1, 'y': 1, 'z': 1}}], 'materialCategory': ['wood'], 'materials_list': [('AI2-THOR/Materials/Wood/WhiteWood', ['white'])], 'materials': ['AI2-THOR/Materials/Wood/WhiteWood'], 'info_string': 'huge massive white brown crib', 'goal_string': 'huge massive white brown crib', 'shape': 'crib'}

ThomasSchellenbergNextCentury:
@rartiss55  Thanks for pointing out that error.  It's happening with specific objects that don't yet have possible "confusor" options (which must be the same as the target in two of the following: color, shape, size) in the `objects.py` file, so we can't currently create a confusor for either object.  I've just committed a change that will catch any scene exception and remake the whole pair.  This should solve the issue appropriately.

rartiss55:
@ThomasSchellenbergNextCentury Thanks for the earlier fix. 

If this is addressed in a more recent commit, please disregard (the last commit I had when testing scenes was 94aff2e),
but I managed to generate a scene with the agent stuck inside of the changing table (I'll post in Teams in a sec). 

I didn't know if it had to do with Chris's comment about addressing the performer's width -- if its completely separate I can make a bug in JIRA.

PR #155, opened 2020-07-13
MCS-310 MCS-262 Fixed bugs with obstructors in scene generator pairs and added Pair 9

ThomasSchellenbergNextCentury:
- Created Pair 9 with obstructors that just block agent navigation (not vision).
- If a target is containerized inside a receptacle, we must use the receptacle's dimensions in size comparisons to find an obstructor that blocks vision.
- In object definitions file, differentiated between possible obstructors that block vision (and navigation) versus navigation only.
- Fixed bugs in pair unit tests.


deanwetherby:
For 2B, if the object is in a box, does that count as occluded? Essentially 2A and 2B are the same.

deanwetherby:
Beautiful. hidden racecar goal object!
image

deanwetherby:
I really wish containers were a special case. Having containers associated with every pair makes testing more difficult.

deanwetherby:
2B goal object not blocked initially by newly added box.

image

deanwetherby:
For 9A/B, this couch/chair is blocking the view for the pair.
image

ThomasSchellenbergNextCentury:
@deanwetherby 
- Can you please send me the example scene with the box and brown ball (and the blue walls)?
- Currently there's a possibility that the target can be inside a box, and we aren't counting that as occluded for 2A.  The psych team signed off on that before, but they've signed off on many things that later they decide they don't like, so I don't know if it's better to keep that possibility or get rid of it.
- Currently random distractors and walls can accidentally occlude the target.  There's an open ticket for that.  I'll see if I can fix it tonight after dinner.  https://nextcentury.atlassian.net/browse/MCS-298

deanwetherby:
[aardvark-0006-1.txt](https://github.com/NextCenturyCorporation/MCS/files/4915262/aardvark-0006-1.txt)
[aardvark-0006-2.txt](https://github.com/NextCenturyCorporation/MCS/files/4915263/aardvark-0006-2.txt)

ThomasSchellenbergNextCentury:
To reiterate what I said at standup earlier, all of the issues should be fixed now.

rartiss55:
For 2B, I have a scene where it looks like the target object for traversal is in front of an object that looks exactly the same (from the json, it looks like they're both ["huge","massive","brown","sofa"] of type "sofa_1":
Screen Shot 2020-07-14 at 2 24 04 PM

ThomasSchellenbergNextCentury:
@rartiss55  Oh, yes, haha, that's a bug.  Nice catch.  I think the issue here is that we're currently validating obstructors as long as they're positioned on the line between the performer and the target's CENTER, when we should probably be checking the 4 lines between the performer and the target's bounding box corners.  That may be an easy fix.

ThomasSchellenbergNextCentury:
@rartiss55  Should be fixed now.

rartiss55:
Things looked good until my very last scene :(

Screen Shot 2020-07-16 at 5 13 56 PM

I also noticed the changing table placement can be a bit further away from the target object than I expected in 2B scenarios, but it looks like you fixed something with the bounding box recently in mcs-private. Wasn't sure if there was a related fix involved that needs to make its way into the python code.

ThomasSchellenbergNextCentury:
@rartiss55  Thanks for all your extensive testing.  This is enough of a head-scratcher that I think we should move it to a separate ticket, if that'd be OK with you.  I've created this ticket, if you wouldn't mind attaching the problematic scene to it.  https://nextcentury.atlassian.net/browse/MCS-314

For the changing tables (and also the wardrobes), the scene generator should leave some extra space in front of them so their drawers can be opened (it's intentional).

rartiss55:
@ThomasSchellenbergNextCentury Sounds fine to me, and I attached the scene file to that ticket.

Did a bit of testing with pair 9. I did notice it took a bit longer to make scenes (more errors but I guess its catching issues ahead of time so thats good!) I have this one 9B scene where a chair leg is sticking through a box -- its a minor issue since you can still accomplish the goal, but thought I'd mention it here anyway and post the file. This might already be covered in a different ticket, not sure.

[pair9-0002-2.txt](https://github.com/NextCenturyCorporation/MCS/files/4935132/pair9-0002-2.txt)
Screen Shot 2020-07-16 at 8 56 38 PM

ThomasSchellenbergNextCentury:
@rartiss55  Ugh, I've had so much trouble with boxes.  I made a new ticket for the issue:  https://nextcentury.atlassian.net/browse/MCS-319

PR #160, opened 2020-07-21
MCS-186: UI Design for Eval2

brianpippin:
This is a work in progress.  Scenes.jsx needs refactoring, but waiting to do more refactoring when we add in Interactive display hooks.

### DIFF
--- a/mcs_docker_setup/analysis-ui/src/components/App/addComment.jsx
+++ b/mcs_docker_setup/analysis-ui/src/components/App/addComment.jsx
@@ -7,10 +7,17 @@ let commentState = {};
 let setNeedToRefetch = null;
 
 const AddComment = ({ mutate }) => {
-    const submitComment = (evt) => {
-        evt.persist();
-        mutate({
-            variables: { 
+     const getMutationVariables = function() {
+        if(commentState.test_type !== undefined && commentState.test_type !== null) {
+            return {
+                testType: commentState.test_type,
+                sceneNum: commentState.scene_num,
+                createdDate: (new Date()).toISOString(),
+                text: document.getElementById("commentTextArea").value,
+                userName: document.getElementById("commentUserName").value
+            };
+        } else {
+            return {
                 test: commentState.test,
                 block: commentState.block,
                 submission: commentState.subm,
@@ -18,7 +25,14 @@ const AddComment = ({ mutate }) => {
                 createdDate: (new Date()).toISOString(),
                 text: document.getElementById("commentTextArea").value,
                 userName: document.getElementById("commentUserName").value
-            }
+            };
+        }
+    }
+
+    const submitComment = (evt) => {
+        evt.persist();
+        mutate({
+            variables: getMutationVariables()
         }).then( res => {
             document.getElementById("commentTextArea").value = '';
             document.getElementById("commentUserName").value = '';
@@ -50,7 +64,28 @@ const addCommentMutation = gql`
     }
 `;
 
+const addCommentTestTypeMutation = gql`
+    mutation saveCommentByTestAndScene($testType: String!, $sceneNum: String!, $createdDate: String!, $text: String!, $userName: String!) {
+        saveCommentByTestAndScene(testType: $testType, sceneNum: $sceneNum, createdDate: $createdDate, text: $text, userName: $userName) {
+            text
+        }
+    }
+`;
+
 const AddCommentWithMutation = graphql(addCommentMutation)(AddComment);
+const AddCommentTestTypeWithMutation = graphql(addCommentTestTypeMutation)(AddComment);
+
+const AddCommentHolder = () => {
+    if(commentState.test_type !== undefined && commentState.test_type !== null) {
+        return (
+            <AddCommentTestTypeWithMutation/>
+        )
+    } else {
+        return (
+            <AddCommentWithMutation/>
+        )
+    }
+}
 
 class AddCommentBlock extends React.Component {
     render() {
@@ -58,7 +93,7 @@ class AddCommentBlock extends React.Component {
         setNeedToRefetch = this.props.setNeedToRefetch;
 
         return(
-            <AddCommentWithMutation/>
+            <AddCommentHolder/>
         )
     }
 }
--- a/mcs_docker_setup/analysis-ui/src/components/App/commentsDisplay.jsx
+++ b/mcs_docker_setup/analysis-ui/src/components/App/commentsDisplay.jsx
@@ -3,8 +3,8 @@ import _ from "lodash";
 import { Query } from 'react-apollo';
 import gql from 'graphql-tag';
 
-const queryName = "getComments"
-//TODO:  Update graphql so that we only return the values we need on the final page
+// For Eval 1 Comments
+const queryName = "getComments";
 const GET_COMMENTS = gql`
     query getComments($test: String!, $block: String!, $submission: String!, $performer: String! ){
         getComments(test: $test, block: $block, submission: $submission, performer: $performer) {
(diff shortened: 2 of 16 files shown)


brianpippin:
This is ready for review.

mcs_docker_setup/analysis-ui/src/components/App/addComment.jsx
```diff
@@ -7,18 +7,32 @@ let commentState = {};
...
-    const submitComment = (evt) => {
-        evt.persist();
-        mutate({
-            variables: { 
+     const getMutationVariables = function() {
+        if(commentState.test_type !== undefined && commentState.test_type !== null) {
+            return {
+                testType: commentState.test_type,
+                sceneNum: commentState.scene_num,
+                createdDate: (new Date()).toISOString(),
```
ThomasSchellenbergNextCentury, line 15: Should we be concerned about time zones here?

mcs_docker_setup/analysis-ui/src/components/App/commentsDisplay.jsx
```diff
@@ -19,48 +19,86 @@ const GET_COMMENTS = gql`
...
+            {commentsInOrder.length === 0 && 
+                <div>There are no comments yet for this test.</div>
+            }
+        </div>
+    );
+}
+
+const CommentBlock = ({props, qName, queryType, queryVars}) => {
+    return (
+        <Query query={queryType} variables={queryVars} fetchPolicy={props.needToRefetch ? 'network-only' : 'cache-first'}>
```
ThomasSchellenbergNextCentury, line 56: Will this fetchPolicy requery if you refresh the webpage?  I want to make sure this works if another user submits a new comment.

mcs_docker_setup/analysis-ui/src/components/App/header.jsx
```diff
@@ -34,7 +41,13 @@ class ListItem extends React.Component {
         }
         
         const urlBasePath = window.location.href.split('?')[0];
-        const params = "?perf=" + this.props.state["perf"] + "&subm=" + this.props.state["subm"] + "&block=" + this.props.state["block"] + "&test=" + this.props.state["test"];
+        let params = ""
+        if(this.props.state["perf"] !== undefined && this.props.state["perf"] !== null) {
+            params = "?perf=" + this.props.state["perf"] + "&subm=" + this.props.state["subm"] + "&block=" + this.props.state["block"] + "&test=" + this.props.state["test"];
+        }
+        if(this.props.state["test_type"] !== undefined && this.props.state["test_type"] !== null) {
+            params = "?test_type=" + this.props.state["test_type"] + "&scene_num=" + this.props.state["scene_num"];
```
ThomasSchellenbergNextCentury, line 49: Should this be an `else if`?  Or maybe a `params +=`?

mcs_docker_setup/analysis-ui/src/components/App/scenes.jsx
```diff
@@ -0,0 +1,399 @@
...
+
+let currentState = {};
+let currentStep = 0;
+
+const PERFORMER_PREFIX_MAPPING = {
+    "IBM-MIT-Harvard-Stanford": "mitibm_",
+    "OPICS (OSU, UU, NYU)": "opics_",
+    "MESS-UCBerkeley": "mess_",
+    "IBM-MIT-Harvard-Stanford-2": "mitibm2_"
+};
```
ThomasSchellenbergNextCentury, line 25: Should we move this performer info to a separate file, so it's easier to maintain over future evals?

mcs_docker_setup/analysis-ui/src/index.js
```diff
@@ -7,7 +7,7 @@ import { ApolloClient } from 'apollo-client'
 import { createHttpLink } from 'apollo-link-http'
 import { InMemoryCache } from 'apollo-cache-inmemory'
 
-console.log("uri", window.GRAPHQL_URL, "1002");
+//console.log("uri", window.GRAPHQL_URL, "1002");
```
ThomasSchellenbergNextCentury, line 10: Should this still be commented-out?

mcs_docker_setup/node-graphql/server.graphql.js
```diff
@@ -232,6 +332,31 @@ const resolvers = {
...
+        },
+        "query": {
+          "bool": {
+            "must": [
+              {"match": {"test_type": args["testType"]}},
+              {"match": {"scene_num":  args["sceneNum"]}}
+            ]
+          }
+        }
+      });
```
ThomasSchellenbergNextCentury, line 359: Not a big deal, but it seems like this should be two separate queries (one for each flag).

mcs_docker_setup/analysis-ui/src/components/App/addComment.jsx
brianpippin: We should potentially consider doing something with timezones, I will try to look into that when doing code reviews.

mcs_docker_setup/analysis-ui/src/components/App/commentsDisplay.jsx
brianpippin: Yes if you refresh, it will pull in other user comments, we don't have a mechanism at the moment that will auto pull the comments every minute or something, we just don't see that many comments usually, but might be something we want to consider if more people start using this feature.

mcs_docker_setup/analysis-ui/src/components/App/header.jsx
brianpippin: This code is actually on my list to refactor, I just don't like and want the eval 1 code and hopefully eval2 code going forward more separate and different components.

mcs_docker_setup/analysis-ui/src/components/App/scenes.jsx
brianpippin: Yes this was the only thing I had to hardcode and agree, it should be in a new file, along with some of the constants for the URLs because we might need to change that code depending on what eval it is, and all of that in a file with specific objects for each eval will make it very easy to update where the images/files are etc.

mcs_docker_setup/analysis-ui/src/index.js
brianpippin: It should either be not commented out or it should be removed.  I'd lean to removing it, because it makes the graphql endpoint easier to find.

mcs_docker_setup/node-graphql/server.graphql.js
brianpippin: Yes you are correct, that would make more sense from a design standpoint, it should be pretty easy to add that.

mcs_docker_setup/analysis-ui/src/components/App/scenes.jsx
```diff
@@ -0,0 +1,399 @@
...
+    }
+
+    findObjectTabName = (sceneObject) => {
+        if(sceneObject.shape !== undefined && sceneObject.shape !== null) {
+            return sceneObject.shape;
+        }
+
+        if(sceneObject.id.indexOf('occluder_wall')) {
+            return "occluder wall";
+        }
```
ThomasSchellenbergNextCentury, line 167: The code doesn't appear to be doing this any more, so maybe something's changed, but if you look at the deployed version:

http://ec2-54-157-23-98.compute-1.amazonaws.com:3000/?test_type=pair_1_retrieval&scene_num=0001

You'll see that the interior walls (with ids of `wall_*`) are called "occluder wall".  We may want to label them "interior wall" for clarity.

mcs_docker_setup/analysis-ui/src/components/App/commentsDisplay.jsx
ThomasSchellenbergNextCentury: That's fine.  I just saw "cache-first" so I thought I'd ask :)

mcs_docker_setup/analysis-ui/src/components/App/header.jsx
ThomasSchellenbergNextCentury: OK, it wasn't clear if those two `if` statements were needed to handle differences between eval 1 and 2, or if both of them could happen on the same scene.

mcs_docker_setup/analysis-ui/src/components/App/commentsDisplay.jsx
brianpippin: Yeah that is just the policy when the page is loaded, but the query always fetches the latest when the page loads.

mcs_docker_setup/analysis-ui/src/components/App/scenes.jsx
brianpippin: Ideally going forward if we know we want to change these labels in the UI we should consider updating the scene generator so the names we want to use are in the files and the UI doesn't need to do these calculations.  I think we probably are discussing the same thing and want to look into it.

mcs_docker_setup/resources/config-staging.yaml
```diff
@@ -442,8 +442,8 @@ layouts:
       idField: "_id_key"
       limit: 100
       ignoreSelf: false
-      #linkPrefix: "http://ec2-54-157-23-98.compute-1.amazonaws.com:3000/?"
-      linkPrefix: "http://localhost:2000/?"
+      linkPrefix: "http://ec2-54-157-23-98.compute-1.amazonaws.com:3000/?"
+      #linkPrefix: "http://localhost:2000/?"
```
ThomasSchellenbergNextCentury, line 446: It seems like there should be a "template" file for the Neon config, since the staging and non-staging versions have a lot in common (the only difference currently is the comments widget).  I'm not sure the best way to handle that, though.

mcs_docker_setup/analysis-ui/src/components/App/header.jsx
brianpippin: Yeah it's an eval1/eval2 if statement, I would like to make them more componentized so that it's obvious in the code base what is going on.

mcs_docker_setup/resources/config-staging.yaml
brianpippin: I think there should potentially be a better way to do some of that.  I think I am going to create a new ticket for trying to improve the Neon configs, template wise and what not to help us moving forward and will leave improvements to that ticket.

ThomasSchellenbergNextCentury: Sounds good.  In one previous project, I had a "template" file and a script that copied the file to my development and production configs, so that's one option, but there may be better ones as well.

mcs_docker_setup/analysis-ui/src/components/App/scenes.jsx
ThomasSchellenbergNextCentury: Good idea.  Are the occluders and the walls the problematic ones, currently?

mcs_docker_setup/analysis-ui/src/components/App/addComment.jsx
brianpippin: So I checked and did some research and we are recording the UTC string, which is Zulu time for when the comment is submitted so it should be the best way to take into account which timezone it is submitted in.  The comments are then sorted by dates, so I think we are good on time zones.

mcs_docker_setup/resources/config-staging.yaml
brianpippin: Created MCS-359 to address the template as well as figuring out how to do something so we don't need to modify the files for running locally.

PR #161, opened 2020-07-24
MCS-295

jaudick:
https://nextcentury.atlassian.net/browse/MCS-295

This works for any json scene though

### DIFF
--- a/scene_generator/intphys_non_matching_ids.py
+++ b/scene_generator/intphys_non_matching_ids.py
@@ -14,7 +14,7 @@ def non_matching_ids():
     assert os.path.exists(user_input), "No file at, "+str(user_input)
     file_names = os.listdir(user_input)
 
-    for file in os.listdir(user_input):
+    for file in file_names:
         found_match = False
         file_to_open = os.path.join(user_input,file)
         with open(file_to_open, 'r') as file_data:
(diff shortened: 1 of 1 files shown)


scene_generator/intphys_non_matching_ids.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def non_matching_ids():
+    found_match = False
+    files = []
+    user_input = Path(input('Enter Dataset directory: '))
+    assert os.path.exists(user_input), "No file at, "+str(user_input)
+    file_names = os.listdir(user_input)
+
+    for file in os.listdir(user_input):
```
arashBehpour, line 17: This is a nitpick but might want to use your file_names variable here.

jaudick: @Yeah that would make sense, dont even think the files_names var is currently doing anything right now. I'll change that

scene_generator/intphys_non_matching_ids.py
```diff
@@ -0,0 +1,44 @@
...
+                    if num_of_times_object_id_appears == 0:
+                        num_of_times_object_id_appears += 1
+                        continue
+                    else:
+                        print("Matching object ID '" + object_b['id'] + "' in " + file)
+                        object_b['id'] = str(uuid.uuid4())
+                        found_match = True
+
+        if found_match:
+            new_scene_file = os.path.join(user_input, 'legacy.' + file)
```
ThomasSchellenbergNextCentury, line 37: Looks good, though `legacy` here implies to me that this file is the OLD file and not the NEW file, so I'd recommend changing the prefix.

PR #162, opened 2020-07-27
Mcs 147

arashBehpour:
Generate painting mostly done except for TODO on line 116 in goal.py and the issue where it still generates edge case below
paintingIssue


scene_generator/goal.py
```diff
@@ -82,12 +90,162 @@ def generate_wall(wall_material: str, wall_colors: List[str], performer_position
     return None
 
 
+def generate_painting(painting_material: str, wall_colors: List[str], performer_position: Dict[str, float],
+                  other_rects: List[List[Dict[str, float]]], generated_wall_rects: List[List[Dict[str, float]]]) -> Optional[Dict[str, Any]]:
+    # Chance to generate a painting in the room, that can only be hanged on a wall
+
+    # 1) Randomly grab x,y,z coordinate like generate_walls 
+    # 2) Make sure painting(to be generated) is close to any wall
```
ThomasSchellenbergNextCentury, line 98: First of all, these are great comments - Thanks!  Second, would it make more sense to restrict the randomly generated X/Z coordinates to only those of known walls?  Instead of using `geometry.random_position_*`, you could make your own functions that choose a random wall from the `generated_wall_rects`, and then choose a random X/Z along the rect.

scene_generator/goal.py
```diff
@@ -82,12 +90,162 @@ def generate_wall(wall_material: str, wall_colors: List[str], performer_position
...
+    # Second goal- make sure I can generate a painting on generated_walls()
+
+    tries = 0
+    performer_rect = geometry.find_performer_rect(performer_position)
+    performer_poly = geometry.rect_to_poly(performer_rect)
+    while tries < util.MAX_TRIES:
+        rotation = random.choice((0, 90, 180, 270))
+        new_x = geometry.random_position_x()
+        new_z = geometry.random_position_z()
+        new_y = geometry.random_position_x()
```
ThomasSchellenbergNextCentury, line 112: The `new_y` is calculated from `random_position_x` - I wonder if that is causing any issues?

ThomasSchellenbergNextCentury, line 112: The Y coordinate is being calculated with the function for X coordinates.  I wonder if that is causing any issues?  The Y coordinate should probably depend on the size of the painting.

scene_generator/geometry.py
```diff
@@ -403,11 +403,14 @@ def get_bounding_polygon(object_or_location: Dict[str, Any]) -> shapely.geometry
...
-def are_adjacent(obj_a: Dict[str, Any], obj_b: Dict[str, Any]) -> bool:
+def are_adjacent(obj_a: Dict[str, Any], obj_b: Dict[str, Any], max_adj_dist: float = None) -> bool:
     poly_a = get_bounding_polygon(obj_a)
     poly_b = get_bounding_polygon(obj_b)
     distance = poly_a.distance(poly_b)
-    return distance <= MAX_ADJACENT_DISTANCE
+    if max_adj_dist:
+        return distance <= max_adj_dist
+    else:
+        return distance <= MAX_ADJACENT_DISTANCE
```
ThomasSchellenbergNextCentury, line 413: Little tidbit: If you want to get fancy, you can do this in just one line!

```python
return distance <= max_adj_dist if max_adj_dist else distance <= MAX_ADJACENT_DISTANCE
```

OR, getting even fancier:

```python
return distance <= (max_adj_dist if max_adj_dist else MAX_ADJACENT_DISTANCE)
```

PS:  `if max_adj_dist` will be `False` if `max_adj_dist` is `0`.  I don't think it really matters here, since `0` doesn't seem very likely.  Still, if you want to check specifically for `None`, and not just any falsey value, you say `if max_adj_dist is not None`

scene_generator/geometry.py
```diff
@@ -80,7 +80,7 @@ def collision(test_rect: List[Dict[str, float]], test_point: Dict[str, float]):
                 0 <= dot_prod_dict(vectorBC, vectorBM) <= dot_prod_dict(vectorBC, vectorBC))
 
 
-def calc_obj_coords(position_x: float, position_z: float, delta_x: float, delta_z: float, offset_x: float,
+def calc_obj_coords(position_x: float, position_z: float, position_y: float, delta_x: float, delta_z: float, offset_x: float,
                     offset_z: float, rotation: float) -> List[Dict[str, float]]:
```
ThomasSchellenbergNextCentury, line 84: Adding the new `position_y` argument here will affect invocations to `calc_obj_coords` in other parts of the code.  I would recommend either:
1. Removing the `position_y` argument and manually updating the `y` coordinate of the output in the Goals' `generate_painting` function.
2. Move `position_y` back so it's the last argument and add a default value, like this:

```python
def calc_obj_coords(position_x: float, position_z: float, delta_x: float, delta_z: float, offset_x: float,
                    offset_z: float, rotation: float, position_y: float = 0) -> List[Dict[str, float]]:
```

scene_generator/goal.py
```diff
@@ -99,6 +257,22 @@ def update_body(self, body: Dict[str, Any], find_path: bool) -> Dict[str, Any]:
         self._tag_to_objects = self.compute_objects(body['wallMaterial'], body['wallColors'])
 
         body['performerStart'] = self._performer_start
+        walls = self.generate_walls(body['wallMaterial'], body['wallColors'], body['performerStart']['position'], [])
+        #for wall in walls: # DO NOT NEED B/C WALLS ALREADY NOT
+            #bounding_rects.append(wall['shows'][0]['bounding_box'])
```
ThomasSchellenbergNextCentury, line 262: You can now get the walls from `self._tag_to_objects`, like this:

```python
walls = self._tag_to_objects['wall'] if 'wall' in self._tag_to_objects else []
```

So you can delete this line and the `generate_walls` function in this file.

scene_generator/goal.py
```diff
@@ -99,6 +257,22 @@ def update_body(self, body: Dict[str, Any], find_path: bool) -> Dict[str, Any]:
...
+        print("hit")
+        print("\n\n\n")
+
+        walls.extend(paintings) #FIXME: need to have seperate tag for paintings. Will remove later.
+        
+        if walls is not None:
+            self._tag_to_objects['wall'] = walls
+
+        if paintings:
+            self._tag_to_objects['paintings'] = paintings
```
ThomasSchellenbergNextCentury, line 274: I think you can remove lines 268-271 since lines 273-274 should save the paintings for you.

scene_generator/goal.py
```diff
@@ -99,6 +257,22 @@ def update_body(self, body: Dict[str, Any], find_path: bool) -> Dict[str, Any]:
         self._tag_to_objects = self.compute_objects(body['wallMaterial'], body['wallColors'])
 
         body['performerStart'] = self._performer_start
+        walls = self.generate_walls(body['wallMaterial'], body['wallColors'], body['performerStart']['position'], [])
+        #for wall in walls: # DO NOT NEED B/C WALLS ALREADY NOT
+            #bounding_rects.append(wall['shows'][0]['bounding_box'])
+
+        paintings = self.generate_paintings(body['paintingMaterial'], body['paintingColors'], body['performerStart']['position'], [], walls)
```
ThomasSchellenbergNextCentury, line 264: First, we should add the room's four original walls to the `generate_painting` function's `generated_wall_rects` argument.

Second, the `generate_painting` function's `other_rects` argument should be `self.get_bounds_list()`.  That function is currently defined in `interaction_goals.py`, but you can move it here (into the Goal class), along with the initialization of `self._bounds_list` in the constructor.

Something like:

```python
all_walls = walls + [
    # These calls do not use the function's position_y argument (see my comment above)
    geometry.calc_obj_coords(0, 5.5, 1, 1, 0, 0, 0),
    geometry.calc_obj_coords(0, -5.5, 1, 1, 0, 0, 0),
    geometry.calc_obj_coords(5.5, 0, 1, 1, 0, 0, 90),
    geometry.calc_obj_coords(-5.5, 0, 1, 1, 0, 0, 90)
]

paintings = self.generate_paintings(body['paintingMaterial'], body['paintingColors'], body['performerStart']['position'], self.get_bounds_list(), all_walls)
```

scene_generator/goal.py
```diff
@@ -82,12 +90,162 @@ def generate_wall(wall_material: str, wall_colors: List[str], performer_position
...
+
+                        x_list = [coord['x'] for coord in wall['shows'][0]['bounding_box']]
+                        x_list.sort()
+                        left_x_boundary = x_list[0]
+                        right_x_boundary = x_list[3] 
+
+                        new_z = wall['shows'][0]['position']['z'] + random.choice([WALL_DEPTH - 0.03, -WALL_DEPTH + 0.03]) # FIXME Test to see paintings on generated wall
+                        
+                        shows_object = {'bounding_box': geometry.calc_obj_coords(new_x, new_z, new_y, new_x_size, PAINTING_DEPTH, 0, 0, rotation)}
+                        cur_painting_obj['shows'] = [shows_object]
```
ThomasSchellenbergNextCentury, line 142: You could calculate the `bounding_box` once, when you initialize `cur_painting_obj`, so you don't have to do it multiple times (here, lines 177-178, and line 204).

scene_generator/goal.py
```diff
@@ -82,12 +90,162 @@ def generate_wall(wall_material: str, wall_colors: List[str], performer_position
...
+            cur_painting_obj = dict()
+            if (rotation == 0 or rotation == 180): #Make sure the painting is on the wall, instead of hovering in front of the wall
+                
+                for wall in generated_wall_rects:
+                    if wall['shows'][0]['rotation']['y'] == 0 or wall['shows'][0]['rotation']['y'] == 180:
+
+                        x_list = [coord['x'] for coord in wall['shows'][0]['bounding_box']]
+                        x_list.sort()
+                        left_x_boundary = x_list[0]
+                        right_x_boundary = x_list[3]
```
ThomasSchellenbergNextCentury, line 137: FYI `bounding_box[0]` should be top-right and `bounding_box[2]` should be bottom-left, if you don't want to have to create a new variable and call sort on it.  (Same thing with lines 144-147, 170-173, and 180-183.)

scene_generator/goal.py
```diff
@@ -82,12 +90,162 @@ def generate_wall(wall_material: str, wall_colors: List[str], performer_position
...
+                        shows_object = {'bounding_box': geometry.calc_obj_coords(new_x, new_z, new_y, new_x_size, PAINTING_DEPTH, 0, 0, rotation)}
+                        cur_painting_obj['shows'] = [shows_object]
+                        
+                        painting_x_list = [coord['x'] for coord in cur_painting_obj['shows'][0]['bounding_box']] #boundaries for current painting
+                        painting_x_list.sort()
+                        painting_left_x_boundary = painting_x_list[0]
+                        painting_right_x_boundary = painting_x_list[3]
+
+                        if geometry.are_adjacent(wall, cur_painting_obj, 0.1) and \
+                            left_x_boundary <= painting_left_x_boundary and right_x_boundary >= painting_right_x_boundary:
```
ThomasSchellenbergNextCentury, line 150: So I'm guessing this `if` statement affects your painting-sticking-half-off-the-wall problem but I don't see an obvious issue with it.  Maybe my above suggestion of refactoring this function (line 98) would end up solving the problem?

scene_generator/goal.py
arashBehpour: Thank you! Yeah so I do something similar in lines 139 & 175 basically increasing the chances of putting a painting on a generated wall. The only thing is instead of randomly choosing an X/Z near a generated wall, I just do it for every generated wall. I can implement that type of function, but after I fix the painting partially off the wall issue. Because right now the way it is set up, I often generate scenes with paintings on generated on walls. I will look in to it!

scene_generator/geometry.py
arashBehpour: I added these changes!

scene_generator/geometry.py
arashBehpour: Added these changes, thank you!

scene_generator/goal.py
arashBehpour: Fixed it, thank you!

scene_generator/goal.py
arashBehpour: I just copied and made a random_position_y() so that the functions goal focuses on grabbing a valid Y value within the room(same as x&z). However since a valid Y value depends on the Y-size of the painting, I have an if condition that makes sure the painting is above the floor.

scene_generator/goal.py
arashBehpour: I changed up the code a bit, however the reason I calculate a new bounding box here and lines 177-178 is because the new_z or new_x changes based on the specific generated wall. Why I do this is b/c I want to increase my chances of placing a painting on a generated wall, rather than the rooms walls.

scene_generator/goal.py
arashBehpour: Yeah I would have just directly grabbed these values. However the values are not consistent when I grab the points from the bounding box. When I want the left most boundary it will not always be bounding_box[0]

scene_generator/goal.py
arashBehpour: Yeah I do not see an issue with it as well. I am thinking about re-coding the generate_painting() function

scene_generator/goal.py
ThomasSchellenbergNextCentury: Well, I meant that there's a typo here!

scene_generator/goal.py
ThomasSchellenbergNextCentury: OK.  Well, better safe than sorry!

scene_generator/goal.py
arashBehpour: Added!

scene_generator/goal.py
arashBehpour: This does not generate walls. How does the scene generate walls to be placed in self._tag_to_objects['wall']?

arashBehpour:
Finished resolving merge conflicts. Everything works. Theoretically lines 425 & 442 on goal.py should have the random.choices() un-commented so scenes that are generated do not have a fixed # of walls nor fixed # of paintings. Left it there if anyone wanted to go through a couple of scenes with paintings being generated on interior walls.

scene_generator/goal.py
```diff
@@ -84,6 +91,150 @@ def generate_wall(wall_material: str, wall_colors: List[str], performer_position
...
+                
+                elif (wall['shows'][0]['rotation']['y'] == 90 or wall['shows'][0]['rotation']['y'] == 270) and \
+                     (rotation == 90 or rotation == 270):
+
+                    z_list = [coord['z'] for coord in wall['shows'][0]['bounding_box']]
+                    z_list.sort()
+                    left_z_boundary = z_list[0]
+                    right_z_boundary = z_list[3]
+
+                    new_x = wall['shows'][0]['position']['x'] + random.choice([WALL_DEPTH - 0.03, -WALL_DEPTH + 0.03]) # Test to see paintings on generated wall
```
rartiss55, line 170: TODO: Consider making this 0.03 value a constant.

scene_generator/goal.py
```diff
@@ -84,6 +91,150 @@ def generate_wall(wall_material: str, wall_colors: List[str], performer_position
...
+                    (len(other_rects) == 0 or not any(separating_axis_theorem.sat_entry(rect, other_rect) for other_rect in other_rects)): 
+                break
+
+        tries += 1
+    
+    if tries < util.MAX_TRIES:
+        new_object = {
+            'id': 'painting_' + str(uuid.uuid4()), 
+            'materials': [painting_material],
+            'type': 'cube',
```
rartiss55, line 216: TODO: This should use a "painting_*" type

scene_generator/goal.py
```diff
@@ -97,18 +248,68 @@ class Goal(ABC):
...
+                'stepBegin': 0,
+                'scale': {'x': x_deltas[i], 'y': WALL_HEIGHT, 'z': z_deltas},
+                'rotation': {'x': 0, 'y': rotations[i], 'z': 0},
+                'position': {'x': x_positions[i], 'y': WALL_Y_POS, 'z': z_positions[i]},
+                'bounding_box': geometry.calc_obj_coords(x_positions[i], z_positions[i], x_deltas[i], z_deltas[i], x_offsets[i], z_offsets[i], rotations[i])
+            }
+            new_object['shows'] = [shows_object]
+            rooms_walls.append(new_object)
+        
+        all_walls = walls + rooms_walls #--> Messes up with generating a painting on a room wall. Could have a painting inside one of the rooms 4 walls.
```
rartiss55, line 301: Need to look into this + remove commented out print statements below.

rartiss55:
Leaving a note for things to fix/address...this needs:
-  uncommenting/fixing lines 425 & 442 like stated above
- make it generate randomized paintings (looks like its generating just blue cubes right now instead of using objects of type "painting_*" and its also using RUBBER_MATERIALS - not sure if this is correct)
- on that note, were there any outstanding issues with that besides anything mentioned in this PR? (@ThomasSchellenbergNextCentury)
- lots of broken unit tests (41 or so)

scene_generator/geometry.py
```diff
@@ -16,8 +16,10 @@
 VALID_ROTATIONS = (0, 45, 90, 135, 180, 225, 270, 315)
 
 ROOM_X_MIN = -4.95
+ROOM_Y_MIN = -4.95
 ROOM_Z_MIN = -4.95
 ROOM_X_MAX = 4.95
+ROOM_Y_MAX = 4.95
```
ThomasSchellenbergNextCentury, line 22: The Y range goes from 0 to 3.

scene_generator/goal.py
```diff
@@ -156,6 +399,72 @@ def update_body(self, body: Dict[str, Any],
                 object_instance['role'] = tag
 
         body['performerStart'] = self._performer_start
+        walls = self.generate_walls(
+            body['wallMaterial'],
+            body['wallColors'],
+            body['performerStart']['position'],
+            [])
```
ThomasSchellenbergNextCentury, line 406: Interior walls are already made by the InteractionGoal, so we don't need to do it again here.

rartiss55: Ah okay, thanks for clarifying -- I didn't really understand what this was doing, but when I took it out I noticed goofy behavior and didn't know if it was related.

scene_generator/goal.py
```diff
@@ -156,6 +399,72 @@ def update_body(self, body: Dict[str, Any],
...
+        # --> Messes up with generating a painting on a room wall. Could have
+        # a painting inside one of the rooms 4 walls.
+        all_walls = walls + rooms_walls  # noqa: F841
+
+        paintings = self.generate_paintings(
+            body['paintingMaterial'],
+            body['paintingColors'],
+            body['performerStart']['position'],
+            self.get_bounds_list(),
+            walls)
```
ThomasSchellenbergNextCentury, line 458: We'll want to move the generation of paintings into the InteractionGoal.

scene_generator/geometry.py
rartiss55: Oops, duh. Thanks for catching that.

rartiss55:
Closing -- this ticket will need some non-trivial additions.

PR #164, opened 2020-07-27
MCS-97 Sphinx Documentation 

jaudick:
Having troubles with Sphinx. The problem is it seems Sphinx generates documentation based on the repo and not the local changes. If I write new comments they do not update when generating the documentation.

-I changed a comment block in `mcs.py` to test the problem.

-Made a `SPHINX_README.md` for setup. 

-You can delete the `docs` folder for a clean setup.


ThomasSchellenbergNextCentury:
I was having some trouble getting all the dependencies installed.  Could another reviewer give it a try sometime?

deanwetherby:
I'll take a look.

deanwetherby:
I got it to work but not without issues. (BTW, our package dependency management is not good. I'll look at cleaning that up.)

Steps

1) Made a new virtual environment from scratch
```
$ python3.6 -m venv --prompt mcs2 venv2
$ . venv2/bin/activate
(mcs2) $ python -m pip install --upgrade pip setuptools wheel
(mcs2) $ python -m pip install -e .
(mcs2) $ python -m pip install -r requirements.txt
(mcs2) $ python -m pip install sphinx-markdown-builder
# Why isn't sphinx-markdown-builder just part of requirements.txt?
```

2) Had some issues installing MarkupSafe from requirements.txt. I changed the version to 1.1.1 to get around the error.

3) Had issues installing mysqlclient. Error complained about missing sql config. Fix was to `sudo apt install libmysqlclient-dev`. (What is this dependency used for?)

4) Followed README directions to make html and markdown. Made changes to comment section in mcs.py and rebuilt docs. Change reflected in new html and in the markdown.

ThomasSchellenbergNextCentury:
I've made some changes and I think this PR is ready for review.
- Removed all the build files from the repo.
- Removed the part of the README about generating the sphinx config files (they're already saved in the repo).
- Removed all the versions from the pip requirements (except ai2thor), since sometimes they weren't the most recent version, and I don't think specific versions were needed.
- Updated the class and function documentation to match the previous API documentation.
- Updated the sphinx "index" config file to look more like the old API documentation.
- Created a pre-commit script to automatically regenerate the sphinx documentation, copy it out of the build folder, and commit it.
- Moved the documentation about MCS configuration from the old API documentation into the README.
- Please note that your sphinx virtual environment must now be setup at `python_api/machine_common_sense/sphinx_venv`

ThomasSchellenbergNextCentury:
Here's the actual documentation: https://github.com/NextCenturyCorporation/MCS/blob/MCS-97/python_api/API.md

I don't think it's as good as the old documentation, but it's a start.

python_api/machine_common_sense/requirements.txt
```diff
@@ -1,33 +1,34 @@
...
+rinoh-typeface-texgyrepagella
+rinohtype
+six
+snowballstemmer
+Sphinx
+sphinxcontrib-websupport
+sphinx-markdown-builder
+SQLAlchemy
+urllib3
+Werkzeug
```
deanwetherby, line 34: It's uncommon to see a requirements.txt file without pinned version numbers. The purpose of the requirements.txt file is to ensure developers have exactly the same environment. If some of these items are really package dependencies, they should be moved into setup.py (install_requires) where they can have version ranges.

ThomasSchellenbergNextCentury: I understand what you're saying, but I think that having pinned version numbers is equally troublesome.  I spent an hour trying to figure out why mysqlclient wasn't installing correctly, and it was because we were using a massively outdated version.  Also, we'd need to update the package versions every few months to properly maintain the file.  However, since both ways are problematic, if you'd like me to add the version numbers again, I can ensure they're all the most recent.

Also, since these are all development dependencies, I wouldn't recommend putting any of them in the setup.py file.

python_api/machine_common_sense/SPHINX_README.md
```diff
@@ -0,0 +1,22 @@
...
+
+- Good Sphinx Tutorial: https://medium.com/@richdayandnight/a-simple-tutorial-on-how-to-document-your-python-project-using-sphinx-and-rinohtype-177c22a15b5b
+- Markdown Builder: https://pypi.org/project/sphinx-markdown-builder/
+
+- Sphinx: https://www.sphinx-doc.org/en/master/
+- Sphinx's own Tutorial: https://www.sphinx-doc.org/en/master/usage/quickstart.html
+
+## Setup
+
+1. Install Python virtual environment: `sudo apt-get install python3-venv`
```
rartiss55, line 11: On my Mac, I did this for Step 1:

```
pip3 install virtualenv
```

python_api/machine_common_sense/SPHINX_README.md
```diff
@@ -0,0 +1,22 @@
...
+2. Switch to the python_api folder: `cd <mcs>/python_api`
```
rartiss55, line 12: We should update this since now the sphinx_venv needs to be in the machine_common_sense/ directory.

python_api/machine_common_sense/SPHINX_README.md
```diff
@@ -0,0 +1,22 @@
...
+3. Create a virtual environment: `python3 -m venv sphinx_venv`
+4. Activate the virtual environment: `source sphinx_venv/bin/activate`
+5. Update pip: `pip3 install --upgrade pip setuptools wheel`
+6. Install the MCS Python library: `cd ..; pip install -e .; cd python_api`
+7. Install other dependencies: `pip install -r machine_common_sense/requirements.txt`
+8. Enter the docs folder: `cd docs`
+9. Generate the markdown docs in the build folder: `make markdown` (or the HTML docs: `make html`)
+10. Finally, to exit your virtual environment: `deactivate`
+
+Note that you may see the following warning: `WARNING: The config value `source_suffix' has type `dict', defaults to `list'.`
```
rartiss55, line 22: I got this 
```
WARNING: html_static_path entry '_static' does not exist
```

but all my docs seemed to generate fine? Thought I'd mention it just in case.

python_api/machine_common_sense/requirements.txt
rartiss55: I played around with the requirements file and removed mysqlclient (along with SQLAlchemy) since I didn't get why those were there. Things seem to be working. I also took a stab at versioning -- I will try to attach that file in another comment since it won't seem to let me do it here in case anyone wants to try it and make sure it works for not just myself.

rartiss55:
This worked for me (replacing the one in machine_common_sense):

[requirements-works-new.txt](https://github.com/NextCenturyCorporation/MCS/files/5138210/requirements-works-new.txt)

python_api/machine_common_sense/requirements.txt
rartiss55: Okay weird, github deleted my reply here: so I deleted mysqlclient and SQLAlchemy and got things working. I tried adding versioning back in as well and attached that in a comment below in case anyone wants to test it out.

ThomasSchellenbergNextCentury: I will test it out!  Glad you got it working!

python_api/machine_common_sense/SPHINX_README.md
ThomasSchellenbergNextCentury: I'll add that in.

python_api/machine_common_sense/SPHINX_README.md
ThomasSchellenbergNextCentury: Good catch.

python_api/machine_common_sense/SPHINX_README.md
ThomasSchellenbergNextCentury: Huh, I don't get that.  It's probably not important.

python_api/machine_common_sense/requirements.txt
ThomasSchellenbergNextCentury: All the versions in your requirements file are the same or more recent than the versions installed on my machine.  So it looks good!

python_api/machine_common_sense/SPHINX_README.md
ThomasSchellenbergNextCentury: Now that I've upgraded my version of sphinx, I see this warning.  I've just commented out the offending line in the conf.py file, since we aren't currently using sphinx to generate HTML docs anyway.

python_api/machine_common_sense/SPHINX_README.md
deanwetherby: I'm confused by this and it speaks to the need to have projects in their own repositories. Shouldn't there be one and only one venv for the python api? Why do we have a sphinx specific venv?

ThomasSchellenbergNextCentury: The virtual environment is specifically here to generate the sphinx documentation, right?  I can change it to `venv` if you think that's less confusing.

ThomasSchellenbergNextCentury: @rartiss55 I talked with Dean, and he uses a virtual environment for all of his python projects, so it makes sense for him to reuse his existing virtual environment, which is called venv.  So I'm updating all of the scripts and README to rename sphinx_venv to venv.  Please rename your folder as well.

PR #165, opened 2020-07-28
MCS-315 Scene Generator Unit Tests: Pairs, Interactive Goals, Geometry/Util Functions

ThomasSchellenbergNextCentury:
Accomplishments:
- Added a bunch of unit tests for pairs, interactive goals, and geometry/util functions
- Fixed an issue with the geometry function that positions an object adjacent to an existing object
- Fixed an issue with finding similarly colored objects
- Fixed issues with some intermittently broken unit tests
- Removed random choices from unit tests (they will now loop over and validate all possible choices)
- Various cleanup

Outstanding:
- Pair tests sometimes fail due to known issues (usually involving object visibility)
- Some quartet and intphys goal unit tests will intermittently fail (they will be fixed in a future ticket)

### DIFF
--- a/scene_generator/containers_test.py
+++ b/scene_generator/containers_test.py
@@ -1,44 +1,59 @@
+import pytest
 import random
 
 import geometry
 import objects
 import util
 from containers import put_object_in_container, put_objects_in_container, Orientation, can_enclose, \
-    can_contain_both, how_can_contain
+    can_contain_both, how_can_contain, get_enclosable_containments, find_suitable_enclosable_list
 from geometry_test import are_adjacent
 
 
 def test_put_object_in_container():
     obj_def = util.finalize_object_definition(random.choice(objects.get_all_object_defs()))
-    obj_location = geometry.calc_obj_pos(geometry.ORIGIN, [], obj_def)
+    obj_location = geometry.calc_obj_pos({'x': 1, 'y': 0, 'z': 1}, [], obj_def)
     obj = util.instantiate_object(obj_def, obj_location)
+    obj_bounds = obj['shows'][0]['bounding_box']
     container_def = util.finalize_object_definition(random.choice(objects.get_enclosed_containers()))
-    container_location = geometry.calc_obj_pos(geometry.ORIGIN, [], container_def)
+    container_location = geometry.calc_obj_pos({'x': -1, 'y': 0, 'z': -1}, [], container_def)
     container = util.instantiate_object(container_def, container_location)
 
-    put_object_in_container(obj_def, obj, container, container_def, 0)
+    area_index = 0
+    put_object_in_container(obj_def, obj, container, container_def, area_index)
+
     assert obj['locationParent'] == container['id']
     assert obj['shows'][0]['position']['x'] == container_def['enclosed_areas'][0]['position']['x']
-    assert obj['shows'][0]['position']['y'] == container_def['enclosed_areas'][0]['position']['y'] - \
-            (container_def['enclosed_areas'][0]['dimensions']['y'] / 2.0) + obj_def.get('position_y', 0)
+    expected_position_y = container_def['enclosed_areas'][0]['position']['y'] - \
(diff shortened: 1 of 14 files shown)


scene_generator/containers_test.py
```diff
@@ -1,45 +1,77 @@
-import random
+import pytest
```
rartiss55, line 1: Should we add pytest to the requirements.txt?

deanwetherby: I think we should however this brings up a bigger issue. The python API uses unittest. The scene generator uses pytest. My initial thought was to pull scene generator out to its own repo. It could stay in the repo as an MCS script which would be available when installing the package but we would have to make sense of the two unit testing frameworks and consolidate to one. Obviously this would mean re-writing one set of tests.

scene_generator/containers_test.py
```diff
@@ -1,44 +1,59 @@
...
 import random
 
 import geometry
 import objects
 import util
 from containers import put_object_in_container, put_objects_in_container, Orientation, can_enclose, \
-    can_contain_both, how_can_contain
+    can_contain_both, how_can_contain, get_enclosable_containments, find_suitable_enclosable_list
```
deanwetherby, line 7: I noticed this anti-pattern in a few place where we are importing many functions from a module and the list continues to grow ad infinitum. It is preferable (at least to me) to just import the module and then use the module as a namespace for these functions.

```
import containers

containers.how_can_contain()
```

Or if the module name is too long, use a strategy similar to numpy.
```
import containers as ctnr

ctnr.how_can_contain()
```

scene_generator/interaction_goals.py
```diff
@@ -18,11 +18,11 @@
 from goal import Goal, GoalCategory, generate_wall
 from machine_common_sense.mcs_controller_ai2thor import MAX_MOVE_DISTANCE, MAX_REACH_DISTANCE, PERFORMER_CAMERA_Y
 from optimal_path import generatepath
-from util import finalize_object_definition, instantiate_object
+from util import finalize_object_definition, finalize_object_materials_and_colors, instantiate_object
 
 
 def generate_image_file_name(target: Dict[str, Any]) -> str:
-    if 'materials' not in target:
+    if 'materials' not in target or not target['materials']:
```
rartiss55, line 25: I probably just don't understand enough about Python but what's different about these?

deanwetherby: The first part is checking if the key "materials" is in the target dictionary. The second part is checking the value associated with that key.

deanwetherby: Another way to do this is to use the python dictionary getter with a default:
```
if target.get('materials', None) is None:
```

scene_generator/containers_test.py
ThomasSchellenbergNextCentury: Yeah, I just continued with whatever format was being used there already.  I'm assuming this is something that a linter would catch and enforce.  If not, I can make a separate ticket for it.

scene_generator/interaction_goals.py
ThomasSchellenbergNextCentury: Oh, right, there are probably a lot of places that could use that type of cleanup.  In this case, we can also just do `if not target.get('materials')`.  I'm not sure if that would be better to do here or as a separate ticket.

scene_generator/containers_test.py
rartiss55: I made a ticket for this + additional rules we want: https://nextcentury.atlassian.net/browse/MCS-343

scene_generator/interaction_goals.py
rartiss55: Made a ticket: https://nextcentury.atlassian.net/browse/MCS-344

scene_generator/containers_test.py
rartiss55: I left a comment in Teams but I vote pulling out the scene generator into its own repo.

PR #166, opened 2020-07-29
MCS-321 MCS-322 MCS-327 Minor Scene Generator Tasks

ThomasSchellenbergNextCentury:
- Split the object definition 'choose' property into multiple separate properties ('chooseMaterial', 'chooseSize', 'chooseType') to reduce redundancy in object definitions.
- Changed object definition properties from snake_case to camelCase for consistency, based on Dean's request.  (This may affect the Eval 3 UI ingest pipeline.)
- Added 'materialCategory', 'shape', 'size', and (if needed) 'color' properties to all object definitions for continuity.
- Removed 'info_string' because it ended up being redundant with 'goal_string'.

### DIFF
--- a/scene_generator/containers.py
+++ b/scene_generator/containers.py
@@ -10,11 +10,11 @@ import util
 def put_object_in_container(obj_def: Dict[str, Any], obj: Dict[str, Any], container: Dict[str, Any], \
         container_def: Dict[str, Any], area_index: int, rotation: Optional[float] = None) -> None:
 
-    area = container_def['enclosed_areas'][area_index]
+    area = container_def['enclosedAreas'][area_index]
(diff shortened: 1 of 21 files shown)


rartiss55:
I think I know the answer but I’ll ask anyway – is it easy/worth it to change “goal_string” to be “goalString"?

ThomasSchellenbergNextCentury:
@rartiss55 There are a bunch of properties like `goal_string` that we're adding throughout the code.  Yes, it would probably be worth renaming them all at some point.  I was lazy at the time and didn't hunt them all down, but it's probably part of the scope of this ticket.

ThomasSchellenbergNextCentury:
@rartiss55 OK, I renamed goal_string to goalString.

PR #168, opened 2020-08-03
MCS-336

jaudick:
https://nextcentury.atlassian.net/browse/MCS-336

Ai2Thor and scene gen task for sofa materials

### DIFF
--- a/scene_generator/objects.py
+++ b/scene_generator/objects.py
@@ -3912,7 +3912,7 @@ OBJECTS_IMMOBILE = [{
     "obstruct": "vision",
     "info": ["huge", "sofa"],
     "mass": 100,
-    "materialCategory": ["sofa 1"],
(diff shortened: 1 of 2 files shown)


scene_generator/objects.py
```diff
@@ -3910,8 +3910,9 @@
 }, {
     "type": "sofa_1",
     "obstruct": "vision",
-    "info": ["huge", "brown", "sofa"],
+    "info": ["huge", "sofa"],
     "mass": 100,
+    "materialCategory": ["sofa 1"],
```
ThomasSchellenbergNextCentury, line 3915: The materialCategory strings should all have underscores instead of spaces (sofa_1, sofa_chair_1, sofa_2)

jaudick: @ThomasSchellenbergNextCentury done!

PR #170, opened 2020-08-04
MCS-38 - Adding linter/autofixer (only in python_api/ for now)

rartiss55:
Trying to move more towards the recommendations listed here: 
https://www.python.org/dev/peps/pep-0008/

[Flake8](https://flake8.pycqa.org/en/latest/) and [autopep8](https://pypi.org/project/autopep8/0.8/) have been added to the requirements.txt (be sure to install), and should be configured to work within the python_api/ project within vscode and on commit (should point out errors and autoformat files on save when possible). Most of our errors had to do with line length, and those sometimes have to be fixed by hand (yay). Some had to do with unused imports -- had some issues with these but I ran the unit tests and ran a scene as I went along to test and make sure things were still working, but if you wanted to take a closer look/retest things yourself, it will be much appreciated :)

I updated the READMEs, but once you reinstall the python_api/requirements.txt, you’ll need to run this to get the git hook setup:
```
pre-commit install 
```

For some of the trickier line length issues here and there that weren’t autoformatted, I used the code formatter black. I think it’s a neat tool, but its very opinionated and people either love it or hate it. I’ll link it here in case anyone’s curious/we might want to use it in the future: https://github.com/psf/black

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,3 @@
+[flake8]
+ignore = E402, W504
+exclude = python_api/machine_common_sense/__init__.py
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,3 @@
 [flake8]
 ignore = E402, W504
-exclude = python_api/machine_common_sense/__init__.py
+exclude = python_api/machine_common_sense/__init__.py,scene_generator,validation,truthing,masks,occluder,hotfix,setup.py,template_playroom.py
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,16 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+files: 'python_api/'
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.4.0
+    hooks:
+    -   id: check-yaml
+-   repo: https://gitlab.com/pycqa/flake8
+    rev: 3.8.3
+    hooks:
+    -   id: flake8
+-   repo: https://github.com/pre-commit/mirrors-autopep8
+    rev: v1.5.4
+    hooks:
+    -   id: autopep8
--- a/python_api/.vscode/settings.json
+++ b/python_api/.vscode/settings.json
@@ -1,0 +1,21 @@
+{
+    "python.linting.enabled": true,
+    "python.linting.flake8Enabled": true,
+    "python.linting.lintOnSave": true,
+    "python.linting.flake8Args": [
+        "--ignore",
+        "E402,W504",
+        "--exclude",
+        "__init__.py"
+    ],
+    "python.formatting.provider": "autopep8",
+    "python.formatting.autopep8Args": [
+        "--aggressive",
+        "--aggressive",
+        "--ignore",
+        "E402"
+    ],
+    "[python]": {
+        "editor.formatOnSave": true
+    }
+}
--- a/python_api/machine_common_sense/README.md
+++ b/python_api/machine_common_sense/README.md
@@ -74,8 +74,8 @@ See https://numpydoc.readthedocs.io/en/latest/format.html
 ## Linting
 
 We are currently using [flake8](https://flake8.pycqa.org/en/latest/) and [autopep8](https://pypi.org/project/autopep8/) for linting and formatting our Python code. This is only enforced within the python_api project for now. Both are [PEP 8](https://www.python.org/dev/peps/pep-0008/) compliant, although we are ignoring the following rules:
-**E402**: Module level import not at top of file
-**W504**: Line break occurred after a binary operator
+- **E402**: Module level import not at top of file
+- **W504**: Line break occurred after a binary operator
 
 A full list of error codes and warnings enforced can be found [here](https://flake8.pycqa.org/en/latest/user/error-codes.html)
 
--- a/python_api/machine_common_sense/README.md
+++ b/python_api/machine_common_sense/README.md
@@ -12,12 +12,7 @@
 2. Install the Python dependencies (I tested on Python v3.6.5)
 
 ```
-pip install ai2thor
-pip install Pillow
-pip install numpy
-pip install autopep8
-pip install flake8
-pip install pre-commit
+pip install -r ../requirements.txt
 ```
 
 3. Run pre-commit install to set up git hook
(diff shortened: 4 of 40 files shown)


.flake8
```diff
@@ -0,0 +1,3 @@
+[flake8]
+ignore = E402, W504
```
rartiss55, line 2: E402: We have some imports in the middle of one of our controllers, and moving them broke some unit tests, so I added that rule to the ignore list. 

W504: PEP8 recommends we line break before a binary operator, but we were doing it after in most places/technically you can do either in flake8, so I'm ignoring this rule.

.flake8
```diff
@@ -0,0 +1,3 @@
...
+exclude = python_api/machine_common_sense/__init__.py
```
rartiss55, line 3: Excluded this file since I got complaints about unused imports.

python_api/machine_common_sense/getchHelper.py
```diff
@@ -12,15 +14,19 @@ def __init__(self):
...
 
-    def __call__(self): return self.impl()
+    def __call__(self):
+        return self.impl()
 
 
 class _GetchUnix:
     def __init__(self):
-        import tty, sys
+        import tty  # noqa: F401
```
rartiss55, line 23: Wherever you see 'noqa', I'm ignoring a rule in-line (in this case, I didn't think this fell under "unused import").

rartiss55:
Also, let me know if it'd be better to merge this into the eval-3 branch instead.

python_api/machine_common_sense/README.md
```diff
@@ -14,8 +14,19 @@
 ```
 pip install ai2thor
 pip install Pillow
+pip install numpy
+pip install autopep8
+pip install flake8
+pip install pre-commit
 ```
```
ThomasSchellenbergNextCentury, line 21: Since we added a requirements file after this was written, would it make more sense to just replace it with a `pip install -r requirements.txt` ?

deanwetherby: Yes!

ThomasSchellenbergNextCentury:
@rartiss55 A few quick questions:
1. It looks like `.flake8` is the rule file?
2. Are the PEP 8 rules all the defaults?
3. How do I manually run the linter? I think a "linter" section in the README would be helpful, maybe with some links to the default rules and the rule IDs (like E402, etc.)
4. Did we end up doing anything with validating type hints? I know we're not using them in the python_api (just the scene_generator), but we might want to add them in the future.

python_api/machine_common_sense/mcs_step_output.py
```diff
@@ -87,23 +106,33 @@ def __init__(
...
+        )
         self.pose = pose
         self.position = {} if position is None else position
         self.return_status = return_status
         self.reward = reward
         self.rotation = rotation
         self.step_number = step_number
-        self.structural_object_list = [] if structural_object_list is None else structural_object_list
+        self.structural_object_list = [
+        ] if structural_object_list is None else structural_object_list
```
ThomasSchellenbergNextCentury, line 135: So all of these line reformats are from either `autopep8` or `black`?  They mostly seem OK, although can be a little weird at times, like here.  But I guess having an automated tool is better than forcing you to do it manually.  We'll probably want to add instructions on installing/using `black` to the README along with the linter instructions.  Also, are there any configuration options for how the lines are reformatted (or maybe they're PEP 8 standards)?

python_api/.vscode/settings.json
```diff
@@ -0,0 +1,21 @@
+{
```
ThomasSchellenbergNextCentury, line 1: Open question: Do we want VS Code files saved in the repo?

rartiss55: I thought in this case yes, since this configures the linter to work properly in vscode.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,16 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+files: 'python_api/'
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.4.0
+    hooks:
+    -   id: check-yaml
```
ThomasSchellenbergNextCentury, line 8: Do we need the `check-yaml`?

ThomasSchellenbergNextCentury:
Looks good.  And don't worry about the eval-3 branch; that's scene generator changes.

python_api/machine_common_sense/mcs_step_output.py
rartiss55: Autopep8 didn't help as much with the line breaks and it seems like that was by design (i.e. how does it really know where it should break up your lines outside of comments). And some of what you're saying is why I didn't add black officially -- I more so used it when I was having difficulty breaking up a really long line, and I copy pasted only those lines back in. Black is PEP8 compliant but has its own formatting standards that it enforces that not everyone agrees with and that you cannot change.

.pre-commit-config.yaml
rartiss55: I thought it'd be useful if we changed this config ever for validating. This came with a bunch of defaults that I removed (things like "add new lines at end of file") but I kept this one.

ThomasSchellenbergNextCentury:
This looks good and works well.  It even verifies that your changes are staged for commit, which is better than our linter in Neon. 
 I think we just need some documentation as you noted at standup this morning.

And don't worry about the eval-3 branch; that's scene generator changes.

python_api/machine_common_sense/mcs_step_output.py
ThomasSchellenbergNextCentury: Personally I'm fine with including it here in the repo, even if I don't agree with all of its formatting, but I'll let @deanwetherby chime in too.

rartiss55:
> @rartiss55 A few quick questions:
> 
> 1. It looks like `.flake8` is the rule file?
> 2. Are the PEP 8 rules all the defaults?
> 3. How do I manually run the linter? I think a "linter" section in the README would be helpful, maybe with some links to the default rules and the rule IDs (like E402, etc.)
> 4. Did we end up doing anything with validating type hints? I know we're not using them in the python_api (just the scene_generator), but we might want to add them in the future.

@ThomasSchellenbergNextCentury 
1.) Yes. Ended up adding it last more to get the git hook to work properly (was running it on the command line previously).
2.) Yes (minus the exceptions listed). flake8 is a wrapper around different checkers and it is PEP8 compliant, and those rules/errors are prefixed with E or W. Admittedly I didn't look super closely at the other checkers contained (like PyFlakes) but I didn't have major complaints about the other rules (minus some of the false positives for unused imports).
3.) Working on adding things to the README :)
4.) No -- I thought I found something that did this but it was a really simple plugin that didn't seem very useful. I think mypy does this but I wanted to play with it some more. I made a ticket for it https://nextcentury.atlassian.net/browse/MCS-342 (thanks for the reminder)

python_api/machine_common_sense/README.md
```diff
@@ -65,6 +71,31 @@ Packaging the project is done with the [setup.py](../../setup.py) file located i
 
 See https://numpydoc.readthedocs.io/en/latest/format.html
 
+## Linting
```
rartiss55, line 74: Added Linting section (hope this was the right place). Also went back and updated our .flake8 config ignore list, since just running the flake8 command in a top level directory will result in errors outside of the python_api/ project.

python_api/machine_common_sense/mcs_step_output.py
ThomasSchellenbergNextCentury: Going to merge this in now, but we can add `black` instructions later if we want.

PR #171, opened 2020-08-07
Rename python classes

arashBehpour:
I renamed for a couple of MCS classes. Just wanted to put my work here so that anyone else can easily finish this ticket.


deanwetherby:
@arashBehpour Thanks for pushing up your progress on this ticket.

For whomever picks up this ticket:
Let's be sure to rename the files/modules associated with these class fixes. 
One potential issue that I noticed is renaming MCS_Object to object. As you know, object is a python keyword. We will need to think of another noun to replace it.

python_api/machine_common_sense/object_.py
```diff
@@ -1,7 +1,8 @@
-from .mcs_util import MCS_Util
+from .material import Material
+from .util import Util
 
 
-class MCS_Object(object):
+class Object(object):
```
rartiss55, line 5: Just some brainstorming if we want to call this something other than object: item, attributes, traits, some combination of those?

ThomasSchellenbergNextCentury: I'm thinking "ObjectMetadata"

ThomasSchellenbergNextCentury:
Closing because I've replaced this PR with https://github.com/NextCenturyCorporation/MCS/pull/179

PR #174, opened 2020-08-17
MCS-38 - Autoformatting on scene generator code

rartiss55:
(Take 2)

Still need to fix additional linter errors, so the scene_generator code hasn't yet been added to the pre-commit hook (there will be an additional PR for this). I added a vscode settings file for the scene generator, so that should cover any automatically fixable formatting issues for the time being. Also added autopep8 and flake8 to the scene generator's requirements.txt since I had left that out before.

@ThomasSchellenbergNextCentury Ran autopep8 on a lower aggressiveness level -- should fix the way line breaks were done per argument on long function names, but if there's still any issue with how the autoformatter works, please let me know (see lines 10 and 176 in containers.py for a couple examples).

### DIFF
--- a/scene_generator/containers.py
+++ b/scene_generator/containers.py
@@ -7,14 +7,15 @@ import objects
 import util
 
 
-def put_object_in_container(obj_def: Dict[str, Any], obj: Dict[str, Any], container: Dict[str, Any], \
-        container_def: Dict[str, Any], area_index: int, rotation: Optional[float] = None) -> None:
+def put_object_in_container(obj_def: Dict[str, Any], obj: Dict[str, Any], container: Dict[str, Any],
+                            container_def: Dict[str, Any], area_index: int, rotation: Optional[float] = None) -> None:
 
     area = container_def['enclosedAreas'][area_index]
     obj['locationParent'] = container['id']
 
     obj['shows'][0]['position'] = area['position'].copy()
-    obj['shows'][0]['position']['y'] += -(area['dimensions']['y'] / 2.0) + obj_def.get('positionY', 0)
+    obj['shows'][0]['position']['y'] += - \
+        (area['dimensions']['y'] / 2.0) + obj_def.get('positionY', 0)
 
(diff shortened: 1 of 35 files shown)


python_api/.vscode/settings.json
```diff
@@ -10,7 +10,6 @@
     ],
     "python.formatting.provider": "autopep8",
     "python.formatting.autopep8Args": [
-        "--aggressive",
         "--aggressive",
```
ThomasSchellenbergNextCentury, line 13: Why was "--aggressive" listed here twice?

scene_generator/containers.py
```diff
@@ -7,14 +7,15 @@
 import util
 
 
-def put_object_in_container(obj_def: Dict[str, Any], obj: Dict[str, Any], container: Dict[str, Any], \
-        container_def: Dict[str, Any], area_index: int, rotation: Optional[float] = None) -> None:
+def put_object_in_container(obj_def: Dict[str, Any], obj: Dict[str, Any], container: Dict[str, Any],
+                            container_def: Dict[str, Any], area_index: int, rotation: Optional[float] = None) -> None:
```
ThomasSchellenbergNextCentury, line 11: This looks much better.  What's the line max character limit?  Will that be updated in a future commit?

python_api/.vscode/settings.json
rartiss55: This argument enables non-whitespace changes, and multiple instances of -a or --aggressive would apply those changes more aggressively (in my case, I wanted to shorten lines more aggressively). It wasn't 100% clear specifically what 2 aggressives vs 3, etc gets you from the documentation, but I just went with two.

https://github.com/hhatto/autopep8#more-advanced-usage

scene_generator/containers.py
rartiss55: Line limit is still 79. Since I took away one of the "aggressive" arguments, autopep8 didn't shorten as many lines automatically (but flake8 will still throw an error).

PR #175, opened 2020-08-20
MCS-320 Removed X and Z offsets from scene generator object definitions.

ThomasSchellenbergNextCentury:
Corresponding Unity PR: https://github.com/NextCenturyCorporation/ai2thor/pull/82

Please note that I didn't change the offsets of the potted plants, because that involves some non-trivial evaluation design decisions.  For now, I've just removed the potted plants from being added to generated scenes, since they have an issue with spontaneously falling over, and I don't think they're adding much to the evaluation anyway (until I'm told otherwise).


python_api/scenes/playroom.json
```diff
@@ -1,4 +1,5 @@
 {
+    "version": 2,
```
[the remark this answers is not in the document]
ThomasSchellenbergNextCentury: If we don't add the commit hashes directly in the JSON files, we can at least make a list of them all in Confluence.

rartiss55, line 2: Are we planning to add an explicit version to the scene generator scenes/is that something we'd want?

ThomasSchellenbergNextCentury: @rartiss55 That's an open question (though approving this PR will finalize that decision).  I had chatted about the idea with @deanwetherby a bit and he seemed to like it.  I'm only really concerned at this point about backwards compatibility, since I wish we had the foresight to add version tags to the Eval 2 datasets.  But I think we should just bite the bullet now and put them in.  I believe that doing so will help us to more easily make breaking changes without completely invalidating older datasets, since we can choose to have different behavior associated with the different version tags.

rartiss55: I think we should go ahead and do this since I agree, it will help us better identify issues going forward. If we really need to look at the old ones we can always add "version: 1" to them right?

PR #182, opened 2020-09-03
Combine Action, ActionKeys and ActionDesc into one enum (MCS-352).

rartiss55:
Action.SOME_ACTION.value will still return the action method name, but now Action has additional attributes for key and desc.

### DIFF
--- a/python_api/machine_common_sense/action.py
+++ b/python_api/machine_common_sense/action.py
@@ -7,7 +7,13 @@ class Action(Enum):
     The actions available in the MCS simulation environment.
     """
 
-    CLOSE_OBJECT = "CloseObject"
+    CLOSE_OBJECT = (
+        "CloseObject",
+        "1",
+        "Close a nearby object. (objectId=string, amount=float " +
+        "(default:1), objectDirectionX=float, objectDirectionY=float, " +
+        "objectDirectionZ=float)"
+    )
     """
     Close a nearby object.
 
@@ -51,7 +57,11 @@ class Action(Enum):
         Unexpected error; please report immediately to development team.
     """
 
-    CRAWL = "Crawl"
+    CRAWL = (
+        "Crawl",
+        "c",
+        "Change pose to 'CRAWLING' (no params)"
+    )
     """
     Change pose to "CRAWLING". Can help you move underneath or over objects.
 
@@ -65,7 +75,11 @@ class Action(Enum):
         Unexpected error; please report immediately to development team.
     """
 
-    DROP_OBJECT = "DropObject"
+    DROP_OBJECT = (
+        "DropObject",
+        "2",
+        "Drop an object you are holding. (objectId=string)"
+    )
     """
     Drop an object you are holding.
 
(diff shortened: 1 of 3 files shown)


python_api/machine_common_sense/action.py
```diff
@@ -487,71 +581,25 @@ class Action(Enum):
...
-    ROTATE_LOOK = "r"
-    ROTATE_OBJECT_IN_HAND = "t"
-    STAND = "u"
-    THROW_OBJECT = "q"
+    def __new__(cls, action, key, desc):
+        obj = object.__new__(cls)
+        obj._value_ = action
+        obj._key = key
+        obj._desc = desc
+        cls._value2member_map_[key] = obj
```
rartiss55, line 589: This line was so that the keyboard key could also be used to access the correct enum (needed on line 65 of run_human_input) -- I used this Stack Overflow answer if anyone's curious: https://stackoverflow.com/questions/43202777/get-enum-name-from-multiple-values-python

deanwetherby: Thanks for the link.

deanwetherby (approved):

python_api/machine_common_sense/run_human_input.py
```diff
@@ -244,7 +243,8 @@ def main():
         print(status)
         exit()
 
-    controller = MCS.create_controller(args.mcs_unity_build_file, debug=args.debug,
+    controller = MCS.create_controller(args.mcs_unity_build_file,
```
brianpippin, line 246: Was this a linter fix or were you moving parameters to their own line, if the latter do you want to split up line 248 as well?

brianpippin (approved):

python_api/machine_common_sense/action.py
ThomasSchellenbergNextCentury: Might be helpful to leave a comment for this line of code.

python_api/machine_common_sense/action.py
```diff
@@ -487,71 +581,25 @@ class Action(Enum):
...
+            ', '.join([self._value_, self._key, self._desc])
+        )
+
+    @ property
+    def key(self):
+        return self._key
+
+    @ property
+    def desc(self):
+        return self._desc
```
ThomasSchellenbergNextCentury, line 605: Anything in here worth unit testing?

python_api/machine_common_sense/action.py
rartiss55: Added a comment for this

python_api/machine_common_sense/run_human_input.py
rartiss55: It was a linter fix, but I think its better to split that line up as well (good catch). Fixed.

python_api/machine_common_sense/action.py
rartiss55: I added tests for the action class -- they're probably nice to have in case we have to update any of the action keyboard keys or descriptions.

ThomasSchellenbergNextCentury (approved):

PR #183, opened 2020-09-08
MCS-334 Scene Generator IntPhys Quartet Refactor and Unit Testing

ThomasSchellenbergNextCentury:
- Refactored and simplified the Pair class.
- Tidied the IntPhysGoal and Quartet classes.
- Expanded the unit test suite for the IntPhysGoal and Quartet.
- Fixed the following bugs:
  - MCS-324 Maintain consistent rotation of teleporting objects
  - MCS-326 Don't switch two objects in fall-down STC quartets because it may look like shapes changing
  - MCS-356 Ensure occluders don't use the same materials as the walls
  - Various other misc. bugs

All of the scene generator's unit tests should now pass, except for possibly the ones testing the interactive goals' pathfinding.


deanwetherby:
Scene generator code is looking good. I did a fresh checkout of MCS and switched to this branch. Made a new virtual environment and installed requirements.txt as well as MCS. pytest currently isn't included in requirements.txt so that had to be installed separately. I would recommend adding that developer dependency to requirements.txt.

When the unit tests were run, a few modules needed to reflect our recent changes in class renames. In particular, I had to change goals.py, interaction_goals.py and interaction_goals_test.py to remove the 'mcs_' prefix from controller_ai2thor. Once those changes were made the tests will now run although I am encountering a SceneException with test_ShapeConstancyQuartet_get_scene_3_fall_down. After a few attempts, the tests passed completely 1 out of 4 times.

If you want, I can commit the changes described above to this branch.

ThomasSchellenbergNextCentury:
@deanwetherby Thanks for all the good feedback.

- I added pytest to the requirements file.
- I updated the MCS class imports to remove the `mcs_` prefix (that's what I get for merging master and not rerunning the tests!)
- I hope the ShapeConstancyQuartet error was `Cannot find shape constancy substitute object`, because that's an issue I forgot to handle yesterday before making this PR.  That issue is now fixed, and I added an extra unit test for it.
- I added "skips" to the finicky pathfinding unit tests in the interaction_goals_test.
-

deanwetherby (approved):

brianpippin (approved):
I couldn't find anything that jumped out at me.

PR #184, opened 2020-09-16
MCS-355 Major scene generator refactor to replace the Goal, Pair, and Quartet classes with Sequences

ThomasSchellenbergNextCentury:
- Renamed IntPhys to Intuitive Physics everywhere.
- Created a Sequence class that makes a "sequence" (pair, quartet, octet, etc.) of one or more linked scenes. Now the Sequence, rather than the Goal, manages the creation of objects (before, this caused some particularly problematic interdependencies between the Pair and InteractionGoal classes).
- Removed the IntPhysQuartet and IntPhysGoal, and added the IntuitivePhysicsSequence class (combining the "goal" with the "sequence", since they are the same for intuitive physics).
- Removed the Goal, InteractionGoal, and Pair, and added the InteractiveGoal (as a standalone class), InteractivePlan (that identifies specific pairs), and InteractiveSequence classes. This combined most functionality from the former Pair and InteractionGoal classes into the InteractiveSequence, decoupling and simplifying the merged code.
- Simplified the scene generator main class.
- Updated the other classes as needed.


deanwetherby:
This seems sort of moot now. I'm tempted to just approve and move on to designing the new tests after moving scene generator elsewhere.

PR #186, opened 2020-09-23
MCS-368 Python package reorg

deanwetherby:
Cleaned up git repository to be more python package friendly.

* removed scene generator code
* removed unneeded python_api folder and bumped everything up a level
* made a scripts folder for run_*.py and generate* scripts
* updated all run scripts to use argparse instead of sys.argv
* fixed docstring locations in util.py. The docstring was above the function instead of below.
* updated unittest procedures. You no longer need to be in the python_api folder.

Comments:
* a few of our modules/classes would benefit from dataclasses (https://docs.python.org/3/library/dataclasses.html). Although this is a 3.7 feature, it is pip installable for 3.6. We could add it as a package dependency.
* I'm not sure why we need the mcs module and controller vs controller_ai2thor. We could just have a single namespaced MCS controller.
* We should move the very large animated gifs to s3 and just provide a link to the markdown rather than having them checked into the repository.

Open questions:
* where to put the non-README markdowns like DEV.md and API.md? Right now I put them in machine_common_sense. It seems too cluttered to have them at the project root.
* run scripts have path assumptions. is that ok? some require that you run from inside the scripts folder whereas you might run like 'python scripts/run_whatever.py'.


ThomasSchellenbergNextCentury:
@deanwetherby I removed the scene_generator folder from the master branch, so you should merge the master branch into this branch and resolve conflicts.

.gitignore
```diff
@@ -27,39 +25,4 @@ dist/
...
-
-# test run script for Python API
-python_api/machine_common_sense/run_mcs_env.py
-
-# python docs
-python_api/machine_common_sense/docs
-python_api/machine_common_sense/venv
-
-# scene runtime debug files
-python_api/machine_common_sense/**/*.gif
```
rartiss55, line 47: Hmm maybe we want to keep the runtime debug ignores in here? (I guess now they would be **/*.extension)

.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
-files: 'python_api/|scene_generator/'
+files: 'machine_common_sense/'
```
rartiss55, line 3: I think we also want scripts/ and tests/ here (so we don't miss any flake8 violations on commit).

.vscode/settings.json
```diff
@@ -1,4 +1,15 @@
...
+    "python.testing.unittestArgs": [
+        "-v",
+        "-s",
+        "./",
+        "-p",
+        "test_*.py"
+    ],
+    "python.testing.pytestEnabled": false,
+    "python.testing.nosetestsEnabled": false,
+    "python.testing.unittestEnabled": true,
```
rartiss55, line 12: I like the testing support!

machine_common_sense/DEV.md
```diff
@@ -0,0 +1,104 @@
...
+See https://numpydoc.readthedocs.io/en/latest/format.html
+
+## Running
+
+We have made multiple run scripts:
+
+To run a script (like `run_human_input.py`) from the terminal with visual output:
+
+```
+run_human_input <mcs_unity_build_file> <mcs_config_json_file>
```
rartiss55, line 79: Doesn't this need to be "python run_human_input.py <mcs_unity_build_file> <mcs_config_json_file>"?

rartiss55:
@deanwetherby I checked out the branch/re-set up the virtual environment and ran a couple scripts to test. Looks good. In addition to my earlier comments:
- It looks like we added some additional files from docs/ into the repo -- just double checking that that was intentional
- If so, would the script that runs for the sphinx commit hook need to be updated to incorporate those things?

rartiss55:
Forgot about the open questions:
- I think putting API.md and DEV.md in machine_common_sense/ is fine
- I think the path assumptions are okay for now as long as its documented

machine_common_sense/DEV.md
deanwetherby: The package's setup.py creates a few scripts that can be run directly. It's what we had in the setup.py already so I kept it.

    entry_points={
        'console_scripts': [
            'run_in_human_input_mode=scripts.run_human_input:main',
            'run_scene_timer=scripts.run_scene_timer:main'
        ]
    }

.vscode/settings.json
deanwetherby: @rartiss55 Once the package was set up in the way vscode recognized, it automatically added those settings! I like it too

deanwetherby:
> * It looks like we added some additional files from docs/ into the repo -- just double checking that that was intentional

I needed a second or third pair of eyes on the sphinx docs. Nothing was intentional from my perspective.

machine_common_sense/DEV.md
rartiss55: Oh yeah, but I thought that part is mentioned in another section of the readme (run_in_human_input_mode) and this is just for running all the scripts in the scripts folder, right?

.gitignore
```diff
@@ -27,39 +25,4 @@ dist/
...
-python_api/machine_common_sense/**/*.json
-python_api/machine_common_sense/**/*.mp4
-python_api/machine_common_sense/**/*.png
```
ThomasSchellenbergNextCentury, line 50: Why are all of these being removed?  Shouldn't we just be removing the `python_api/` prefix?

README.md
```diff
@@ -1,17 +1,212 @@
...
-## Scene Generator
+Python virtual environments are recommend when using the MCS package. All steps presume the activation of the virtual environment as shown below.
 
-Scripts for generating scenes of training data.
-[Link to Scene Generator](scene-generator/README.md) 
+```
+python3.6 -m venv --prompt mcs venv
+source venv/bin/activate
+python -m pip install --upgrade pip setuptools wheel
+```
```
ThomasSchellenbergNextCentury, line 15: So originally this README was intended for non-developers (users like the TA1s) while the `machine_common_sense/README` was intended for developers.  It seems to me that we shouldn't be making non-developers install a virtual environment just to use our software?

.gitignore
```diff
@@ -27,39 +25,4 @@ dist/
...
-python_api/scenes/*.json
-python_api/scenes/images/*.gif
-
 **/SCENE_HISTORY
-
-# test run script for Python API
-python_api/machine_common_sense/run_mcs_env.py
-
-# python docs
-python_api/machine_common_sense/docs
```
ThomasSchellenbergNextCentury, line 43: I believe that we definitely should NOT be including all of the build docs in our repo.  All we need is the final markdown API file.

.gitignore
ThomasSchellenbergNextCentury: I think we should just move them all into the `scripts/` folder, like `scripts/**/*.gif`, etc.  It would make debugging much easier.

machine_common_sense/DEV.md
ThomasSchellenbergNextCentury: You either have to run `python run_human_input.py` or `run_in_human_input_mode` (see how we rename the script in the `console_scripts`).

ThomasSchellenbergNextCentury:
> - I'm not sure why we need the mcs module and controller vs controller_ai2thor. We could just have a single namespaced MCS controller.

There was an open question as to whether we'd need to support hooks into multiple environments.  Unless you feel strongly otherwise, I'm inclined to maintain the existing structure, since we don't know one way or another right now.

Additionally, we've been using the `MCS` class to provide public utility functions for the TA1s (as opposed to the `util` class, which has our internal utility functions).  I'm honestly not sure if those would make sense in the controller class.

> - We should move the very large animated gifs to s3 and just provide a link to the markdown rather than having them checked into the repository.

Sure.

> - where to put the non-README markdowns like DEV.md and API.md? Right now I put them in machine_common_sense. It seems too cluttered to have them at the project root.

My opinion:  DEV and other development-specific files make sense in machine_common_sense, but API should be in root, since non-developers (users like the TA1s) will need to access it.

> - run scripts have path assumptions. is that ok? some require that you run from inside the scripts folder whereas you might run like 'python scripts/run_whatever.py'.

I think that's fine, since those scripts are intended for use by our team as a convenience to generate examples for the psych team.  Someone (I'll volunteer) could add a short README to the `scripts` folder to describe each script.

README.md
deanwetherby: It's a recommendation/suggestion to use venv and I provided that example. They do whatever they want including no venv, conda, or any of the numerous other python solutions but nearly all python project development is done some sort of isolated virtual environment.

deanwetherby:
I removed the base controller module and renamed the controller_ai2thor to just controller. We don't have a large enough team to maintain two simulation environments although I understand the thought process behind the base class. If we do have to switch to another environment, we'll just rewrite the controller module but we won't need the base class.

I also removed the MCS class inside the mcs module and bubbled up the class methods to functions. That particular class only had static methods so it was not needed. By removing it and changing the package init to expose the module functions, usage of the package becomes much cleaner and this removes the last remnants of the MCS over-marketing. Rather than 
```
from machine_common_sense.mcs import MCS
controller = MCS.create_controller()
```
 we can now just do 
```
controller = mcs.create_controller()
```.

This should look familiar to typical numpy usage
```
import numpy as np
arr = np.ndarray()
```

All of our class can now be namespaced this way (mcs.Action, mcs.Reward, etc) which I think is cleaner and more explicit.  A developer using MCS can still choose to do it a different way but our examples and unit tests should use the mcs namespacing.

File: .gitignore (hunk 1 above)
Comment (rartiss55) [reply .gitignore:47]: I like the idea of having these debug files in a designated folder.

Comment (deanwetherby) [reply .gitignore:47]: Can I get some more clarity on this? I'm not on the same page as you two. Are these the scene images?

Comment (ThomasSchellenbergNextCentury) [reply .gitignore:47]: If you run a scene with the Python API in "debug" mode, it generates a lot of files containing the input and output metadata (image and JSON files).  It's usually convenient for me to run the scripts in the python_api or machine_common_sense directory so I don't have to keep changing directories back and forth while I'm working, and don't have to worry about the path to the script file.  Since the debug files are generated within the current directory, we don't want to accidentally commit them into the repo.  That's why these lines are in the gitignore file.

**Comment by rartiss55** [comment]
> > * It looks like we added some additional files from docs/ into the repo -- just double checking that that was intentional
> 
> I needed a second or third pair of eyes on the sphinx docs. Nothing was intentional from my perspective.

@deanwetherby I don't think we need all the files that are checked in for `docs/`. I think Thomas mentioned this elsewhere, but the markdown created in `docs/` by sphinx gets copied to API.md anyway. These were the only files present in the repo from the sphinx `docs/` folder prior to this:
- Makefile
- conf.py
- index.rst 
- make.bat

I think we need to put back the .gitignore for these (which I guess would now be `docs/` instead of `machine_common_sense/docs/`).

File: scripts/generate_sphinx_docs.sh [hunk 8]
```diff
@@ -0,0 +1,21 @@
...
+then
+    exit 0
+fi
+echo "moving to docs"
+cd docs/
+echo "calling make'"
+make markdown
+echo "moving to API"
+cp _build/markdown/index.md ../machine_common_sense/API.md
+git add machine_common_sense/API.md
```
rartiss55, line 20: Just ran into this -- this needs to be `git add **../**machine_common_sense/API.md`

machine_common_sense/DEV.md
rartiss55: Rereading this -- in case my comment was unclear, I think this was a typo -- this either needs to be `python run_human_input.py` or `run_in_human_input_mode` (the way it is now, `run_human_input`, will say "command not found").

ThomasSchellenbergNextCentury (approved):

rartiss55 (approved):

PR #187, opened 2020-10-06
MCS-383: Add History Enabled Flag

brianpippin:
This code review is doing multiple things:
1.  Adds the flag for history_enabled so that a user could turn off the history file logging if wanted
2. Moves all the history file code to a new module
3. Cleans up the history txt file and makes it a proper json file

In another ticket we will add all the new fields Thomas created in the scene files to the history file.

### DIFF
--- a/python_api/machine_common_sense/controller_ai2thor.py
+++ b/python_api/machine_common_sense/controller_ai2thor.py
@@ -305,7 +305,11 @@ class ControllerAI2THOR(Controller):
         self.__scene_configuration = config_data
         self.__step_number = 0
         self._goal = self.retrieve_goal(self.__scene_configuration)
-        self.__history_writer = HistoryWriter(config_data)
+
+        if self.__history_writer is None:
+            self.__history_writer = HistoryWriter(config_data)
+        else:
+            self.__history_writer.check_file_written()
 
         skip_preview_phase = (True if 'goal' in config_data and
                               'skip_preview_phase' in config_data['goal']
--- a/python_api/machine_common_sense/controller_ai2thor.py
+++ b/python_api/machine_common_sense/controller_ai2thor.py
@@ -73,7 +73,8 @@ class ControllerAI2THOR(Controller):
         A seed for the Python random number generator.
         (default None)
     history_enabled: boolean, optional
-        Whether to save all the history files and generated image history to local disk or not.
+        Whether to save all the history files and generated image
+        history to local disk or not.
         (default False)
     """
 
@@ -158,7 +159,7 @@ class ControllerAI2THOR(Controller):
 
     def __init__(self, unity_app_file_path, debug=False,
                  enable_noise=False, seed=None, size=None,
-                 depth_masks=False, object_masks=False, 
+                 depth_masks=False, object_masks=False,
                  history_enabled=True):
         super().__init__()
 
@@ -193,7 +194,7 @@ class ControllerAI2THOR(Controller):
             self.__screen_height = size / 3 * 2
 
     def _update_internal_config(self, enable_noise=None, seed=None,
-                                depth_masks=None, object_masks=None, 
+                                depth_masks=None, object_masks=None,
                                 history_enabled=None):
 
         if enable_noise is not None:
@@ -261,7 +262,6 @@ class ControllerAI2THOR(Controller):
     def get_seed_value(self):
         return self.__seed
 
-
     # Override
     def end_scene(self, choice, confidence=1.0):
         """
@@ -281,7 +281,6 @@ class ControllerAI2THOR(Controller):
 
         super().end_scene(choice, confidence)
         self.__history_writer.write_history_file(choice, confidence)
-
 
     # Override
     def start_scene(self, config_data):
@@ -648,7 +647,6 @@ class ControllerAI2THOR(Controller):
                     'ContentType': 'image/png',
                 }
             )
-
 
     def generate_time(self):
         return datetime.datetime.now().strftime("%Y%m%d-%H%M%S")
--- a/python_api/machine_common_sense/controller_ai2thor.py
+++ b/python_api/machine_common_sense/controller_ai2thor.py
@@ -36,6 +36,7 @@ from .reward import Reward
 from .scene_history import SceneHistory
 from .step_metadata import StepMetadata
 from .util import Util
+from .history_writer import HistoryWriter
 
 # From https://github.com/NextCenturyCorporation/ai2thor/blob/master/ai2thor/server.py#L232-L240 # noqa: E501
 
@@ -71,6 +72,9 @@ class ControllerAI2THOR(Controller):
     seed : int, optional
         A seed for the Python random number generator.
         (default None)
+    history_enabled: boolean, optional
+        Whether to save all the history files and generated image history to local disk or not.
+        (default False)
     """
 
     ACTION_LIST = [item.value for item in Action]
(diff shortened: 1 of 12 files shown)


python_api/machine_common_sense/object_metadata.py
```diff
@@ -89,3 +89,23 @@ def __init__(
...
+        yield 'distance_in_steps', self.distance_in_steps
+        yield 'distance_in_world', self.distance_in_world
+        yield 'held', self.held
+        yield 'mass', self.mass
+        yield 'material_list', self.material_list
+        yield 'position', self.position
+        yield 'rotation', self.rotation
+        yield 'shape', self.shape
+        yield 'texture_color_list', self.texture_color_list
+        yield 'visible', self.visible
```
deanwetherby, line 110: Another couple of ways to convert a class into a dict for your consideration.

```
from machine_common_sense.object_metadata import ObjectMetadata
ob = ObjectMetadata()
vars(ob)
{'uuid': '', 'color': {}, 'dimensions': {}, 'direction': {}, 'distance': -1.0, 'distance_in_steps': -1.0, 'distance_in_world': -1.0, 'held': False, 'mass': 0.0, 'material_list': [], 'position': {}, 'rotation': {}, 'shape': '', 'texture_color_list': [], 'visible': False}

# or with the built-in class attribute

ob.__dict__
{'uuid': '', 'color': {}, 'dimensions': {}, 'direction': {}, 'distance': -1.0, 'distance_in_steps': -1.0, 'distance_in_world': -1.0, 'held': False, 'mass': 0.0, 'material_list': [], 'position': {}, 'rotation': {}, 'shape': '', 'texture_color_list': [], 'visible': False}
```

deanwetherby (approved):

python_api/machine_common_sense/controller_ai2thor.py
```diff
@@ -311,14 +304,9 @@ def start_scene(self, config_data):
...
         self.__step_number = 0
-        self.__history_list = []
         self._goal = self.retrieve_goal(self.__scene_configuration)
-        if 'screenshot' in config_data and config_data['screenshot']:
-            self.__scene_history_file = None
-        else:
-            self.__scene_history_file = os.path.join(
-                self.HISTORY_DIRECTORY, config_data['name'].replace(
-                    '.json', '') + "-" + self.generate_time() + ".txt")
+        self.__history_writer = HistoryWriter(config_data)
```
ThomasSchellenbergNextCentury, line 308: We had talked about maybe calling `write_history_file` in `start_scene` in case they forgot to call `end_scene`?

python_api/machine_common_sense/history_writer.py
```diff
@@ -0,0 +1,70 @@
...
+            if 'target' in history.output.goal.metadata.keys():
+                del history.output.goal.metadata['target']['image']
+            if 'target_1' in history.output.goal.metadata.keys():
+                del history.output.goal.metadata['target_1']['image']
+            if 'target_2' in history.output.goal.metadata.keys():
+                del history.output.goal.metadata['target_2']['image']
+        return history
+
+    # Add a new step to the array of history steps
+    def add_step(self, stepObj=Dict):
```
ThomasSchellenbergNextCentury, line 55: Nitpick: In python, variables are normally named in snake_case (`step_obj`).

deanwetherby: Now that Thomas pointed out that line, is the variable supposed to default to an empty dictionary or are you type hinting here?

brianpippin: It should be step_obj and I was just type hinting, but I can remove the "=Dict"

ThomasSchellenbergNextCentury: You can still use type hinting, just use a colon: `step_obj: Dict`

deanwetherby: Right. I prefer the code to leverage type hinting so, as Thomas says, just replace the equals with a colon.

brianpippin: Just updated this.

python_api/machine_common_sense/controller_ai2thor.py
brianpippin: I added something for this, let me know if this is what you were thinking.

deanwetherby: I'm not sure about this. You are assuming that a second scene is being called with the same instantiated controller. Evaluations were done with a single scene at a time so this would not have helped. There's also no way in the template to provide multiple scenes either.

brianpippin: I think if they don't call end scene and instantiate a new controller there is nothing they can do, I believe Thomas was under the impression some of their code was calling start_scene with the same controller when looping through the scenes, but perhaps their code doesn't behave this way.  If we don't think we need this I can remove it, but I am not sure there is any harm for checking if they do call start_scene with the same controller?

ThomasSchellenbergNextCentury: @deanwetherby  They SHOULD be using the same controller to run multiple scenes sequentially.  Creating a new controller loads the entire Unity app, so recreating the controller would shut down and reload the Unity app, which is not only unnecessary but also a major performance hit.  Using the `Initialize` action with the new scene's config data (which we do when they call `start_scene`) is the only thing needed to start a new scene.  Plus, they're told to use the same controller in our README instructions:  https://github.com/NextCenturyCorporation/MCS/tree/master/python_api#import

brianpippin: I do remember you mentioning that @ThomasSchellenbergNextCentury just let me know if this is what you were thinking then?

ThomasSchellenbergNextCentury: @brianpippin (Sorry, I didn't notice the update). Yes, that looks good. We write a separate history file for each scene, correct?

Also, another nitpick: python function block comments are usually positioned underneath the function `def` and use triple quotes.

```python
def my_function():
  """My function's block comment."""
  print('Hello world!')
```

brianpippin: Yes the file writes as scene_name+datetime+json so even if they run the same scene file multiple times, each instance would have a new timestamp and make a different file.

Python function comments are silly then.  But I will update.

ThomasSchellenbergNextCentury (approved):

python_api/machine_common_sense/controller_ai2thor.py
deanwetherby: @ThomasSchellenbergNextCentury I get you are OFFENDED by this notion but that's not how things worked and it's unlikely to work that way in an automation pipeline as in AWS Batch. I understand that there are inefficiencies with reloading the controller/app but it also provides flexibility in provisioning jobs with scenes. The template we provided as was code reviewed runs on a single scene file.

ThomasSchellenbergNextCentury: @deanwetherby (I'm assuming that emphasis was sarcasm) I guess that reinforces the takeaway from yesterday's meeting that we need to update our GitHub documentation with the latest running-in-batch instructions. Sorry I wasn't up-to-date with the latest routine.

This may be a stupid question, but are we sure that parallelizing the jobs, but requiring each job to create a new controller for each scene, is actually faster than running all the scenes with a single controller?  (Or having batches of, say, 100 scenes, and each batch only creates one controller.)

PR #188, opened 2020-10-08
MCS-384 and MCS-387

rartiss55:
- [MCS-384] Updating metadata tiers for eval 3 (new levels described in README)
- [MCS-387] Adding consistentColors property to randomize segmentation masks (associated ai2thor PR here: https://github.com/NextCenturyCorporation/ai2thor/pull/87)

[MCS-384]: https://nextcentury.atlassian.net/browse/MCS-384
[MCS-387]: https://nextcentury.atlassian.net/browse/MCS-387


machine_common_sense/API.md
```diff
@@ -24,7 +24,7 @@
 ## Controller
 
 
-### class machine_common_sense.controller.Controller(unity_app_file_path, debug=False, enable_noise=False, seed=None, size=None, depth_masks=False, object_masks=False, history_enabled=True)
+### class machine_common_sense.controller.Controller(unity_app_file_path, debug=False, enable_noise=False, seed=None, size=None, depth_masks=None, object_masks=None, history_enabled=True)
```
rartiss55, line 27: We had some discussion about this and we likely need to revisit how we handle different properties now that we have a config file. The issue specific to this ticket was: if depth_masks/object_masks are set, and we have the metadata level specified in the config, what decides whether or not we are generating masks?

For now, if the user specifies these properties explicitly, they'll take precedence over the metadata level in the config (so if I am on level1 metadata, which shouldn't give me object masks, but I set object_masks to true, I will get object masks). 

Since that's the strategy we're going with for now, IMO it felt odd to have these default to something explicitly when the controller is created and automatically override the config, so I set object_masks and depth_masks to default to None instead of False (now, we'll only give them a default of False if the user specified nothing AND there's no config).

I didn't make any changes to run_in_human_input_mode since I didn't think that mattered as much, so those properties there will still default to False.

machine_common_sense/controller.py
```diff
@@ -231,6 +231,31 @@ def _on_init(self, debug=False, enable_noise=False, seed=None,
         self.__history_writer = None
 
         self._config = self.read_config_file()
+        self._mode = (
+            self._config[self.CONFIG_METADATA_MODE]
+            if self.CONFIG_METADATA_MODE in self._config
+            else ''
+        )
+
+        # Order of preference for depth/object mask settings:
```
rartiss55, line 240: I feel like there is likely a better way to write this but the properties vs config thing gave me kind of a headache.

ThomasSchellenbergNextCentury: Seems fine to me.

machine_common_sense/controller.py
```diff
@@ -724,38 +737,24 @@ def restrict_object_output_metadata(self, object_output):
...
         if (
-            mode == self.CONFIG_METADATA_MODE_NO_VISION or
-            mode == self.CONFIG_METADATA_MODE_NONE
+            self._mode == self.CONFIG_METADATA_MODE_LEVEL_1 or
+            self._mode == self.CONFIG_METADATA_MODE_LEVEL_2
         ):
             step_output.camera_aspect_ratio = None
             step_output.camera_clipping_planes = None
             step_output.camera_field_of_view = None
             step_output.camera_height = None
```
ThomasSchellenbergNextCentury, line 757: Shoot, I completely forgot about this until now, but the final eval plan had the camera data included in ALL metadata tiers (and I think I forgot to update the ticket since it was a last minute change).

machine_common_sense/controller.py
```diff
@@ -143,13 +143,13 @@ class Controller():
...
-    CONFIG_METADATA_MODE_NO_VISION = 'no_vision'
-    # No metadata, except for the images and haptic/audio feedback
-    CONFIG_METADATA_MODE_NONE = 'none'
+    CONFIG_METADATA_MODE_ORACLE = 'oracle'
+    # No metadata, except for the images, depth masks, object masks,
+    # and haptic/audio feedback
+    CONFIG_METADATA_MODE_LEVEL_2 = 'level2'
+    # No metadata, except for the images, depth masks, and haptic/audio
+    # feedback
+    CONFIG_METADATA_MODE_LEVEL_1 = 'level1'
```
ThomasSchellenbergNextCentury, line 152: Not sure if we should rename "MODE" to "TIER" since that's the term that our subs started using.

machine_common_sense/controller.py
```diff
@@ -820,13 +819,7 @@ def retrieve_object_colors(self, scene_event):
...
 
     def retrieve_object_list(self, scene_event):
-        mode = (
-            self._config[self.CONFIG_METADATA_MODE]
-            if self.CONFIG_METADATA_MODE in self._config
-            else ''
-        )
-
-        if mode == self.CONFIG_METADATA_MODE_FULL:
+        if self._mode == self.CONFIG_METADATA_MODE_ORACLE:
```
ThomasSchellenbergNextCentury, line 822: I want to apologize, because I think I did a horrible job of brainstorming this ticket...

Currently, with Level 1 and Level 2 metadata, we're setting specific properties of the ObjectMetadata to None in `restrict_object_output_metadata`.  However, I think `retrieve_object_list` should just return an empty list (or None) with those metadata tiers.  The object's mask `color` isn't important (since the object mask is randomized), and our subs didn't care about "haptic feedback", so there's no reason to return any object output metadata at all.  Plus having a non-empty list would tell the performers exactly how many objects are visible in the current frame.  So I think we can just delete `restrict_object_output_metadata` and handle the tiers in this function.

(Remember to update the tiers in the README and `retrieve_structural_object_list` too.)

rartiss55: Its fine -- I knew I was missing something with this ticket too but it didn't immediately jump out at me when I looked over the eval plan and I probably got distracted by other issues. And yes, you're right about the objects, thank you for pointing that out -- we don't want to give them any inkling of how many objects there are. I also missed removing the object ids somehow (ugh) so that fixes that problem as well.

machine_common_sense/controller.py
rartiss55: Should be fixed.

machine_common_sense/controller.py
rartiss55: Updated to tier/metadata_tier

machine_common_sense/controller.py
rartiss55: Lists for objects + structural objects should be empty for levels 1 and 2, restrict_object_output_metadata is gone, and README is updated.

rartiss55: Also note that for now, if no config is set, I'm returning visible objects (thought that would be better for us when debugging and running locally).

rartiss55:
@ThomasSchellenbergNextCentury I think I addressed all the comments here. Let me know if there's anything else.

machine_common_sense/controller.py
```diff
@@ -837,18 +811,20 @@ def retrieve_object_list(self, scene_event):
...
-                self.retrieve_object_output(
-                    object_metadata, self.retrieve_object_colors(scene_event)
-                )
-                for object_metadata in scene_event.metadata['objects']
-                if object_metadata['visibleInCamera'] or
-                object_metadata['isPickedUp']
-            ],
-            key=lambda x: x.uuid
-        )
+        else:
```
ThomasSchellenbergNextCentury, line 814: Technically you don't need the `else` here, but that's a minor nitpick.

README.md
```diff
@@ -161,10 +161,9 @@ To use an MCS configuration file, set the `MCS_CONFIG_FILE_PATH` environment var
 
 The `metadata` property describes what metadata will be returned by the MCS Python library. The `metadata` property is available so that users can run baseline or ablation studies during training. It can be set to one of the following strings:
 
-- `full`: Returns the metadata for all the objects in the scene, including visible, held, and hidden objects.
-- `no_navigation`: Does not return the position or rotation of the player or the objects in the scene.
-- `no_vision`: Does not return the depth or object masks, camera properties, or object dimensions, directions, or distances.
-- `none`: Only returns the images (but not the masks), object IDs, and properties corresponding to the player themself (like head tilt or pose) and haptic feedback (like mass or materials of held objects).
+- `oracle`: Returns the metadata for all the objects in the scene, including visible, held, and hidden objects.
+- `level2`: Only returns the images (with depth masks AND object masks), camera info, and properties corresponding to the player themself (like head tilt or pose). No information about specific objects will be included.
```
ThomasSchellenbergNextCentury, line 165: We should probably say something about randomizing the object mask colors between actions in Level 2 versus keeping them consistent in Oracle.

ThomasSchellenbergNextCentury (approved):

machine_common_sense/controller.py
rartiss55: I think I do (there's an elif above that might've gone unnoticed).

README.md
rartiss55: Done!

machine_common_sense/controller.py
ThomasSchellenbergNextCentury: But they all `return` something, right?

brianpippin (approved):

PR #189, opened 2020-10-11
MCS-391 - Using pixel coordinates instead of directional vectors for actions

rartiss55:
Mostly renaming properties + documentation updates. 

Associated ai2thor PR here: https://github.com/NextCenturyCorporation/ai2thor/pull/88


machine_common_sense/controller.py
```diff
@@ -183,6 +181,11 @@ def __init__(self, unity_app_file_path, debug=False,
         self._on_init(debug, enable_noise, seed, depth_masks,
                       object_masks, history_enabled)
 
+    # Pixel coordinates are expected to start at the top left, but
+    # in Unity, (0,0) is the bottom left.
+    def _convert_y_image_coord_for_unity(self, y_coord):
```
rartiss55, line 186: I think most people expect (0,0) for pixels to be the top left (and not bottom left like it is in Unity), so figured we'd need to do this.

deanwetherby: This _may_ have an off-by-one issue. If the screen is 640x480, the screen height is 480 with zero-based indexing of those values (0 -> 479). Then if the y_coord to convert is at index 479, this will convert to 1 rather than 0. Likewise, if the y_coord is 0, this will convert to 480 and be out of bounds if used to index the image values.

ThomasSchellenbergNextCentury: Regardless of choosing top-left or bottom-left, we should state that choice with each ImageCoords parameter in the Action Enum so it appears in our API.

rartiss55: I'll repeat my standup comment here just in case: Looked at this again, and since (0,0) is actually the top left of the screen (and not the center of that first pixel), I made it so our defaults are (0,0) if nothing is entered (instead of -1,-1) and values greater than that are required in order for anything to happen. Note also that we shouldn't get any out of bounds issues, since Unity can make sense of off screen pixels and will likely throw a different error (object not visible, not interactable, etc).

Also, I added a note about choosing the top-left in the documentation/values greater than (0,0) being required to the docs. If people prefer the bottom-left being (0,0) instead (doubtful but who knows) we can always change that later.

ThomasSchellenbergNextCentury (approved):

machine_common_sense/controller.py
rartiss55: Ran out of time but since we started talking about documentation: MCS-357 to update the docs to say we no longer use object ids will need to be done before the beta release.

brianpippin (approved):

PR #190, opened 2020-10-12
MCS-350 Normalizing movement

deanwetherby:
Removed 'amount' from movement actions (wasd) and normalized to 0.25. Anything smaller wouldn't register for some reason (grid size?).

Replaced RotateLook(rotation,horizon) with RotateLeft, RotateRight, LookUp, and LookDown. Each rotation happens in 10 degree increments.

There will be a corresponding ai2thor PR in the same vane.


brianpippin (approved):

machine_common_sense/action.py
```diff
@@ -471,26 +444,29 @@ class Action(Enum):
...
-        "current view. (rotation=float, horizon=float)"
+    LOOK_UP = (
+        "LookUp",
+        "i",
+        "Rotate your view up."
     )
     """
-    Rotate your viewport left/right and/or up/down based on your current
-    viewport.
+    Rotate your viewport up.
```
ThomasSchellenbergNextCentury, line 453: We should state the exact number of degrees for each Look and Rotate action in the docs here.

machine_common_sense/controller.py
```diff
@@ -21,7 +21,7 @@
 MAX_REACH_DISTANCE = 1.0
 
 # How far the player can move with a single step.
-MAX_MOVE_DISTANCE = 0.5
+MAX_MOVE_DISTANCE = 0.25
```
ThomasSchellenbergNextCentury, line 24: We should remove the "MAX_" from this variable's name since it's just a static value now.

ThomasSchellenbergNextCentury, line 24: @deanwetherby Also, I'm pretty sure we had talked about setting the move distance to something like `0.05`, because `0.25` is just too far.

ThomasSchellenbergNextCentury:
@deanwetherby I see that you set the Look and Rotate degrees to `10`.  Should we have a discussion with our subs as to what the correct number should be?

PR #191, opened 2020-10-13
MCS-411 Splitting action step into step and prediction

deanwetherby:
Moving VOE prediction into a separate function. The prediction function appends attributes to the step's history item before adding it to the history writer. The current step's history item is added to the history writer on the following step or when end scene is called.

```
#loop
output = controller.step(action)
# calculate voe prediction
controller.make_step_prediction(confidence=1.0, ...)
#end loop
controller.end_scene()
```

A user of the package isn't forced to call make_step_prediction and will still get scene history. The prediction values (confidence etc) will be null.
```
#loop
output = controller.step(action)
#end loop
controller.end_scene()

### DIFF
--- a/machine_common_sense/controller.py
+++ b/machine_common_sense/controller.py
@@ -229,6 +229,7 @@ class Controller():
         self.__step_number = 0
         self.__s3_client = None
         self.__history_writer = None
+        self.__history_item = None
 
         self._config = self.read_config_file()
 
@@ -274,6 +275,7 @@ class Controller():
             with violation-of-expectation or classification goals.
             Is not required for other goals. (default None)
         """
+        self.__history_writer.add_step(self.__history_item)
         self.__history_writer.write_history_file(choice, confidence)
 
     def start_scene(self, config_data):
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: machine_common_sense/controller.py [hunk 1]
```diff
@@ -536,6 +511,9 @@ def step(self, action: str, choice: str = None,
             "last_step" of this scene.
         """
 
+        if self.__step_number > 0:
+            self.__history_writer.add_step(self.__history_item)
+
```
brianpippin, line 516: If I am reading all of this correct, whenever they call step or end scene, you are writing the step information for the previous step to the history file at that point?  Instead of writing it when they call step to account for the prediction function?

brianpippin (approved):

machine_common_sense/controller.py
deanwetherby: That is correct although the actual file writing doesn't happen until the end (as you know). I was going to do it another way but the add_step function converts the class into a dict and puts on a list which makes it tougher to manipulate with the prediction values after the fact. This was the cleanest way I could think do it for now.

brianpippin: Makes sense, just wanted to make sure I understood it fully.

PR #193, opened 2020-10-20
MCS-417 Set snapToGrid to False

ThomasSchellenbergNextCentury:
Fixes the movement bug that Falk reported today on Slack.

We will also need the corresponding Unity fix:
https://github.com/NextCenturyCorporation/ai2thor/pull/92

Also added some more terminal output for human input mode.

### DIFF
--- a/machine_common_sense/controller.py
+++ b/machine_common_sense/controller.py
@@ -332,6 +332,18 @@ class Controller():
                               'skip_preview_phase' in config_data['goal']
                               else False)
 
+        if self.__debug_to_terminal:
+            if config_data['name']:
+                print("STARTING NEW SCENE: " + config_data['name'])
+            else:
+                print("STARTING NEW SCENE")
+            if self._metadata_tier:
+                print("METADATA TIER: " + self._metadata_tier)
+            else:
+                print("METADATA TIER: DEFAULT (NOT CONFIGURED)")
+            print("STEP: 0")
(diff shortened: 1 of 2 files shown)


deanwetherby (approved):

machine_common_sense/controller.py
```diff
@@ -1055,6 +1073,7 @@ def wrap_step(self, **kwargs):
             logs=True,
             renderDepthImage=self.__depth_masks,
             renderObjectImage=self.__object_masks,
+            snapToGrid=False,
```
brianpippin, line 1076: I'm pretty sure we don't ever want snapToGrid, but is there an instance where we do, where perhaps this should be a config?  Or is that just a future down the road thought that probably won't ever be needed?

brianpippin (approved):

machine_common_sense/controller.py
ThomasSchellenbergNextCentury: Good question for the team.  Personally, I don't think so?

deanwetherby: I can't imagine needing snap-to-grid. If we ever do, we talk about exposing it via config then.

brianpippin: I didn't think we would, just wanted to bring it up to make sure.

rartiss55 (approved):

PR #196, opened 2020-10-26
Make sure image data is an array and not a string (MCS-405).

rartiss55:
It looks like we are saving most of our images as PIL.Images, but rereading the original Slack comment and looking at the code, it looked like the issue here was actually that we were saving the image data specifically for target object data as a string -- Falk mentioned that a NumPy array would be acceptable so that's what I went with. 

This change wasn’t hard, but I needed to get the image generator working (or in my case, at least get around current issues well enough to generate a scene with target objects). 

In addition, I realized that this change might be more of a backwards compatibility thing. Our target objects are always going to be the same anyway for Eval 3, and I’m not even sure we’re going to need to save the information for the target objects the same way we used to going forward.

Also wanted to verify the debug output file size -- looks like a .5 MB increase for this change (~~I did notice the depth mask update bumped up my MCS output file for this to about 8-9 MB, so just something to keep in mind~~ Nevermind this was because I was using an old unity app build from before we removed substeps).

### DIFF
--- a/machine_common_sense/controller.py
+++ b/machine_common_sense/controller.py
@@ -9,6 +9,7 @@ import random
 import sys
 import yaml
 import PIL
+import ast
 from typing import Dict, List
 
 import ai2thor.controller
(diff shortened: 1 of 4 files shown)


machine_common_sense/controller.py
```diff
@@ -733,6 +734,32 @@ def restrict_goal_output_metadata(self, goal_output):
                 'image' in goal_output.metadata['target_2']
             ):
                 goal_output.metadata['target_2']['image'] = None
+        else:
+            # need to convert goal image data from string to array
```
brianpippin, line 738: could you condense this similar to something like this (could potentially eliminate some of the duplicate code and make it easier to add targets in the future?:  (also this might not make sense feel free to ignore the suggestion)

target_str_array = ["target", "target_1", "target_2"]
for target_str in target_str_array:
      if (
                target_str' in goal_output.metadata and
                'image' in goal_output.metadata[target_str] and
                isinstance(goal_output.metadata[target_str]['image'], str)
            ):
                image_list_string = goal_output.metadata[target_str]['image']
                goal_output.metadata[target_str]['image'] = numpy.array(
                    ast.literal_eval(image_list_string)).tolist()

ThomasSchellenbergNextCentury:
@rartiss55 Can you also please verify that the `target.image`, `target_1.image`, and `target_2.image` documentation in `goal_metadata.py` is correct? https://github.com/NextCenturyCorporation/MCS/blob/master/machine_common_sense/goal_metadata.py#L134

machine_common_sense/controller.py
rartiss55: This was a good suggestion! I went ahead and consolidated things for this + the previous code above it in restrict_goal_output_metadata()

rartiss55:
> @rartiss55 Can you also please verify that the `target.image`, `target_1.image`, and `target_2.image` documentation in `goal_metadata.py` is correct? https://github.com/NextCenturyCorporation/MCS/blob/master/machine_common_sense/goal_metadata.py#L134

@ThomasSchellenbergNextCentury Yes, its a list of lists of lists.

brianpippin (approved):

ThomasSchellenbergNextCentury (approved):

deanwetherby:
Shouldn't this be a list of np.array? not a list of list of list of integers?

rartiss55:
> Shouldn't this be a list of np.array? not a list of list of list of integers?

@deanwetherby updated

PR #197, opened 2020-10-26
MCS-415: Add code that calculates a reward penalty based on the number of steps

brianpippin:


rartiss55:
Code looks good -- I did notice some flake8 warnings though if you wouldn't mind fixing those (only about 8 or so).

brianpippin:
> Code looks good -- I did notice some flake8 warnings though if you wouldn't mind fixing those (only about 8 or so).

Is there some code we run to fix these?  It seems super tedious to have to fix these things that are 10 characters over the line limits, 79 characters is a very small number of characters as well.  Or do we manually correct these every time?

rartiss55:
> > Code looks good -- I did notice some flake8 warnings though if you wouldn't mind fixing those (only about 8 or so).
> 
> Is there some code we run to fix these? It seems super tedious to have to fix these things that are 10 characters over the line limits, 79 characters is a very small number of characters as well. Or do we manually correct these every time?

Sometimes I have to fix stuff myself, but in vscode/settings.json for the project, its specified to run the autopep8 formatter on save, which should be catching some of this stuff as you go/fixing it for you. Not sure why it isn't working (perhaps autopep8 isn't installed or you're missing a plugin?)

rartiss55:
Had to double check the docs but to just command line fix things, after pip installing autopep8, you could just do:
`autopep8 --in-place --aggressive --recursive <path-to-MCS>`

rartiss55 (approved):

deanwetherby (approved):

ThomasSchellenbergNextCentury (approved):

PR #199, opened 2020-10-28
Update for setting config path + metadata properties

rartiss55:
Can't tackle all the config changes for MCS-410, but this should address what we need for Eval 3:

- Moving config path environment variable check to controller initialization to help ensure it is set the way we want per run
- Also adding a MCS_METADATA_LEVEL override to initialization. This will be prioritized over whatever level is in the configuration file. 

I'm hoping this is good enough, but again, may need to be tweaked a bit later since we haven't fully nailed down what the automation pipeline will look like.

### DIFF
--- a/machine_common_sense/controller.py
+++ b/machine_common_sense/controller.py
@@ -234,11 +234,18 @@ class Controller():
         self.__history_item = None
 
         self._config = self.read_config_file()
-        self._metadata_tier = (
-            self._config[self.CONFIG_METADATA_TIER]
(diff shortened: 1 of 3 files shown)


brianpippin (approved):
Is there any way to add a new test for the config file load or the metadata tier?

rartiss55:
> Is there any way to add a new test for the config file load or the metadata tier?

@brianpippin New tests for these added in test_env_variables.py

ThomasSchellenbergNextCentury (approved):

PR #200, opened 2020-10-29
Fixed bug with history_enabled create_controller argument. Disabled history on developer run scripts.

ThomasSchellenbergNextCentury:


machine_common_sense/controller.py
```diff
@@ -327,10 +327,11 @@ def start_scene(self, config_data):
...
 
-        if self.__history_writer is None:
-            self.__history_writer = HistoryWriter(config_data)
-        else:
-            self.__history_writer.check_file_written()
+        if self.__history_enabled:
+            if self.__history_writer is None:
+                self.__history_writer = HistoryWriter(config_data)
+            else:
+                self.__history_writer.check_file_written()
```
brianpippin, line 334: I think if we wanted to fix that one issue you chatted me about here as well, to reinit the file each time, you could add:  self.__history_writer = HistoryWriter(config_data) on line 335 in the else block and it would fix that issue?

ThomasSchellenbergNextCentury: Yes, I think that would work. Do you want me to do that now? Like I said in the chat, I wasn't sure if that behavior was intentional or not.

brianpippin: I don't think we want all the history in one file if running more scenes, each scene should have it's own history file, so I think it was a good catch, and that would fix it, I would add it to this ticket.

ThomasSchellenbergNextCentury: Done!

brianpippin (approved):

rartiss55 (approved):

PR #203, opened 2020-11-01
MCS-374 Threaded video writer

deanwetherby:
Added the ability to capture Unity images renders as video frames. Rather than saving individual images to disk for each step taken by the agents, we queue the in-memory frame in the video recorder. The video recorder uses a separate thread to dump those frames to the video file. Uploads to s3 to support evaluation review occur when end_scene is called and includes the various videos as well as the history file. This eliminates the need for additional post-processing to convert the images to video for evaluator consumption and reduces storage needs in our s3 bucket.

Added opencv-python as a dependency for the video writing capability.

Added an evaluation flag to the configuration file. When 'true', the video recorders are created/used and s3 uploads occur. 

mcs.config
```
evaluation: true
metadata: oracle
s3_bucket: evaluation-images
s3_folder: eval-3
...
```

I did notice that turning off depth_map and object_masks in the controller overrides the metadata level from the config. We need to take a closer look at this to support evaluations. This is part of the reason we need to get to a single place for configuration.

Reviewers: Note that there are a few non-functional WIP commits as part of this development process.


machine_common_sense/controller.py
```diff
@@ -36,6 +35,8 @@
 from .reward import Reward
 from .scene_history import SceneHistory
 from .step_metadata import StepMetadata
+from .recorder import VideoRecorder  # , HistoryRecorder
```
ThomasSchellenbergNextCentury, line 38: Remove unused code or add a comment here.

machine_common_sense/controller.py
```diff
@@ -91,6 +92,9 @@ class Controller():
     # really matter because we set continuous to True in
     # the step input.)
     GRID_SIZE = 0.1
+    # Is there any way to get FPS from ai2thor?
+    # 20 FPS relies on the assumption that Time.deltaTime is 0.05 in Unity
+    FPS_FRAME_RATE = 20
```
ThomasSchellenbergNextCentury, line 97: It's currently set in MCSController, using [PHYSICS_SIMULATION_STEPS](https://github.com/NextCenturyCorporation/ai2thor/blob/master/unity/Assets/Scripts/MCSController.cs#L23) and the [SimulatePhysicsOnce](https://github.com/NextCenturyCorporation/ai2thor/blob/master/unity/Assets/Scripts/MCSController.cs#L547-L549) method. We could create a new ticket to make the frame rate configurable or at least return it in the Unity output.

machine_common_sense/controller.py
```diff
@@ -286,11 +292,52 @@ def _on_init(self, debug=False, enable_noise=False, seed=None,
...
-                self.__s3_client = boto3.client('s3')
+    def _create_video_recorders(self):
+        '''Create video recorders used to capture evaluation scenes for review
+        '''
+        print(self.__output_folder)
+        output_folder = pathlib.Path(self.__output_folder)
+        team = self._config.get(self.CONFIG_TEAM, '')
+        scene = self.__scene_configuration.get(
+            'name', '').replace('json', '')
+        placeholder = 'placeholder'
```
ThomasSchellenbergNextCentury, line 303: Should 'placeholder', 'visual', 'heatmap', 'depth', and 'segmentation' all be defined as global static variables?

machine_common_sense/controller.py
```diff
@@ -286,11 +292,52 @@ def _on_init(self, debug=False, enable_noise=False, seed=None,
...
+        timestamp = self.generate_time()
+        basename_template = '_'.join(
+            [team, scene, placeholder, timestamp]) + '.mp4'
+
+        visual_video_filename = basename_template.replace(
+            placeholder, 'visual')
+        self.__image_recorder = VideoRecorder(
+            vid_path=output_folder / visual_video_filename,
```
ThomasSchellenbergNextCentury, line 311: I haven't seen `/` used with strings before. Is this a concatenation?

machine_common_sense/recorder.py
```diff
@@ -0,0 +1,102 @@
...
+            height (int): video height dimension
+            fps (int): video frame rate per second
+            fourcc (str): opencv fourcc value
+            timeout (float): thread sleep timeout
+
+        Returns:
+            None
+        '''
+
+        self.Q = None
```
ThomasSchellenbergNextCentury, line 35: Is this variable name a well-known standard? If not, I'd recommend renaming this variable so its name A. is not one letter and B. does not start with a capital letter.

machine_common_sense/recorder.py
```diff
@@ -0,0 +1,102 @@
...
+                 fourcc: str = 'mp4v',
+                 timeout: float = 1.0):
+        '''Create the video recorder.
+
+        Args:
+            vid_path (pathlib.path): the output video file path
+            width (int): video width dimension
+            height (int): video height dimension
+            fps (int): video frame rate per second
+            fourcc (str): opencv fourcc value
```
ThomasSchellenbergNextCentury, line 28: What is a `fourcc`? A longer description may be useful.

machine_common_sense/controller.py
deanwetherby: pathlib.Path has a '/' operator to concatenate directories and files. It's so much nicer in my opinion than os.path.

machine_common_sense/recorder.py
deanwetherby: It's an opencv term essentially meaning the codec.

"FourCC is a 4-byte code used to specify the video codec. The list of available codes can be found in fourcc.org. It is platform dependent. The following codecs work fine for me.

In Fedora: DIVX, XVID, MJPG, X264, WMV1, WMV2. (XVID is more preferable. MJPG results in high size video. X264 gives very small size video)
In Windows: DIVX (More to be tested and added)
In OSX: MJPG (.mp4), DIVX (.avi), X264 (.mkv)."

machine_common_sense/controller.py
```diff
@@ -286,11 +292,52 @@ def _on_init(self, debug=False, enable_noise=False, seed=None,
...
 
-        if self.CONFIG_SAVE_IMAGES_TO_S3_BUCKET in self._config:
-            if 'boto3' not in sys.modules:
-                import boto3
-                from botocore.exceptions import ClientError  # noqa: F401
-                self.__s3_client = boto3.client('s3')
+    def _create_video_recorders(self):
+        '''Create video recorders used to capture evaluation scenes for review
+        '''
+        print(self.__output_folder)
```
rartiss55, line 298: Did this print statement get left here by mistake?

machine_common_sense/controller.py
```diff
@@ -286,11 +292,52 @@ def _on_init(self, debug=False, enable_noise=False, seed=None,
...
+
+        heatmap_video_filename = basename_template.replace(
+            placeholder, 'heatmap')
+        self.__heatmap_recorder = VideoRecorder(
+            vid_path=output_folder / heatmap_video_filename,
+            width=self.__screen_width,
+            height=self.__screen_height,
+            fps=self.FPS_FRAME_RATE)
+
+        if self.__depth_masks:
```
rartiss55, line 324: This was recently changed to __depth_maps, right? Just want to check/make sure it gets updated when everything is merged.

requirements.txt
```diff
@@ -15,6 +15,7 @@ imagesize==1.2.0
 itsdangerous==1.1.0
 Jinja2==2.11.2
 MarkupSafe==1.1.1
+opencv-python=4.4.0.44
```
rartiss55, line 18: Should have a '=='

machine_common_sense/controller.py
```diff
@@ -314,6 +361,52 @@ def end_scene(self, choice, confidence=1.0):
...
 
+        if self._config.get(self.CONFIG_EVALUATION, False):
+            # Create the uploader
+            self.__uploader = S3Uploader(
+                s3_bucket=self._config.get(self.CONFIG_S3_BUCKET, None)
+            )
+
+            folder_prefix = self._config.get(self.CONFIG_S3_FOLDER, None)
+
+            history_filename = pathlib.Path(
```
brianpippin, line 372: In the future should there be a function in the history writer that not only writes to disk but uploads it to S3 and that's contained in that module?

rartiss55:
We don't have to do this but, wondering if it would make life easier if the metadata tier was part of the various filenames?

rartiss55:
> I did notice that turning off depth_map and object_masks in the controller overrides the metadata level from the config. We need to take a closer look at this to support evaluations. This is part of the reason we need to get to a single place for configuration.
> 

We discussed this at some point and decided to let these parameters override the metadata levels for now if they were provided. If they’re not specified, it should just determine this based on the metadata tier. 

Additionally, we may want to strip out the depth_map and object_masks parameters entirely and just stick with deciding this via metadata tiers, or we can ignore those parameters when evaluation is “true”. I realized when thinking about the config that its still confusing to have both these parameters + metadata levels even if they’re all collected in the same configuration file (I’ll repost the question in the config ticket since it might’ve gotten lost in my initial long comment).

machine_common_sense/controller.py
deanwetherby: Yes it did.

requirements.txt
deanwetherby: And I need to add it to requirements.txt as was pointed out in the other PR.

machine_common_sense/controller.py
deanwetherby: I thought about that approach as well. We'll have to revisit this upon reflection.

machine_common_sense/controller.py
```diff
@@ -314,6 +367,52 @@ def end_scene(self, choice, confidence=1.0):
...
+            )
+
+            self.__heatmap_recorder.finish()
+            video_filename = self.__heatmap_recorder.path.name
+            self.__uploader.upload_video(
+                video_path=self.__heatmap_recorder.path,
+                s3_filename=folder_prefix + '/' + video_filename
+            )
+
+            if self.__depth_masks:
```
rartiss55, line 401: I think the other line got changed but this should be updated to "depth_maps" as well.

deanwetherby:
> We discussed this at some point
> its still confusing

yes

machine_common_sense/controller.py
ThomasSchellenbergNextCentury: I made this ticket: https://nextcentury.atlassian.net/browse/MCS-424

ThomasSchellenbergNextCentury:
@rartiss55 @deanwetherby I personally find it helpful to be able to override the defaults using the `create_controller` arguments since I switch back and forth between different levels of testing frequently so it's nice to just have everything defined in my run script. (However, I'm sure I could adapt my workflow if there's a strong argument against having the current behavior.)

machine_common_sense/controller.py
```diff
@@ -314,6 +367,52 @@ def end_scene(self, choice, confidence=1.0):
...
+            folder_prefix = self._config.get(self.CONFIG_S3_FOLDER, None)
+
+            history_filename = pathlib.Path(
+                self.__history_writer.scene_history_file).name
+            self.__uploader.upload_history(
+                history_path=self.__history_writer.scene_history_file,
+                s3_filename=(folder_prefix + '/' +
+                             self._config[self.CONFIG_TEAM] +
+                             '_' + history_filename)
+            )
```
ThomasSchellenbergNextCentury, line 385: Please wrap these lines with `if self.__history_enabled:`

machine_common_sense/controller.py
```diff
@@ -652,42 +759,7 @@ def make_step_prediction(self, choice: str = None,
...
-            self.__s3_client.upload_fileobj(
-                in_memory_file,
-                self._config[self.CONFIG_SAVE_IMAGES_TO_S3_BUCKET],
-                s3_file_name,
-                ExtraArgs={
-                    'ACL': 'public-read',
-                    'ContentType': 'image/png',
-                }
-            )
+            self.__heatmap_recorder.add(heatmap_img)
```
rartiss55, line 762: Missed this earlier -- I think we need a check here to see if we're in eval mode before trying to save to the recorder.

rartiss55 (approved):

ThomasSchellenbergNextCentury (approved):

PR #204, opened 2020-11-02
MCS-422: Write oracle level data to the history file

brianpippin:


machine_common_sense/controller.py
```diff
@@ -367,9 +367,11 @@ def start_scene(self, config_data):
             for file_path in file_list:
                 os.remove(file_path)
 
-        output = self.wrap_output(self._controller.step(
+        pre_restrict_output = self.wrap_output(self._controller.step(
             self.wrap_step(action='Initialize', sceneConfig=config_data)))
 
+        output = self.restrict_step_output_metadata(pre_restrict_output)
```
ThomasSchellenbergNextCentury, line 373: Since python classes are pass-by-reference, won't `pre_restrict_output` and `output` be the same object at this point? I think you need to call `copy.deepcopy` BEFORE calling `restrict_step_output_metadata`?

machine_common_sense/controller.py
```diff
@@ -725,34 +729,25 @@ def read_config_file(self):
...
-                # need to convert goal image data from string to array
-                if (
-                    target_name in goal_output.metadata and
-                    'image' in goal_output.metadata[target_name] and
-                    isinstance(goal_output.metadata[target_name]['image'], str)
-                ):
-                    image_list_string = goal_output.metadata[target_name]['image']  # noqa: E501
-                    goal_output.metadata[target_name]['image'] = numpy.array(
-                        ast.literal_eval(image_list_string)).tolist()
+                image_list_string = goal_output.metadata[target_name]['image']  # noqa: E501
```
ThomasSchellenbergNextCentury, line 742: I don't think this line needs the `# noqa: E501` part now that it's fewer than 80 characters long.

machine_common_sense/controller.py
ThomasSchellenbergNextCentury: Oh, I guess you don't really use the `pre_restrict_output` in this function anyway.  Never mind.

brianpippin: I did that as more of a semantic thing so the code was easier to read more than anything else, was my goal with that.  So it would be clear looking at the code what was going on.

machine_common_sense/controller.py
brianpippin: I will double check that.

rartiss55 (approved):

machine_common_sense/controller.py
ThomasSchellenbergNextCentury: Makes sense. For some reason when I first read it I thought it was part of the same function as the `pre_restrict_output` and `history_copy` below.

ThomasSchellenbergNextCentury (approved):

PR #205, opened 2020-11-02
MCS-375 Topdown plot

deanwetherby:
Generating a topdown plot for each frame when in evaluation mode as well as creating and uploading the video of those plots.

### DIFF
--- a/machine_common_sense/controller.py
+++ b/machine_common_sense/controller.py
@@ -30,6 +30,7 @@ PERFORMER_CAMERA_Y = 0.4625
 from .action import Action
 from .goal_metadata import GoalMetadata
 from .object_metadata import ObjectMetadata
+from .plotter import TopDownPlotter
 from .pose import Pose
 from .return_status import ReturnStatus
 from .reward import Reward
@@ -313,6 +314,14 @@ class Controller():
             height=self.__screen_height,
             fps=self.FPS_FRAME_RATE)
 
+        topdown_video_filename = basename_template.replace(
+            placeholder, 'topdown')
+        self.__topdown_recorder = VideoRecorder(
+            vid_path=output_folder / topdown_video_filename,
+            width=self.__screen_width,
+            height=self.__screen_height,
+            fps=self.FPS_FRAME_RATE)
+
         heatmap_video_filename = basename_template.replace(
             placeholder, 'heatmap')
         self.__heatmap_recorder = VideoRecorder(
@@ -321,7 +330,7 @@ class Controller():
             height=self.__screen_height,
             fps=self.FPS_FRAME_RATE)
 
-        if self.__depth_masks:
+        if self.__depth_maps:
             depth_video_filename = basename_template.replace(
                 placeholder, 'depth')
             self.__depth_recorder = VideoRecorder(
@@ -362,7 +371,6 @@ class Controller():
             self.__history_writer.write_history_file(choice, confidence)
 
         if self._config.get(self.CONFIG_EVALUATION, False):
-            # Create the uploader
             self.__uploader = S3Uploader(
                 s3_bucket=self._config.get(self.CONFIG_S3_BUCKET, None)
             )
@@ -376,6 +384,13 @@ class Controller():
                 s3_filename=(folder_prefix + '/' +
                              self._config[self.CONFIG_TEAM] +
                              '_' + history_filename)
+            )
+
+            self.__topdown_recorder.finish()
+            topdown_filename = self.__topdown_recorder.path.name
+            self.__uploader.upload_video(
+                video_path=self.__topdown_recorder.path,
+                s3_filename=folder_prefix + '/' + topdown_filename
             )
 
             self.__image_recorder.finish()
@@ -392,7 +407,7 @@ class Controller():
                 s3_filename=folder_prefix + '/' + video_filename
             )
 
-            if self.__depth_masks:
+            if self.__depth_maps:
                 self.__depth_recorder.finish()
                 video_filename = self.__depth_recorder.path.name
                 self.__uploader.upload_video(
@@ -462,6 +477,11 @@ class Controller():
                 os.remove(file_path)
 
         if self._config.get(self.CONFIG_EVALUATION, False):
+            team = self._config.get(self.CONFIG_TEAM, '')
+            scene = self.__scene_configuration.get(
+                'name', '').replace('json', '')
+            self.__plotter = TopDownPlotter(
+                team, scene, self.__screen_width, self.__screen_height)
             self._create_video_recorders()
 
         output = self.wrap_output(self._controller.step(
@@ -1046,6 +1066,8 @@ class Controller():
 
             if self._config.get(self.CONFIG_EVALUATION, False):
                 self.__image_recorder.add(scene_image)
+                self.__topdown_recorder.add(
+                    self.__plotter.plot(scene_event, self.__step_number))
 
             if self.__depth_maps:
                 # The Unity depth array (returned by Depth.shader) contains
--- a/machine_common_sense/plotter.py
+++ b/machine_common_sense/plotter.py
@@ -1,0 +1,92 @@
+import io
+import math
+import PIL
+
+import matplotlib.pyplot as plt
+from typing import Dict
+
+
+class TopDownPlotter():
+
+    HEADING_LENGTH = 0.4
+
+    def __init__(self, team: str, scene_name: str,
+                 plot_width: int, plot_height: int):
+        self._team = team
+        self._scene_name = scene_name
+        self._plot_width = plot_width
+        self._plot_height = plot_height
+
+    def plot(self, scene_event: Dict, step_number: int) -> PIL.Image.Image:
+
+        self._draw_agent(scene_event.metadata['agent'])
+
+        colors = scene_event.events[-1].object_id_to_color
+
+        self._draw_object_bounds(scene_event.metadata['objects'], colors)
+
+        # set scene extents
+        plt.xlim(-5, 5)
+        plt.ylim(-5, 5)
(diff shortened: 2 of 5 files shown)


ThomasSchellenbergNextCentury:
@deanwetherby I changed the PR base to branch MCS-374 so we aren't re-reviewing the same files from that PR. I hope that's OK.

machine_common_sense/controller.py
```diff
@@ -313,6 +314,14 @@ def _create_video_recorders(self):
             height=self.__screen_height,
             fps=self.FPS_FRAME_RATE)
 
+        topdown_video_filename = basename_template.replace(
+            placeholder, 'topdown')
+        self.__topdown_recorder = VideoRecorder(
+            vid_path=output_folder / topdown_video_filename,
+            width=self.__screen_width,
+            height=self.__screen_height,
+            fps=self.FPS_FRAME_RATE)
```
ThomasSchellenbergNextCentury, line 323: Maybe this idea is for a new ticket, but should we add a `create_controller` argument to enable the creation of top-down videos without requiring evaluation mode? Seems like they could be useful for debugging.

machine_common_sense/plotter.py
```diff
@@ -0,0 +1,92 @@
...
+        self._draw_agent(scene_event.metadata['agent'])
+
+        colors = scene_event.events[-1].object_id_to_color
+
+        self._draw_object_bounds(scene_event.metadata['objects'], colors)
+
+        # set scene extents
+        plt.xlim(-5, 5)
+        plt.ylim(-5, 5)
+        plt.text(5.05, -4.95, step_number)
```
ThomasSchellenbergNextCentury, line 31: Should the scene extents be global static variables?

machine_common_sense/plotter.py
```diff
@@ -0,0 +1,92 @@
...
+
+        # convert figure to PIL Image
+        fig = plt.gcf()
+        buf = io.BytesIO()
+        fig.savefig(buf)
+        buf.seek(0)
+        img = PIL.Image.open(buf)
+
+        # resize image to match screen dimensions
+        img = img.resize((self._plot_width, self._plot_height))
```
ThomasSchellenbergNextCentury, line 42: Do the top-down images NEED to be the same size/ratio as the frame images? We wouldn't want that if, for example, we reduced the frame image size to something much smaller like 64x64. If we made its size independent, we wouldn't have the stretching issue.

machine_common_sense/plotter.py
```diff
@@ -0,0 +1,92 @@
...
+        '''Plot the agent position and heading'''
+        agent_x = agent_metadata['position']['x']
+        agent_z = agent_metadata['position']['z']
+        rotation = agent_metadata['rotation']['y']
+
+        # draw agent position
+        circle = plt.Circle(
+            (agent_x, agent_z),
+            radius=0.2,
+            label='agent')
```
ThomasSchellenbergNextCentury, line 57: What color is the agent? Should we specify it here? Should it be greyscale?

machine_common_sense/plotter.py
```diff
@@ -0,0 +1,92 @@
...
+            color = colors.get(uuid, None)
+            bounds = obj.get('objectBounds', None)
+            if bounds is not None:
+                dimensions = bounds.get('objectBoundsCorners', None)
+
+                if color and all(color):
+                    rgb = tuple(
+                        [color[0] / 255, color[1] / 255, color[2] / 255])
+                else:
+                    rgb = tuple([1.0, 0.0, 0.0])
```
ThomasSchellenbergNextCentury, line 84: Are there any objects that aren't present in the color dict? If so, that may be an indication of a bigger issue.

machine_common_sense/controller.py
deanwetherby: I was thinking that for after Eval 3.

machine_common_sense/plotter.py
deanwetherby: Sure. I was worried about room dimension changes in the future and wondering if the better solution wouldn't be asking something else for the dimensions of the scene. Figured magic numbers were just as bad as a constant in that regard.

machine_common_sense/plotter.py
deanwetherby: When I make the video recorder threads. I make them all the screen height and width. This works for all the image types but got awkward for this as you see. Since I am making the video recorders on start_scene, I don't have the dimensions of the plot yet. The recorder could wait until the first recorder.add(PIL.Image) to make the writer. Hmm.

machine_common_sense/plotter.py
deanwetherby: I'm letting matplotlib decide with defaults but I did wonder if we should make it more obvious and consistent.

machine_common_sense/plotter.py
deanwetherby: I am getting some Nones for colors. If I understand correctly, these are for drawers and such.

ThomasSchellenbergNextCentury: Ah, OK, that makes sense. Should we render the drawers?

machine_common_sense/plotter.py
```diff
@@ -0,0 +1,92 @@
+import io
+import math
+import PIL
+
+import matplotlib.pyplot as plt
```
ThomasSchellenbergNextCentury, line 5: matplotlib needs to be in the requirements file.

machine_common_sense/plotter.py
```diff
@@ -0,0 +1,92 @@
...
+
+        colors = scene_event.events[-1].object_id_to_color
+
+        self._draw_object_bounds(scene_event.metadata['objects'], colors)
+
+        # set scene extents
+        plt.xlim(-5, 5)
+        plt.ylim(-5, 5)
+        plt.text(5.05, -4.95, step_number)
+        plt.title(f"{self._team} {self._scene_name}")
```
ThomasSchellenbergNextCentury, line 32: Maybe add the eval number to the title as well?

brianpippin:
We probably don't have time now, but I feel like plotter could use some tests

deanwetherby:
> We probably don't have time now, but I feel like plotter could use some tests

As big of a proponent of testing as I am, I'm not even sure what would be tested or how we would go about testing this.

machine_common_sense/plotter.py
deanwetherby: Eval number?

ThomasSchellenbergNextCentury: Like `Eval 3 OPICS Scene1234`

deanwetherby: Funny. I was thinking something else. Do we have that information anywhere or would this just be a hard-coded string for now?

machine_common_sense/plotter.py
deanwetherby: i don't think it's a big deal. This is really the first cut at the top-down view. I'm sure we will get plenty of feedback to improve for next time. The big point was to get it into the code base rather than having it as a post-processing step.

machine_common_sense/controller.py
ThomasSchellenbergNextCentury: I made this ticket for us: https://nextcentury.atlassian.net/browse/MCS-423

requirements.txt
```diff
@@ -15,6 +15,7 @@ imagesize==1.2.0
 itsdangerous==1.1.0
 Jinja2==2.11.2
 MarkupSafe==1.1.1
+matplotlib=3.3.2
```
rartiss55, line 18: Should also be '==' here

machine_common_sense/plotter.py
ThomasSchellenbergNextCentury: Hmm, I thought we had a specific variable in the config file for this already, but I guess we don't. It might make sense for us to add one at some point, but don't feel obligated to do it now.

machine_common_sense/plotter.py
ThomasSchellenbergNextCentury: If the TopDownPlotter's scene extents were global static variables, the Controller could reference those variables when creating the VideoRecorder.

ThomasSchellenbergNextCentury:
@deanwetherby Re: testing, your `draw_agent` and `draw_object_bounds` functions currently do two things: they convert the step's output metadata into arguments for the pyplot functions (radius/points, color, label, fill, etc.), and then call the pyplot functions. If you split those two things into separate functions, you could then write tests for the first part (converting metadata into pyplot function arguments).

machine_common_sense/plotter.py
deanwetherby: So matplotlib uses a dpi setting that makes this a little awkward to implement fully before code freeze. I think it is what it is at this point.

deanwetherby:
> @deanwetherby Re: testing, your `draw_agent` and `draw_object_bounds` functions currently do two things: they convert the step's output metadata into arguments for the pyplot functions (radius/points, color, label, fill, etc.), and then call the pyplot functions. If you split those two things into separate functions, you could then write tests for the first part (converting metadata into pyplot function arguments).

Tests will have to be forthcoming for the next development cycle. We're out of time for code freeze.

machine_common_sense/plotter.py
```diff
@@ -93,8 +94,10 @@ def _draw_object_bounds(self, objects: Dict, colors: Dict) -> None:
                     obj_clr = 'ivory'
 
                 if dimensions is not None:
-                    obj_pts = [[d['x'], d['z']] for d in dimensions]
-                    poly = plt.Polygon(obj_pts,
+                    obj_pts = [(d['x'], d['z']) for d in dimensions[:4]]
```
ThomasSchellenbergNextCentury, line 97: @deanwetherby Why is this code just using the first four points in `dimensions`? Maybe that's why the rolling cylinders in my example look like lines?

PR #207, opened 2020-11-05
MCS-427 - Add evaluation_name property to config

rartiss55:
Using this property in names for uploaded files. 

I also added metadata_tier to the filenames because I thought that was helpful, but I can take that out.

Evaluation name + metadata are first in the filenames, but I can change the ordering in case we want to keep the team name first.

Example file name: Eval3_level1_ra_playroom_visual_20201104-202325.mp4

### DIFF
--- a/machine_common_sense/controller.py
+++ b/machine_common_sense/controller.py
@@ -161,6 +161,7 @@ class Controller():
     CONFIG_S3_BUCKET = 's3_bucket'
     CONFIG_S3_FOLDER = 's3_folder'
     CONFIG_TEAM = 'team'
+    CONFIG_EVALUATION_NAME = 'evaluation_name'
 
     def __init__(self, unity_app_file_path, debug=False,
                  enable_noise=False, seed=None, size=None,
@@ -305,12 +306,14 @@ class Controller():
         '''Create video recorders used to capture evaluation scenes for review
         '''
         output_folder = pathlib.Path(self.__output_folder)
+        eval_name = self._config.get(self.CONFIG_EVALUATION_NAME, '')
         team = self._config.get(self.CONFIG_TEAM, '')
         scene = self.__scene_configuration.get(
             'name', '').replace('json', '')
         timestamp = self.generate_time()
-        basename_template = '_'.join(
-            [team, scene, self.PLACEHOLDER, timestamp]) + '.mp4'
+        basename_template = ('_'.join(
+            [eval_name, team, scene,
+             self.PLACEHOLDER, timestamp]) + '.mp4').replace(' ', '')
 
         visual_video_filename = basename_template.replace(
             self.PLACEHOLDER, self.VISUAL)
@@ -389,8 +392,9 @@ class Controller():
                 self.__uploader.upload_history(
                     history_path=self.__history_writer.scene_history_file,
                     s3_filename=(folder_prefix + '/' +
-                                 self._config[self.CONFIG_TEAM] +
-                                 '_' + history_filename)
+                                 self._config[self.CONFIG_EVALUATION_NAME] +
+                                 '_' + self._config[self.CONFIG_TEAM] +
+                                 '_' + history_filename).replace(' ', '')
                 )
 
             self.__topdown_recorder.finish()
--- a/machine_common_sense/controller.py
+++ b/machine_common_sense/controller.py
@@ -315,9 +315,9 @@ class Controller():
             scene_name = scene_name.rsplit('/', 1)[1]
 
         timestamp = self.generate_time()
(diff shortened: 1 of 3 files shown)


machine_common_sense/controller.py
```diff
@@ -389,8 +392,12 @@ def end_scene(self, choice, confidence=1.0):
                 self.__uploader.upload_history(
                     history_path=self.__history_writer.scene_history_file,
                     s3_filename=(folder_prefix + '/' +
-                                 self._config[self.CONFIG_TEAM] +
-                                 '_' + history_filename)
+                                 self._config.get(
```
brianpippin, line 395: You need to add every property you are using to build the history file name into the history file itself, especially if that is going to be used to retrieve other files.  It should go into the info section of the history file.  The goal was to not have to dissect the file name to pull some of these variables out.  So eval name, metadata tier, team, etc should go into that history info block.

machine_common_sense/controller.py
```diff
@@ -305,12 +306,14 @@ def _create_video_recorders(self):
...
         output_folder = pathlib.Path(self.__output_folder)
+        eval_name = self._config.get(self.CONFIG_EVALUATION_NAME, '')
         team = self._config.get(self.CONFIG_TEAM, '')
         scene = self.__scene_configuration.get(
             'name', '').replace('json', '')
         timestamp = self.generate_time()
-        basename_template = '_'.join(
-            [team, scene, self.PLACEHOLDER, timestamp]) + '.mp4'
+        basename_template = ('_'.join(
+            [eval_name, team, scene,
```
deanwetherby: The evaluation number is currently used as the s3 folder as a filename prefix. Do we need it in the filename as well?

For this partial config, we would have something like 'eval-3-test/eval3_mess_scene1_heatmap_time.mp4'.
s3_bucket = evaluation-images
s3_folder = eval-3-test

machine_common_sense/controller.py
```diff
@@ -312,7 +312,7 @@ def _create_video_recorders(self):
             'name', '').replace('json', '')
         timestamp = self.generate_time()
         basename_template = ('_'.join(
-            [eval_name, team, scene,
+            [eval_name, self._metadata_tier, team, scene,
```
deanwetherby, line 315: :+1: :+1:

machine_common_sense/controller.py
rartiss55: That's a good point -- I don't think it necessarily hurts to leave it in either just in case (and it might help better identify things locally), but it is redundant if we are on s3 and already had the correct folder specified. 

I can leave it in, take it out entirely (leaving in metadata tier though) or maybe just leave it in the info in the history file?

brianpippin: I think it should definitely be in the history file info section.

rartiss55: Okay, I left the name as is for now and added it to the history file.

machine_common_sense/controller.py
rartiss55: Thanks for pointing this out. I added anything I could think of that might be useful, so now the info block looks like this:

"info": {
    "evaluation_name": "Eval 3",
    "evaluation": true,
    "metadata": "level1",
    "team": "ra",
    "name": "playroom",
    "timestamp": "20201105-144006"
},

machine_common_sense/controller.py
rartiss55: Looking it over we might want to reconsider having both an evaluation_name and evaluation property, but I'll add that to the config ticket for later.

machine_common_sense/controller.py
rartiss55: @brianpippin I'll also mention that I'm stripping out spaces from anything for naming the files, but I'm leaving them in the info block (this is mostly applicable for the evaluation name field) -- mentioning it in case its relevant for anything on the front end or during ingest.

ThomasSchellenbergNextCentury:
Everything seems fine to me, but I'll let Brian and Dean take another look.

machine_common_sense/controller.py
brianpippin: I don't think we should be stripping out the spaces.  We need to be able to use the info to exactly build the file names to match ideally without having to do transformations or logic on the front end.  If we want to remove the spaces, we might want to add another field that says exactly what the file name builds to or something in the file so we can then use that to get all the other files.

rartiss55: @brianpippin Okay, if there are spaces they should be left in now.

brianpippin: I think this is fine for now.

brianpippin (approved):

PR #209, opened 2020-11-06
MCS-425 Refactored plotter class and added unit tests

deanwetherby:


deanwetherby:
Added wall and occluder structural objects to the top-down plotter. Floor was causing z-buffer issues that overwrote the other objects. 

obj_perm_topdown

shp_const_topdown

stc_topdown

agency_topdown

machine_common_sense/plotter.py
```diff
@@ -24,85 +48,163 @@ def __init__(self, team: str, scene_name: str,
...
+                obj_pts = [(pt['x'], pt['z']) for pt in obj.bounds]
+                polygon = geometry.MultiPoint(obj_pts).convex_hull
+                pts = polygon.exterior.coords
+                self._draw_object_bounds(obj, pts)
+
+    def _draw_structural_objects(self, objects: Dict) -> None:
+        for o in objects:
+            obj = self._create_object(o)
+            if obj.bounds is not None and (obj.uuid.startswith(
+                    'occluder') or obj.uuid.startswith('wall')):
```
ThomasSchellenbergNextCentury, line 129: This will need to stay in sync with the scene generator code (to ensure that we don't change the UUIDs of the occluders or walls), but I don't think that will be a big issue.

Also, if you don't want to show the occluders' poles (I'm not sure if anyone thinks they're distracting), you can change `'occluder'` to `'occluder_wall'`.

machine_common_sense/plotter.py
```diff
@@ -24,85 +48,163 @@ def __init__(self, team: str, scene_name: str,
...
+    def plot(self, scene_event: ai2thor.server.Event,
+             step_number: int) -> PIL.Image.Image:
 
-        colors = scene_event.events[-1].object_id_to_color
-
-        self._draw_object_bounds(scene_event.metadata['objects'], colors)
+        plt = self._initialize_plot(step_number=step_number)
+        self._draw_structural_objects(
+            scene_event.metadata.get(
+                'structuralObjects', None))
```
ThomasSchellenbergNextCentury, line 57: If you change this call to

```python
self._draw_objects([
  obj for obj in scene_event.metadata.get('structuralObjects', [])
  if (obj.uuid.startswith('occluder') or obj.uuid.startswith('wall'))
])
```

then you can reuse `_draw_objects` and don't have to have `_draw_structural_objects` at all.

machine_common_sense/plotter.py
```diff
@@ -24,85 +48,163 @@ def __init__(self, team: str, scene_name: str,
...
+        self._draw_objects(scene_event.metadata.get('objects', None))
+        self._draw_agent(scene_event.metadata.get('agent', None))
```
ThomasSchellenbergNextCentury, line 59: I think the default value for all of these `get` functions should be an empty array? I think `None` would cause errors in the draw functions?

machine_common_sense/plotter.py
```diff
@@ -24,85 +48,163 @@ def __init__(self, team: str, scene_name: str,
...
-        agent_x = agent_metadata['position']['x']
-        agent_z = agent_metadata['position']['z']
-        rotation = agent_metadata['rotation']['y']
-
-        # draw agent position
+        if agent_metadata is None:
+            return None
+        agent = self._create_agent(agent_metadata)
+        self._draw_agent_position(agent)
+        self._draw_agent_heading(agent)
```
ThomasSchellenbergNextCentury, line 92: Small nitpick, but, in all of these "agent" functions / variables, when you say "agent", I think of the agent in the agent scenes, and not the AI performer agent. It might be clearer to rename them (or maybe it's just me)? Maybe to something like "performer_agent"?

machine_common_sense/plotter.py
```diff
@@ -24,85 +48,163 @@ def __init__(self, team: str, scene_name: str,
...
+    def _convert_color(self, color: str) -> str:
+        '''Convert color string to xkcd string'''
+        # use default of black if no color present
+        if not color:
+            color = 'black'
+        # white color does not show up in plot but ivory does
+        if color == 'white':
+            color = 'ivory'
+        # prefix with xkcd string
+        return 'xkcd:' + color
```
ThomasSchellenbergNextCentury, line 210: All of the color strings that we're using are HTML color names, so this should be a good, robust solution.

ThomasSchellenbergNextCentury:
The GIFs are beautiful, the unit test suite looks great, and the code seems tidy overall. Plus, Koleen will be happy to see situations like the yellow cylinder rolling a little in your shape constancy example. Good job!

machine_common_sense/plotter.py
deanwetherby: I'm handling the None condition internally to those functions.

machine_common_sense/plotter.py
deanwetherby: I'm ok with the poles being rendered. There may be a better way than uuid to determine what to draw for the next evaluation.

machine_common_sense/plotter.py
deanwetherby: Maybe what I'll do is is combine the regular object list with the filtered structural object list and make one draw objects call.

machine_common_sense/plotter.py
deanwetherby: robot? We have a terminology problem for sure.

machine_common_sense/plotter.py
ThomasSchellenbergNextCentury: Ah, I must have missed that. Awesome. Thanks.

machine_common_sense/plotter.py
ThomasSchellenbergNextCentury: Robot is fine too.

brianpippin (approved):

ThomasSchellenbergNextCentury (approved):

PR #213, opened 2020-12-02
MCS-410 - Separate configuration handling from controller

rartiss55:
- moved all properties from controller initialization to config file - now just the config file path can be passed to the controller
- using configparser and .ini file (https://docs.python.org/3/library/configparser.html) - previous yaml config file will no longer work 
- created config_manager.py + tests
- there are environment variable overrides for the following properties: MCS_CONFIG_FILE_PATH, MCS_METADATA_LEVEL, MCS_DEBUG_MODE
- Updated descriptions in README.md for config file/properties (only put the properties that I thought would be of interest to performers -- full list is in DEV.md)
- Brief summary of other changes: renamed enable_noise to noise_enabled, split debug property into debug and debut_output, and removed depth/object masks (now using metadata tier to determine whether or not to generate them and there's a metadata 'none' tier to turn both off)


README.md
```diff
@@ -79,14 +79,13 @@ Example usage of the MCS library:
...
 
 # We will give you the Unity app file.
-controller = mcs.create_controller(unity_app_file_path, depth_maps=True,
-                                   object_masks=True)
+controller = mcs.create_controller(unity_app_file_path, config_file_path='./some-path/config.ini')
 
-# Either load the config data dict from an MCS config JSON file or create your own.
-# We will give you the training config JSON files and the format to make your own.
-config_data, status = mcs.load_config_json_file(config_json_file_path)
+# Either load the scene data dict from an MCS scene config JSON file or create your own.
```
rartiss55, line 84: Had to update references for config JSON to "scene json" or "scene config json" so it wouldn't be conflated with the .ini config file.

machine_common_sense/controller.py
```diff
@@ -147,28 +132,21 @@ class Controller():
     # No metadata, except for the images, depth masks, and haptic/audio
     # feedback
     CONFIG_METADATA_TIER_LEVEL_1 = 'level1'
-
-    CONFIG_EVALUATION = 'evaluation'
+    # No metadata, except for the images and haptic/audio
+    # feedback
+    CONFIG_METADATA_TIER_NONE = 'none'
```
rartiss55, line 137: Since I got rid of depth_maps and object_masks, added a 'none' tier in case we want to turn both off since there wasn't a way to do that with the previous levels. I don't think we need to cover the case where we would want object masks but not depth maps, but I wanted to call attention to that in case that assumption was incorrect.

README.md
```diff
@@ -127,49 +125,78 @@ for config_json_file_path in config_json_file_list:
...
 
 ### Config File Properties
 
+#### debug
+
+(boolean, optional)
+
+Whether to save MCS output debug files in this folder. Will default to `False`.
+
+#### debug_output
```
rartiss55, line 156: Added 'debug_output' since we had an option to set to debug to a boolean OR string ('file' or 'terminal') -- thought it'd be easier for type checking properties to have them be separate.

brianpippin (approved):
I think this all looks good.  It should definitely be an eval 4 merge and we should merge it early on.

README.md
```diff
@@ -127,49 +125,78 @@ for config_json_file_path in config_json_file_list:
...
 ```
 
 Run options:
-- `--debug`
-- `--depth_maps`
-- `--noise`
-- `--object_masks`
-- `--seed <python_random_seed>`
-- `--size <screen_width_450_or_more>`
+- `--config_file_path`
```
ThomasSchellenbergNextCentury, line 132: Should this be something like `-config_file_path <file_path>`?

README.md
```diff
@@ -127,49 +125,78 @@ for config_json_file_path in config_json_file_list:
...
-To use an MCS configuration file, set the `MCS_CONFIG_FILE_PATH` environment variable to the path of your MCS configuration file.
+To use an MCS configuration file, you can either pass in a file path via the `config_file_path` property in the create_controller() method, or set the `MCS_CONFIG_FILE_PATH` environment variable to the path of your MCS configuration file (note that the configuration must be an INI file -- see [sample_config.ini](./sample_config.ini) for an example).
 
 ### Config File Properties
 
+#### debug
+
+(boolean, optional)
+
+Whether to save MCS output debug files in this folder. Will default to `False`.
```
ThomasSchellenbergNextCentury, line 154: Might want to add that this option also prints debug output to the terminal.

machine_common_sense/DEV.md
```diff
@@ -95,6 +95,83 @@ First, install ffmpeg. Then (change the frame rate with the `-r` option):
 ffmpeg -r 3 -i frame_image_%d.png output.gif
 ```
 
+## Full List of Config Options
```
ThomasSchellenbergNextCentury, line 98: Do we need this info copied here if it's already in the main README?

machine_common_sense/config_manager.py
```diff
@@ -0,0 +1,183 @@
...
+    CONFIG_EVALUATION = 'evaluation'
+    CONFIG_EVALUATION_NAME = 'evaluation_name'
+    CONFIG_HISTORY_ENABLED = 'history_enabled'
+    CONFIG_METADATA_TIER = 'metadata'
+    CONFIG_NOISE_ENABLED = 'noise_enabled'
+    CONFIG_S3_BUCKET = 's3_bucket'
+    CONFIG_S3_FOLDER = 's3_folder'
+    CONFIG_SEED = 'seed'
+    CONFIG_SIZE = 'size'
+    CONFIG_TEAM = 'team'
```
ThomasSchellenbergNextCentury, line 27: Maybe a silly idea (or a separate ticket), but if we commented these properties and had the auto-doc read this file, we could auto-generate the config file setup docs rather than manually maintaining them (they'd be in API rather than in README but we could put a link to API in README).

machine_common_sense/config_manager.py
```diff
@@ -0,0 +1,183 @@
...
+    CONFIG_DEFAULT_SECTION = 'MCS'
+
+    CONFIG_AWS_ACCESS_KEY_ID = 'aws_access_key_id'
+    CONFIG_AWS_SECRET_ACCESS_KEY = 'aws_secret_access_key'
+    CONFIG_DEBUG = 'debug'
+    CONFIG_DEBUG_OUTPUT = 'debug_output'
+    CONFIG_EVALUATION = 'evaluation'
+    CONFIG_EVALUATION_NAME = 'evaluation_name'
+    CONFIG_HISTORY_ENABLED = 'history_enabled'
+    CONFIG_METADATA_TIER = 'metadata'
```
ThomasSchellenbergNextCentury, line 21: I know this is a little extra (maybe a separate ticket), but should metadata tier (and maybe debug output) actually be an enum? We have a cleaner way to validate the properties now with this new class.

machine_common_sense/controller.py
```diff
@@ -1070,11 +1012,9 @@ def retrieve_return_status(self, scene_event):
...
 
     def retrieve_structural_object_list(self, scene_event):
-        # Return structural for all tier levels, the restrict output function
-        # will then strip out the necessary metadata
-        if (self._metadata_tier == self.CONFIG_METADATA_TIER_LEVEL_1 or
-                self._metadata_tier == self.CONFIG_METADATA_TIER_LEVEL_2 or
-                self._metadata_tier == self.CONFIG_METADATA_TIER_ORACLE):
+        # Return structural object list for all tier levels, the restrict
+        # output function will then strip out the necessary metadata
+        if (self._metadata_tier != ''):
```
ThomasSchellenbergNextCentury, line 1017: I know this would be a little extra (maybe a separate ticket), but should metadata tier (and maybe debug output) actually be an enum? We have a cleaner way to validate the properties now with this new class. Then we wouldn't have to do things like `metadata_tier != ''`

machine_common_sense/config_manager.py
```diff
@@ -0,0 +1,183 @@
...
+
+            debug_to_terminal = (
+                self.is_debug() or self.get_debug_output() == 'terminal'
+            )
+
+            if debug_to_terminal is True:
+                print('MCS Config File Path: ' + self._config_file)
+                print('Read MCS Config File:')
+                print({section: dict(self._config[section])
+                       for section in self._config.sections()})
```
ThomasSchellenbergNextCentury, line 60: This is a little nitpicky, but I think all of the debug_to_terminal printing should be in the same class, so it's easier to find.

machine_common_sense/DEV.md
rartiss55: So even though TA1 can see everything regardless, I kind of wanted to separate the full list that we need for our purposes versus a more watered down version if that makes sense? That was kind of the issue I was running into towards the end/how to document everything (I'm open to suggestions) -- I ended up going with "abbreviated list in README vs full list in DEV"

machine_common_sense/config_manager.py
rartiss55: I like the idea but didn't do this for the reason listed above for the DEV file comment. If we decide it isn't that big a deal though we can go this route.

machine_common_sense/controller.py
rartiss55: I think this is a good idea -- I added a ticket for this (MCS-476). I also realized that not having a metadata tier specified accomplishes the same goal I had in mind for 'None' (having a way to not turn on depth_masks and object_masks), so not sure if I should leave that in or not (although I consider 'not specified' to be more useful for debugging for us since it includes visible objects).

machine_common_sense/DEV.md
ThomasSchellenbergNextCentury: Got it. Makes sense.

machine_common_sense/config_manager.py
ThomasSchellenbergNextCentury: That's fine. Another option is to just comment the public/TA1 properties, so only those appear in the API doc, and keep the more detailed info in the DEV doc, but it's not a big deal.

ThomasSchellenbergNextCentury (approved):
@rartiss55 @deanwetherby Anything else need to be done on this ticket? Should this be a part of version 0.3.7 that we're theoretically releasing tomorrow?

ThomasSchellenbergNextCentury:
@rartiss55 I'm getting some odd errors while running the unit tests. Any ideas? Does this test not work if I have my MCS_CONFIG_FILE_PATH set?

```
======================================================================
FAIL: test_init (tests.test_config_manager.Test_Config_Manager)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/tschellenberg/Development/darpa/MCS/tests/test_config_manager.py", line 25, in test_init
    self.assertEquals(
AssertionError: '/home/tschellenberg/Development/darpa/MCS/mcs_config.yaml' != './mcs_config.ini'
- /home/tschellenberg/Development/darpa/MCS/mcs_config.yaml
+ ./mcs_config.ini


======================================================================
FAIL: test_init_with_arg (tests.test_config_manager.Test_Config_Manager)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/tschellenberg/Development/darpa/MCS/tests/test_config_manager.py", line 33, in test_init_with_arg
    self.assertEquals(config_mngr._config_file, file_path)
AssertionError: '/home/tschellenberg/Development/darpa/MCS/mcs_config.yaml' != './arg-test.ini'
- /home/tschellenberg/Development/darpa/MCS/mcs_config.yaml
+ ./arg-test.ini
```

rartiss55:
> @rartiss55 I'm getting some odd errors while running the unit tests. Any ideas? Does this test not work if I have my MCS_CONFIG_FILE_PATH set?
> 
> ```
> ======================================================================
> FAIL: test_init (tests.test_config_manager.Test_Config_Manager)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/tschellenberg/Development/darpa/MCS/tests/test_config_manager.py", line 25, in test_init
>     self.assertEquals(
> AssertionError: '/home/tschellenberg/Development/darpa/MCS/mcs_config.yaml' != './mcs_config.ini'
> - /home/tschellenberg/Development/darpa/MCS/mcs_config.yaml
> + ./mcs_config.ini
> 
> 
> ======================================================================
> FAIL: test_init_with_arg (tests.test_config_manager.Test_Config_Manager)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/tschellenberg/Development/darpa/MCS/tests/test_config_manager.py", line 33, in test_init_with_arg
>     self.assertEquals(config_mngr._config_file, file_path)
> AssertionError: '/home/tschellenberg/Development/darpa/MCS/mcs_config.yaml' != './arg-test.ini'
> - /home/tschellenberg/Development/darpa/MCS/mcs_config.yaml
> + ./arg-test.ini
> ```

@ThomasSchellenbergNextCentury nice find! This should be fixed now, but let me know if you run into the same issue with a previously set environment variable.

PR #219, opened 2020-12-17
MCS-477 Read scripted object states from scene configuration files and return them in step output object metadata

ThomasSchellenbergNextCentury:
Depends on https://github.com/NextCenturyCorporation/MCS/pull/213

Read scripted object-specific `states` from the scene configuration files at each step and return a `state_list` in the ObjectMetadata.

Lets us show when a gravity support pole is "active" or "inactive", and is flexible in case we need to give new states to objects in future tasks.


machine_common_sense/controller.py
```diff
@@ -946,37 +946,60 @@ def retrieve_object_output(self, object_metadata, object_id_to_color):
...
+            distance_in_world=(object_metadata['distance']),
+            held=object_metadata['isPickedUp'],
+            mass=object_metadata['mass'],
+            material_list=material_list,
+            position=object_metadata['position'],
+            rotation=object_metadata['rotation'],
+            shape=object_metadata['shape'],
+            state_list=self.retrieve_object_states(
+                object_metadata['objectId']
+            ),
```
ThomasSchellenbergNextCentury, line 973: This is the new part of this specific diff block (the rest was just a change to indentation).

tests/test_controller.py
```diff
@@ -1130,10 +1132,32 @@ def test_retrieve_object_list(self):
...
+        self.assertEqual(actual[1].state_list, [])
         self.assertEqual(actual[1].texture_color_list, ['c2', 'c3'])
         self.assertEqual(actual[1].visible, True)
 
+    def test_retrieve_object_list_with_states(self):
+        self.controller.start_scene({
+            'name': 'test name',
+            'objects': [{
+                'id': 'testId1',
+                'states': [['a', 'b'], ['c', 'd']]
```
ThomasSchellenbergNextCentury, line 1144: This is what `states` will look like in the scene configuration files.

rartiss55:
@ThomasSchellenbergNextCentury Your code looks fine, but I think since I squashed and merged my branch that this was relying on, the commit history and diff is weird. I cherry picked your commit and opened another PR here: https://github.com/NextCenturyCorporation/MCS/pull/223

PR #224, opened 2021-01-22
MCS-423 Added config option to generate top-down videos in non-evaluation mode

ThomasSchellenbergNextCentury:
Also fixed a couple of minor related bugs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,12 @@ A seed for the Python random number generator (defaults to None).
 
 Desired screen width. If value given, it must be more than `450`. If none given, screen width will default to `600`.
 
+#### video_enabled
+
+(boolean, optional)
+
+Save videos of the RGB frames, depth masks, object instance segmentation masks (if returned in the output by the chosen metadata tier), 2D top-down scene views, and the heatmap images given to us in `make_step_prediction` by the AI performer.
+
 ### Using the Config File to Generate Scene Graphs or Maps
 
 1. Save your .ini MCS configuration file with:
--- a/machine_common_sense/controller.py
+++ b/machine_common_sense/controller.py
@@ -336,6 +336,15 @@ class Controller():
             self.__history_writer.add_step(self.__history_item)
             self.__history_writer.write_history_file(choice, confidence)
 
+        if self._config.is_evaluation() or self._config.is_video_enabled():
+            self.__topdown_recorder.finish()
+            self.__image_recorder.finish()
+            self.__heatmap_recorder.finish()
+            if self.__depth_maps:
+                self.__depth_recorder.finish()
+            if self.__object_masks:
+                self.__segmentation_recorder.finish()
+
         if self._config.is_evaluation():
             self.__uploader = S3Uploader(
                 s3_bucket=self._config.get_s3_bucket()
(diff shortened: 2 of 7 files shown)


machine_common_sense/plotter.py
```diff
@@ -67,15 +67,14 @@ def _find_plottable_objects(
...
         filtered_structural_objects = [
             obj for obj in structural_objects
-            if obj.get('objectId', '').startswith('occluder') or
-            obj.get('objectId', '').startswith('wall')
+            if not obj.get('objectId', '').startswith('ceiling') and not
+            obj.get('objectId', '').startswith('floor')
         ]
         objects = scene_event.metadata.get('objects', [])
-        return objects + filtered_structural_objects
+        return filtered_structural_objects + objects
```
ThomasSchellenbergNextCentury, line 77: I tested this PR on a scene containing an interior wall with an ID that didn't begin with "wall" and thus didn't appear in the top-down view so I made the change that you see here.

machine_common_sense/recorder.py
```diff
@@ -107,8 +107,11 @@ def finish(self) -> None:
             shutil.move(self._path, 'temp.mp4')
             os.system(
                 f'ffmpeg -loglevel quiet -i temp.mp4'
-                f' -vcodec libx264 {self._path}')
+                f' -vcodec libx264 -vf format=yuv420p {self._path}')
```
ThomasSchellenbergNextCentury, line 110: Allow videos to run on Macs.

machine_common_sense/recorder.py
```diff
@@ -107,8 +107,11 @@ def finish(self) -> None:
...
             os.remove('temp.mp4')
+        else:
+            # Remove the unused video file without any frames.
+            os.remove(self._path)
```
ThomasSchellenbergNextCentury, line 114: Fixed a bug in which the heatmap videos were being saved without any frames in them.

README.md
```diff
@@ -204,6 +204,12 @@ A seed for the Python random number generator (defaults to None).
 
 Desired screen width. If value given, it must be more than `450`. If none given, screen width will default to `600`.
 
+#### video_enabled
+
+(boolean, optional)
+
+Save videos of the RGB frames, depth masks, object instance segmentation masks (if returned in the output by the chosen metadata tier), 2D top-down scene views, and the heatmap images given to us in `make_step_prediction` by the AI performer.
```
ThomasSchellenbergNextCentury, line 211: I figured, for now, I would just have one config file option to make all types of videos, and we could split it out into multiple separate options later if it is ever needed.

machine_common_sense/controller.py
```diff
@@ -355,36 +364,32 @@ def end_scene(self, choice, confidence=1.0):
                                  '_' + history_filename)
                 )
 
-            self.__topdown_recorder.finish()
             topdown_filename = self.__topdown_recorder.path.name
             self.__uploader.upload_video(
                 video_path=self.__topdown_recorder.path,
```
brianpippin, line 369: I know this isn't your code, but I noticed it when reviewing, it looks like we set video_path a bunch of places but never use it, do we need it?

brianpippin (approved):
This looked good.  Posting comments for us to consider, might need to create additional tickets if thought to be a good idea. 
1 - Would it make sense to have config enabled flags for each video type or will you always want all of them
2 - I think at some point we want the video file names/paths to be in the history file, that way the UI can just use that URL instead of having to build configs and find the URL. 

 Sorry for all the additional thoughts.

machine_common_sense/controller.py
deanwetherby: video_path is a keyword argument to the upload_video method. So if I had a python function

```
def foo(a,b):
  print(f"{a} {b}")
```

I could call it using positional arguments

```
foo(1,"bob")
1 bob
```
or I could use the keyword arguments (which is safer IMHO)

```
foo(a=1,b="bob")
1 bob
foo(b="bob", a=1)
1 bob
```

Does that address your concern?

README.md
deanwetherby: We'll likely have a follow-on ticket to split out the configs to be more granular.

machine_common_sense/recorder.py
deanwetherby: It was a feature! :D

deanwetherby (approved):

machine_common_sense/controller.py
brianpippin: Oh I see, this was just my python noobness not recognizing what was going on.

PR #226, opened 2021-02-02
MCS-544: Script to run sample commands often given to TA2 by TA1

brianpippin:


scripts/run_scene_with_command_file.py
```diff
@@ -0,0 +1,93 @@
+import argparse
+import cmd
+import os
+import machine_common_sense as mcs
+
+commandList = []
+
+os.environ['MCS_CONFIG_FILE_PATH'] = "../mcs_config.yaml"
```
rartiss55, line 8: Looks good -- the only thing I noticed was that this is referencing the old config file format and should probably be updated since many of the arguments below passed along to the controller are a part of that file now.

brianpippin: @rartiss55 - I addressed this comment if you could give it a look.

rartiss55: Cool, thanks! Looks up to date now. Your recent comment on another PR made me realize I should double check flake8 -- looks like there's a few minor warnings that shouldn't be too hard to fix:
./scripts/run_scene_with_command_file.py:2:1: F401 'cmd' imported but unused
./scripts/run_scene_with_command_file.py:3:1: F401 'os' imported but unused
./scripts/run_scene_with_command_file.py:38:1: W293 blank line contains whitespace
./scripts/run_scene_with_command_file.py:68:1: W391 blank line at end of file

rartiss55 (approved):

PR #229, opened 2021-02-04
Added CPU image and updated GPU image to Ubuntu 20.04

fplk:


CPU_Container.dockerfile
```diff
@@ -0,0 +1,67 @@
...
+
+RUN wget https://github.com/NextCenturyCorporation/MCS/releases/download/${mcsversion}/MCS-AI2-THOR-Unity-App-v${mcsversion}.x86_64 && \
+    wget https://github.com/NextCenturyCorporation/MCS/releases/download/${mcsversion}/MCS-AI2-THOR-Unity-App-v${mcsversion}_Data.tar.gz && \
+    python3 -m pip install git+https://github.com/NextCenturyCorporation/MCS@${mcsversion}#egg=machine_common_sense && \
+    tar -xzvf MCS-AI2-THOR-Unity-App-v${mcsversion}_Data.tar.gz && \
+    chmod a+x MCS-AI2-THOR-Unity-App-v${mcsversion}.x86_64 && \
+    rm MCS-AI2-THOR-Unity-App-v${mcsversion}_Data.tar.gz && echo "[MCS]\nmetadata: oracle" > /mcs_config_oracle.yaml && \
+    echo "[MCS]\nmetadata:  level1" > /mcs_config_level1.yaml && echo "[MCS]\nmetadata: level2" > /mcs_config_level2.yaml
+
+# CMD xvfb-run -s "-screen 0 1400x900x24" python3 /test.py
```
cedorman, line 67: It's not clear what this line should be used for, since I think it would conflict with the other X setup.  maybe add a line about when / how to use this.

README.md
```diff
@@ -1,18 +1,5 @@
 # MCS Python Package
 
-- [Installation](#installation)
```
cedorman, line 3: I find having these useful for navigating around.  It would be better to have them autogenerated, but if not, should be maintained manually.

README.md
```diff
@@ -335,6 +322,80 @@ DISPLAY=:0 python test.py
...
+```shell
+docker build -f CPU_Container.dockerfile -t mcs-playroom-cpu .
+```
+
+#### Run Image (bash)
+```shell
+docker run -it -p 5900:5900 -v ${PWD}/machine_common_sense/scenes:/input -v ${PWD}/scripts:/scripts mcs-playroom-cpu bash
+```
+
+#### Run with VNC
```
cedorman, line 374: This part is confusing to me.  Are all the commands below run within the docker container, outside of the container, or some of each?

cedorman (commented):
it builds and runs, which is good. I am confused about the different running options though, and what commands need to be run within the docker container vs outside of the container.

CPU_Container.dockerfile
fplk: Removed. It is described under "Run Fully Headless" in the Readme: If you want to just run MCS headless / in batch mode without any window showing its output (i.e. you only work on the maps encapsulated in the step metadata) you can just put virtual framebuffers under MCS to allow it to run. If you just tried executing it directly, it would blow up, whereas xvfb provides the bare minimum to allow it to run without having to run a full X or VNC server as well.

README.md
fplk: All inside the container except for command to connect to container via VNC which is run from host. Added clarification.

README.md
fplk: Very disturbing why this is missing in the branch. Not sure how this happened, but I made sure it is back in and that nothing else got compromised. Of course, I also added the containerization section. Thanks for paying attention - good catch.

fplk:
Could you take another look? Does this fulfill your requirements or is there anything else to change? Thanks in advance.

PR #247, opened 2021-02-10
Added preliminary JSON and MsgPack serialization and deserialization

fplk:


machine_common_sense/serialization.py
```diff
@@ -0,0 +1,233 @@
...
+    """
+    Serializes MCS Step Metadata as MsgPack, e.g.
+    serialized = msgpack.packb(output, default=ext_pack, strict_types=True)
+    """
+    if isinstance(x, StepMetadata):
+        return msgpack.ExtType(1, msgpack.packb([x.action_list, x.camera_aspect_ratio, x.camera_clipping_planes,
+                                                 x.camera_field_of_view, x.camera_height, x.depth_map_list, x.goal,
+                                                 x.head_tilt, x.image_list, x.object_list, x.object_mask_list, x.pose,
+                                                 x.position, x.return_status, x.reward, x.rotation, x.step_number,
+                                                 x.structural_object_list], default=ext_pack, strict_types=True))
```
ThomasSchellenbergNextCentury, line 35: Should we use something like `dir()` here, and for the GoalMetadata and ObjectMetadata classes, so we don't have to update this code each time we add a new attribute to the classes?

machine_common_sense/serialization.py
```diff
@@ -0,0 +1,233 @@
...
+    image.save(byte_io, format='PNG')
+    return byte_io.getvalue()
+
+
+def bytes_to_image(bytes_image):
+    """Converts bytes to PIL image."""
+    return Image.open(io.BytesIO(bytes_image))
+
+
+def ext_pack(x):
```
ThomasSchellenbergNextCentury, line 25: It may be better to move `ext_pack` and `ext_unpack` into their own specific class (within this file, or in a separate file), to more easily differentiate between the two methods of (de)serialization: MCS-MsgPack and MCS-JSON.

machine_common_sense/serialization.py
```diff
@@ -0,0 +1,233 @@
...
+                    object_raw['direction'],
+                    object_raw['distance'],
+                    object_raw['distance_in_steps'],
+                    object_raw['distance_in_world'],
+                    object_raw['held'],
+                    object_raw['mass'],
+                    object_raw['material_list'],
+                    object_raw['position'],
+                    object_raw['rotation'],
+                    object_raw['visible'])
```
ThomasSchellenbergNextCentury, line 180: To make this code more extensible (along with the code for GoalMetadata and StepMetadata in deserialize_step_metadata_from_json), could we create an empty ObjectMetadata, iterate over the class's attributes, and reassign them using the raw JSON data?

machine_common_sense/serialization.py
```diff
@@ -0,0 +1,233 @@
...
+                    'image_list': x.image_list,
+                    'object_list': x.object_list,
+                    'object_mask_list': x.object_mask_list,
+                    'pose': x.pose,
+                    'position': x.position,
+                    'return_status': x.return_status,
+                    'reward': x.reward,
+                    'rotation': x.rotation,
+                    'step_number': x.step_number,
+                    'structural_object_list': x.structural_object_list}
```
ThomasSchellenbergNextCentury, line 129: Could we just use `super().default()` here, and with GoalMetadata and ObjectMetadata below, so we don't have to list out all of the class's attributes?

ThomasSchellenbergNextCentury:
Falk, thanks for the PR. I hope you don't mind being our guinea pig for the TA1/TA2 collaborative development process :)

This may already be on your radar, but for any PRs:
1. Please run our Python auto-linter. We use the PEP8 standards for the most part. See our documentation here: https://github.com/NextCenturyCorporation/MCS/blob/development/machine_common_sense/DEV.md#linting
2. Please add unit tests. We use the Python `unitttest` library. Our tests are located in the `tests/` folder. See our documentation here: https://github.com/NextCenturyCorporation/MCS/blob/development/machine_common_sense/DEV.md#testing
3. If you introduce any new dependencies, please add them to our pip `requirements` file.

brianpippin:
@deanwetherby I think had some ideas about some of this and python packages, so we should make sure he chimes in before we merge this as well.

machine_common_sense/serialization.py
fplk: Good idea, but no (at least for now) - i) we use strict types in MsgPack and they are quite finicky in terms of parameter order ii) if fields change I want the serialization to break, so a human sanity checks serialization. I'm still afraid that otherwise we might flip values that match by type without noticing or so. We can continue automating this as we go, but I would recommend to first make sure the current version is rock solid.

machine_common_sense/serialization.py
fplk: Yeah, good point. This also means we probably want an interface or abstract class to keep the contract consistent. Give me an hour or so and I will push a refactored version.

machine_common_sense/serialization.py
fplk: Same concern with the order. The current MsgPack serialization relies on order. We could change that, but would have to rewrite it pretty much and worse - rewrite the Julia equivalent. I would recommend doing these generic things as a step 2 in a future enhancement unless you strongly object.

machine_common_sense/serialization.py
fplk: Technically, sure. The reason I didn't was that we need to make sure we overwrite the values with default values, i.e. distance, distance_in_steps, distance_in_world, held, mass and visible. So if I pull out values 1-by-1 I need to also check all values with default values are defined. And how do I generically find out which ones have default values (in case they change)? So it just seemed easier to set all and expect the test to fail if the signature changes. [An alternative would be repalcing the hard indexing with get() calls that set a default value of None or so.]

fplk:
@ThomasSchellenbergNextCentury Do the code changes I just pushed address your concerns? As aforementioned in response to your comments I would rather hold off with generic enhancements for now, since serialization can depend a lot on order, esp. for the flags we set, so I would prefer doing this iteratively. The class refactoring should be done, simple tests are added and my code should always be PEP8 compliant. Please let me know if there is anything you need to approve this.

machine_common_sense/serialization.py
ThomasSchellenbergNextCentury: Hmm. I better understand the challenge here now. However, if this code becomes part of the TA2 repository, then it becomes TA2's responsibility to ensure it's properly maintained and tested, since we don't want to release buggy code (especially as we anticipate more public scrutiny from organizations outside the program). TA2 must either spend its time to update this code, or notify you to do it, which may cause some lag time in the release process and may become problematic if there are changes to your team's personnel. On the other hand, we want to do what we (reasonably) can to help your team via this collaborative effort, and incorporating this code seems like it would help you a lot. I want to let @deanwetherby respond to this issue.

fplk: I think that being able to store MCS objects to file, to send them over the network as well as to use them in other languages is more a general MCS extensions that is useful for many users rather than a niche feature for a subteam. Especially since it works both in JSON as a lingua franca and MsgPack which is essentially JSON with efficient binary support. So I would see it more as a general contribution to MCS. One example: If you want to store full test runs from TA1 performers and can do so in 2 MB per step rather than 60 MB in a format that is easier and more efficient to parse, that should be pretty useful even for TA2 itself.

tests/test_serializer.py
```diff
@@ -1,31 +1,35 @@
...
-    assert unpacked_metadata.reward == 0.0, 'Reward unexpected.'
-    assert unpacked_metadata.rotation == 0.0, 'Rotation unexpected.'
+    def test_serialization_json(self):
+        unpacked_metadata = TestSerializer._helper_get_step_metadata()
 
+        serializer = SerializerJson()
+        json_dump = serializer.serialize(unpacked_metadata)
+        unpacked_metadata = serializer.deserialize(json_dump)
+        self.assertEqual(len(unpacked_metadata.depth_map_list), 1)
+        self.assertIsInstance(unpacked_metadata.depth_map_list[0], np.ndarray)
```
deanwetherby, line 27: Missing numpy package import for this test module to support the assertion under test.

tests/test_serializer.py
```diff
@@ -1,31 +1,35 @@
...
+class TestSerializer(unittest.TestCase):
+
+    @staticmethod
+    def _helper_get_step_metadata():
+        with open('./test.msgpack', 'rb') as file:
+            packed_bytes = file.read()
 
-    serializer = SerializerMsgPack()
-    return serializer.deserialize(packed_bytes)
+        serializer = SerializerMsgPack()
```
deanwetherby, line 12: Prefer our MCS modules to be namespaced in unit tests. The unit tests serve double duty as documentation as well so to be consistent with other unit tests, plase use 'mcs.SerializerMsgPack()'. The import on L2 should also be 'import machine_common_sense as mcs' rather than using the star import approach.

tests/test_serializer.py
```diff
@@ -1,31 +1,35 @@
...
+        assert unpacked_metadata.rotation == 0.0, 'Rotation unexpected.'
 
-def test_serialization_msgpack():
-    unpacked_metadata = _helper_get_step_metadata()
-    assert unpacked_metadata.reward == 0.0, 'Reward unexpected.'
-    assert unpacked_metadata.rotation == 0.0, 'Rotation unexpected.'
+    def test_serialization_json(self):
+        unpacked_metadata = TestSerializer._helper_get_step_metadata()
 
+        serializer = SerializerJson()
```
deanwetherby, line 23: Use the mcs namespace for SerializerJson. 

`serializer = mcs.SerializerJson()`

tests/test_serializer.py
```diff
@@ -1,31 +1,35 @@
...
 
 
-def _helper_get_step_metadata():
-    with open('./test.msgpack', 'rb') as file:
-        packed_bytes = file.read()
+class TestSerializer(unittest.TestCase):
+
+    @staticmethod
+    def _helper_get_step_metadata():
+        with open('./test.msgpack', 'rb') as file:
```
deanwetherby, line 9: Since unit tests are run from the MCS root folder, this helper function is unable to find the msgpack file. Replace the '.' with 'tests' to resolve.

`with open('tests/test.msgpack', 'rb') as file`

All MCS unit tests are run from root using this:
`(venv) $ python -m unittest`

Specific unit tests (such as the serializer) are run this way:
`(venv) $ python -m unittest tests.test_serializer`

tests/test_serializer.py
deanwetherby: Also, in order to use the Serializer classes this way, you will need to add the two class from the serializer module to the MCS __init__.py module.

Add this line to the bottom of __init__.py in the machine_common_sense folder.
`from .serializer import SerializerMsgPack, SerializerJson`

machine_common_sense/serializer.py
```diff
@@ -0,0 +1,371 @@
...
+import numpy as np
+import pathlib
+import PIL.Image as Image
+
+from abc import ABCMeta, abstractmethod
+from typing import Dict, Union
+
+import machine_common_sense as mcs
+
+from machine_common_sense import GoalMetadata, ObjectMetadata, StepMetadata
```
deanwetherby, line 12: The import style we are using for internal MCS code is to use relative imports. Replace these lines with 

`from .goal_metadata import GoalMetadata`
`from .object_metadata import ObjectMetadata`
`from .step_metadata import StepMetadata`

There's no need for the "import as mcs" line.

deanwetherby (commented):
I had some minor style recommendations to be consistent with other modules. The only other items is that the serializer module has lines that extend beyond our flake8 limit of 80. I tried to 'autopep8' to fix it but had some issues. If we could get the PEP8 formatting fixed for that module along with the other recommendations, I'll have no problem with merging.

tests/test_serializer.py
fplk: Weird. I can successfully run the test without it. But I can add it nevertheless.

tests/test_serializer.py
fplk: Good point. Added. Thank you for the step-by-step instructions as well.

tests/test_serializer.py
fplk: Done. Works perfectly. Thanks for the explanation.

machine_common_sense/serializer.py
fplk: Makes a lot of sense. Added.

fplk:
Thank you very much for your review and helpful explanations. I have applied all suggestions and am looking forward to a final check and hopefully merge of this PR. Please just let me know if there is anything else I can improve.

deanwetherby (approved):

PR #248, opened 2021-02-11
Adding atexit to always call end_scene (MCS-479).

rartiss55:
This was to take care of logs that weren't getting saved/uploaded when the software crashed.


ThomasSchellenbergNextCentury (approved):

brianpippin (approved):
This code looks good, but I am just worried that this won't cover anything if the Docker container crashes, or their code crashes, we still won't have a history file for that, but perhaps I am worried about nothing?

PR #259, opened 2021-03-05
Updating testing README and unit tests

deanwetherby:


tests/test_action.py
```diff
@@ -3,7 +3,7 @@
 import machine_common_sense as mcs
 
 
-class Test_Action(unittest.TestCase):
+class TestAction(unittest.TestCase):
```
deanwetherby, line 6: Class names should not have underscores

https://docs.python.org/2/library/unittest.html#basic-example

tests/test_action.py
```diff
@@ -237,3 +237,7 @@ def test_pass(self):
         )
         self.assertEqual(mcs.Action("Pass"), mcs.Action.PASS)
         self.assertEqual(mcs.Action(" "), mcs.Action.PASS)
+
+
+if __name__ == '__main__':
+    unittest.main()
```
deanwetherby, line 243: Providing a `__main__` allows flexibility in running unit tests individually.

https://docs.python.org/3/library/unittest.html#unittest.main

brianpippin (approved):

rartiss55 (approved):

PR #261, opened 2021-03-07
Adding scene history cleanup after test module is run

deanwetherby:
Scene history and mcs controller left history artifacts after unit tests. tearDownClass was added to both to remove that folder and all of its contents.

### DIFF
--- a/tests/test_controller.py
+++ b/tests/test_controller.py
@@ -11,6 +11,7 @@ from .mock_controller import (
 
 import os
 import glob
+import shutil
 
 SCENE_HIST_DIR = "./SCENE_HISTORY/"
 TEST_FILE_NAME = "test controller"
@@ -21,6 +22,10 @@ class Test_Controller(unittest.TestCase):
     def setUp(self):
         self.controller = MockControllerAI2THOR()
         self.controller.set_metadata_tier('')
+
+    @classmethod
+    def tearDownClass(cls) -> None:
+        shutil.rmtree(SCENE_HIST_DIR)
 
     def create_mock_scene_event(self, mock_scene_event_data):
         # Wrap the dict in a SimpleNamespace object to permit property access
--- a/machine_common_sense/history_writer.py
+++ b/machine_common_sense/history_writer.py
@@ -6,13 +6,15 @@ import os
 
 
 class HistoryWriter(object):
+
+    HISTORY_DIRECTORY = "SCENE_HISTORY"
+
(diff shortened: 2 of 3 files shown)


tests/test_controller.py
```diff
@@ -22,6 +23,10 @@ def setUp(self):
         self.controller = MockControllerAI2THOR()
         self.controller.set_metadata_tier('')
 
+    @classmethod
+    def tearDownClass(cls) -> None:
+        shutil.rmtree(SCENE_HIST_DIR)
```
brianpippin, line 28: Should we Remove line 16 and just use this now here:  mcs.HistoryWriter.HISTORY_DIRECTORY?

deanwetherby: I think it's fine that way. The tear down needed the folder name without having to create a writer. By moving the variable to be a class attribute instead of an instance attribute, we were able to accomplish the clean up. Instances can still access the class attribute and it probably reads better that way anyway.

rartiss55:
Hmm do we care if that default scene history directory had history files from previous runs? Because those would be deleted along with the scene history files created during tests.

deanwetherby:
@rartiss55 That is a good point. We could set the test history folder via the writer class attribute in setUp() to be something like 'TEST_SCENE_HISTORY' and then delete that folder on tearDownClass(). I'm finding the output folder awkward in places. Might be nice to specifically put that in the config rather than assuming.

ThomasSchellenbergNextCentury:
@deanwetherby FYI There is already a config file variable called "history_enabled": https://github.com/NextCenturyCorporation/MCS#history_enabled

I think Brian wanted the variable to default to true because we'd rather have history files and not need them than not have history files and need them (like during an eval).

brianpippin:
Yes I think we agreed the default was true, because we didn't want to run an EVAL and then realize we didn't have them and need to rerun the eval.

deanwetherby:
@ThomasSchellenbergNextCentury I was referring to the 'where' to save the history. Does it always have to be SCENE_HISTORY or can the log location be provided with in config?

rartiss55:
@deanwetherby I don't think we can change that directory by passing one in currently. We can potentially add that to the config like you're saying, or alternatively, delete specifically prefixed files only from that directory, since the history files are named using the name field of the scene.json file used (i.e. history file for a scene file with {name: "scene_1"} would be scene_1-<some_timestamp>.json).

rartiss55 (approved):

PR #266, opened 2021-03-10
Improve shared argument parsing in integration test scripts

deanwetherby:
I noticed that integration test scripts did not allow for the useful `-h` option even though the script leverage argparse. A better way to handle shared argument parsing is to create the parser in each individual script and pass that parser to the helper function which adds the added arguments and returns the parser. The original script makes the call to parse_args() so you still get the args but the `-h` option still works.

```python
(venv) $ python run_tests.py -h
usage: run_tests.py [-h] [--metadata {level1,level2,oracle}] [--test TEST]
                    [--dev]
                    mcs_unity_build_file_path

Run All Integration Tests

positional arguments:
  mcs_unity_build_file_path
                        Path to MCS unity build file

optional arguments:
  -h, --help            show this help message and exit
  --metadata {level1,level2,oracle}
                        Metadata tier to run (by default, test each metadata
                        tier)
  --test TEST           Specific test filename prefix to run (by default, all
                        files)
  --dev                 Run in dev mode (useful for adding new test scenes)
```

### DIFF
--- a/integration_tests/integration_test_utils.py
+++ b/integration_tests/integration_test_utils.py
@@ -12,10 +12,8 @@ def print_divider():
     print('========================================')
 
 
-def retrieve_test_args(name, handmade_only=False):
-    parser = argparse.ArgumentParser(
-        description=(f'Run {name} Integration Tests')
-    )
+def add_test_args(parser: argparse.ArgumentParser,
+                  handmade_only=False) -> argparse.ArgumentParser:
     parser.add_argument(
         'mcs_unity_build_file_path',
         help='Path to MCS unity build file'
@@ -42,4 +40,4 @@ def retrieve_test_args(name, handmade_only=False):
         action='store_true',
         help='Run in dev mode (useful for adding new test scenes)'
     )
-    return parser.parse_args()
+    return parser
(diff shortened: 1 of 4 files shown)


deanwetherby:
I also noticed that PyLance complains about the imports in this folder and disables autocomplete for those packages. Although I tried using relative imports to get around the import problem (which it did), this resulted in a script execution error when the integration tests were run. PyLance's suggestion was to add the folder to settings.json as extraPaths. You can get there with File -> Preferences -> Settings -> Extensions -> PyLance and look for "Extra Paths".
https://github.com/microsoft/pylance-release/issues/236#issuecomment-673594586
https://code.visualstudio.com/docs/python/editing

This is my new settings.json:

```
{
    "python.languageServer": "Pylance",
    "workbench.editorAssociations": [
        {
            "viewType": "jupyter.notebook.ipynb",
            "filenamePattern": "*.ipynb"
        }
    ],
    "python.autoComplete.extraPaths": [
        "integration_tests"
    ],
    "python.analysis.extraPaths": [
        "integration_tests"
    ]
}
```

ThomasSchellenbergNextCentury (approved):

brianpippin (approved):

PR #269, opened 2021-03-18
Mcs 551

kdrumm:
This pull request adds a new field deltaTimeMillis to each step in the history output. 

Questions:

- JSON output already has mixed CamelCase and Underscores.  I made the field CamelCase, should it stay that way?
- I outputted the time value in milliseconds.  Should it stay milliseconds?  Should it be whole unit milliseconds?  Currently it looks like this: "deltaTimeMillis": 2591.8941180109978.  I didn't find any other durations to keep consistency.

### DIFF
--- a/machine_common_sense/controller.py
+++ b/machine_common_sense/controller.py
@@ -781,7 +781,7 @@ class Controller():
             args=kwargs,
             params=params,
             output=history_copy,
-            deltaTimeMillis=0)
+            delta_time_millis=0)
 
         output = self.restrict_step_output_metadata(pre_restrict_output)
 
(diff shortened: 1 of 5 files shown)


brianpippin:
I think the JSON labels can be camelCase, but any python variables should be underscore.  Also, I think we should add tests for the millis to make sure it is working as designed.

ThomasSchellenbergNextCentury:
+1 to what Brian said. Otherwise it looks good!

kdrumm:
Added a unit test.  Updated a controller unit test.  updated formatting.

brianpippin (approved):
The changes look good to me.  However you are going to need to make a corresponding ticket in mcs-ingest that makes sure you add last_step_time_millis to the step object.  If you do add it (or just create a ticket for it), do not worry about tests or flake8 with ingest as that is a ticket I am working on adding those now.  Also if there isn't, there should be a follow up ticket about what to do now that we have the milliseconds between steps and what things we want to calculate or visualize with them.

deanwetherby (approved):

PR #270, opened 2021-03-19
Mcs 550

kdrumm:
Added Star on goal object.  Screenshot can be seen in JIRA ticket.

Due to low resolution, the star doesn't look amazing to me.  I gave it an alpha value so you can see the object borders as well.  It is easy to remove the bounds of the object if desired.

### DIFF
--- a/machine_common_sense/plotter.py
+++ b/machine_common_sense/plotter.py
@@ -53,7 +53,7 @@ class TopDownPlotter():
 
     def plot(self, scene_event: ai2thor.server.Event,
              step_number: int,
-             goal: mcs.goal_metadata.GoalMetadata
+             goal: mcs.goal_metadata.GoalMetadata = None
              ) -> PIL.Image.Image:
 
         plt = self._initialize_plot(step_number=step_number)
@@ -131,7 +131,7 @@ class TopDownPlotter():
         plt.gca().add_line(heading)
 
     def _draw_objects(self, objects: Dict,
-                      goal: mcs.goal_metadata.GoalMetadata) -> None:
+                      goal: mcs.goal_metadata.GoalMetadata = None) -> None:
         '''Plot the object bounds for each object in the scene'''
         goalId = None
         # Is there a better way to do this test?
--- a/machine_common_sense/plotter.py
+++ b/machine_common_sense/plotter.py
@@ -8,7 +8,6 @@ import matplotlib.pyplot as plt
 
 from typing import Dict, NamedTuple, List
 from shapely import geometry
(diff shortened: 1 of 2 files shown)


machine_common_sense/plotter.py
```diff
@@ -127,15 +131,36 @@ def _draw_robot_heading(self, robot: Robot) -> None:
                              lw=1)
         plt.gca().add_line(heading)
 
-    def _draw_objects(self, objects: Dict) -> None:
+    def _draw_objects(self, objects: Dict,
+                      goal: mcs.goal_metadata.GoalMetadata) -> None:
         '''Plot the object bounds for each object in the scene'''
+        goalId = None
+        # Is there a better way to do this test?
```
deanwetherby, line 137: Curious if we could just pass the goalId to plot instead of the entire GoalMetadata. Then this test would be done from the caller of the function while the plotter code remains clean-ish.

machine_common_sense/plotter.py
```diff
@@ -127,15 +131,36 @@ def _draw_robot_heading(self, robot: Robot) -> None:
...
+        if (goal is not None and goal.metadata is not None):
+            goalId = goal.metadata.get('target', {}).get('id', None)
         for o in objects:
             obj = self._create_object(o)
             if obj.bounds is not None:
                 obj_pts = [(pt['x'], pt['z']) for pt in obj.bounds]
                 polygon = geometry.MultiPoint(obj_pts).convex_hull
                 pts = polygon.exterior.coords
                 self._draw_object_bounds(obj, pts)
+                if goalId is not None and o['objectId'] == goalId:
```
deanwetherby, line 147: Should there be an else here to draw the object bounds if not the goal (star) object?

kdrumm: The bounds are drawn already on line 147.  I debated whether they should both be drawn, but wasn't sure if the size was important for the goal.  It is certainly easy to do what you say if that is the desired effect.

machine_common_sense/plotter.py
kdrumm: Agreed and fixed.

machine_common_sense/plotter.py
deanwetherby: Right. So do we want the bounds with the star or just the star. Originally I was thinking just the star since most of the time the goal object is small. I guess it depends on what the plot looks like. Your call.

deanwetherby (approved):

ThomasSchellenbergNextCentury (commented):
I know I'm late, but LGTM anyway :)

PR #277, opened 2021-03-24
Adding integration tests for wall materials

kdrumm:
Added integration tests, documentation and sample scene for individually configurable wall materials.   

Is dependent on and corresponds to: https://github.com/NextCenturyCorporation/ai2thor/pull/119


ThomasSchellenbergNextCentury (commented):
I think you already have an integration test numbered 062 in your MCS-95 PR? (There's nothing preventing you from making two tests with the same number, but it's probably clearer if they all have different numbers.) Otherwise everything looks good.

kdrumm:
You're right!  I skipped 1 because i knew there was some in MCS-95.  Obviously counting to 2 is not my strong suit...  Fixed now.

ThomasSchellenbergNextCentury (approved):

PR #281, opened 2021-03-29
[ON HOLD] Variable Movement and Rotation

erick-u:
Added parameter validation and descriptions to actions


erick-u:
Related AI2Thor PR https://github.com/NextCenturyCorporation/ai2thor/pull/122

deanwetherby (approved):

brianpippin (approved):

brianpippin:
I know I approved this already, but should we add anything to the unit or integration tests for this?

kdrumm (commented):
I agree with Brian.  Code looks good, but I think we should add a test.

machine_common_sense/controller.py
```diff
@@ -128,11 +128,18 @@ class Controller():
     MIN_FORCE = 0
     MAX_AMOUNT = 1
     MIN_AMOUNT = 0
+    MAX_MOVEMENT = 100
```
ThomasSchellenbergNextCentury, line 131: We should add the max move distance to our docs. (We should probably add the max rotation and horizon amounts as well; even though they SHOULD be obvious, I'd be extra clear...)

ThomasSchellenbergNextCentury:
Agreed concerning adding new integration tests. The documentation is [here](https://github.com/NextCenturyCorporation/MCS/tree/development/integration_tests). It would be good to not only have tests that verify moving and rotating large amounts, but also have tests in which another object is moving that will verify the time delta. You can see an example of some scripted object movement (without requiring interaction from the player) in [integration test 55](https://github.com/NextCenturyCorporation/MCS/blob/development/integration_tests/data/055.scripted_force.scene.json).

erick-u:
Great points on having the integration tests, I will get them added.

brianpippin (approved):

machine_common_sense/controller.py
erick-u: Should be worth nothing this max amount is arbitrary as it will always cap out at the further it can travel before hitting a wall.

machine_common_sense/controller.py
```diff
@@ -645,6 +659,24 @@ def validate_and_convert_params(self, action, **kwargs):
             self.MAX_FORCE,
             self.DEFAULT_FORCE,
             self.FORCE_KEY)
+        movement = Util.is_in_range(
+            float(movement),
```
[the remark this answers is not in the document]
deanwetherby: This is fine for now. I did notice that force and amount should probably be converted to floats as well though. I'm going to refactor the controller soon and utils might go the way of the dodo.

brianpippin (approved):

brianpippin:
@ThomasSchellenbergNextCentury - What should we do with this code review?  It is pretty far in the past?  Should we close and move on, or is this worth revisiting any time soon?

ThomasSchellenbergNextCentury:
@brianpippin I think this is too old at this point, and no one from TA1 has asked us about it in a long time, so I vote to close it. My previous concerns remain, and we'd still need to do a good deal of work to make this ready to merge, and it isn't a priority right now. If it ever becomes important later, the PR doesn't disappear when closed, so we can refer back to it if necessary.

PR #282, opened 2021-03-29
Mcs 472 platforms

kdrumm:
Integration tests and scene file for pull request found at: https://github.com/NextCenturyCorporation/ai2thor/pull/123

### DIFF
--- a/integration_tests/data/065.platform.actions.txt
+++ b/integration_tests/data/065.platform.actions.txt
@@ -1,0 +1,17 @@
+MoveAhead
+MoveAhead
+MoveAhead
+MoveAhead
+MoveAhead
+MoveAhead
+MoveBack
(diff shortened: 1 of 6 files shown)


rartiss55 (approved):
Approving. Noticed the oracle outputs file was a lot bigger than the others (I think that usually happens but it seemed significantly bigger this time so wanted to flag it just in case).

deanwetherby (approved):

kdrumm:
> Approving. Noticed the oracle outputs file was a lot bigger than the others (I think that usually happens but it seemed significantly bigger this time so wanted to flag it just in case).

Good to note.  I did look into it just to make sure.  It seems the size is mostly caused by it outputing structural objects and the platforms add 2 extra structural objects per step.

PR #285, opened 2021-03-30
Serialization Fix

fplk:
To prevent the shape from being a bool instead of a string after serialization I have upgraded the serializers.
This PR supercedes https://github.com/NextCenturyCorporation/MCS/pull/284 which was accidentally filed against the development branch rather than master [development is currently dysfunctional for outside contributors, since we don't have access to 0.4.1, yet => filed against master].


tests/test_serializer.py
```diff
@@ -30,8 +31,8 @@ def test_serialization_json(self):
         self.assertIsInstance(unpacked_metadata.depth_map_list[0], np.ndarray)
         self.assertEqual(unpacked_metadata.depth_map_list[0].shape, (400, 600))
         self.assertIsInstance(unpacked_metadata, mcs.StepMetadata)
-        self.assertEqual(unpacked_metadata.reward, 0.0)
-        self.assertEqual(unpacked_metadata.rotation, 0.0)
+        self.assertEqual(abs(unpacked_metadata.reward - (-0.036000000000000004)) < 1e-04, True)
+        self.assertEqual(abs(unpacked_metadata.rotation - 0.0) < 1e-04, True)
```
deanwetherby, line 37: We use `assertAlmostEqual` in our other tests for floating point comparisons.  You could use the default epilson value of 1e-07 or specify your own delta value. 

```
        self.assertAlmostEqual(
            writer.current_steps[2]["delta_time_millis"], 0, delta=.1)
```

https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertAlmostEqual

requirements.txt
```diff
@@ -17,6 +17,7 @@ Jinja2==2.11.3
 MarkupSafe==1.1.1
 matplotlib==3.3.2
 msgpack==1.0.0
+numpy==1.19.5
```
deanwetherby, line 20: Our opencv-python 4.4.0.44 dependency pulls in numpy 1.19.4. Is there something in the 1.19.5 you need or are you thinking that it's better to specify this dependency?

tests/test_serializer.py
```diff
@@ -27,12 +28,14 @@ def test_serialization_json(self):
...
         json_dump = serializer.serialize(unpacked_metadata)
         unpacked_metadata = serializer.deserialize(json_dump)
+
         self.assertEqual(len(unpacked_metadata.depth_map_list), 1)
         self.assertIsInstance(unpacked_metadata.depth_map_list[0], np.ndarray)
         self.assertEqual(unpacked_metadata.depth_map_list[0].shape, (400, 600))
         self.assertIsInstance(unpacked_metadata, mcs.StepMetadata)
         self.assertEqual(abs(unpacked_metadata.reward - (-0.036000000000000004)) < 1e-04, True)
         self.assertEqual(abs(unpacked_metadata.rotation - 0.0) < 1e-04, True)
+        self.assertEqual(type(unpacked_metadata.object_list[-1].shape), str)
```
deanwetherby, line 38: You could use `assertIsInstance` here

```
self.assertIsInstance(unpacked_metadata.object_list[-1].shape, str)
```

https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIsInstance

fplk:
Thank you for the feedback. All change requests should be implemented now. Please reevaluate.

deanwetherby (approved):

deanwetherby:
Missed formatting

PR #289, opened 2021-03-31
MCS-540 teleport during end habituation

rartiss55:
Associated ai2thor PR here: https://github.com/NextCenturyCorporation/ai2thor/pull/125

Note that I used number 66 for my integration tests (65 looks like I skipped it because I branched off development before a recently merged test).


brianpippin (approved):
The integration tests passed, and I don't see anything glaring in the code.

machine_common_sense/action.py
```diff
@@ -564,12 +564,28 @@ class Action(Enum):
...
     )
     """
-    Ends a habituation trial for the scene by blanking the screen for one
-    action. Sometimes needed for the passive tasks.
+    Ends a habituation trial for the scene by blanking the screen for one action
+    (and teleporting the agent if needed). Sometimes needed depending on the task type. 
+
+    Note that we currently plan to use the starting position/rotation as teleport 
+    parameters here for applicable cases. We cannot currently guarantee that using 
+    a random position won't cause issues or errors.
```
ThomasSchellenbergNextCentury, line 577: Maybe rather than "random position" say something like "position intersecting another object or outside the room" to be more specific.

rartiss55: Thanks for the suggestion, updated it. Also when I was doing this realized I didn't have all my pre commit hook stuff/linting setup correctly (using a fresh checkout from when I was testing the ai2thor upgrade) so I fixed that too :)

kdrumm (approved):
Changes look good.  All tests pass.

ThomasSchellenbergNextCentury (approved):

PR #297, opened 2021-04-06
Updated images to 0.4.1.1

fplk:
I don't think the current Dockerfile update could work. You currently do
```
ARG mcsversion=0.4.1.1
ADD https://github.com/NextCenturyCorporation/MCS/releases/download/${mcsversion}/MCS-AI2-THOR-Unity-App-v${mcsversion}.x86_64 /mcs
```
but the first occurrence of `mcsversion` should have a minus instead of a last dot. Thus, I think that URL is invalid. This PR incorporates the minus at the right location. I have not extensively tested it beyond building, but it seems more correct than the current version. Another advantage is that the PR parameterizes the library version and also makes the CPU and GPU image more consistent, since currently one puts MCS into root ant the other into `/mcs`.

Or put another way: https://github.com/NextCenturyCorporation/MCS/releases/download/0.4.1.1/MCS-AI2-THOR-Unity-App-v0.4.1.1.x86_64 is invalid, whereas https://github.com/NextCenturyCorporation/MCS/releases/download/0.4.1-1/MCS-AI2-THOR-Unity-App-v0.4.1.1.x86_64 should work.

### DIFF
--- a/CPU_Container.dockerfile
+++ b/CPU_Container.dockerfile
@@ -10,6 +10,11 @@ ENV TZ=America/New_York
 
 # --build-arg mcsversion=0.0.x to override default in docker build command
 ARG mcsversion=0.4.1.1
+ARG mcsversion_minus=0.4.1-1
+ARG mcs_library_version=master
+
+WORKDIR /mcs
+
 
 ENV DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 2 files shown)


ThomasSchellenbergNextCentury (approved):
We'll try to avoid causing this issue again moving forward, but it's good to have this fix in place anyway.

deanwetherby (approved):

PR #299, opened 2021-04-08
MCS-441 - Add performer reach and radius to metadata

### LINKED ISSUE #239: Return Performer Agent Radius and Reach in Output Step Metadata
From: Falk, Mangannr

Priority: Backlog

Components: Python API, Unity Env

Blockers: None

Goal: Return the performer’s radius and reach so the TA1s always know their exact value, in case we change them in the future, since they’re useful for the TA1s in developing the ML models.

In Unity:
- Add two new variables (maybe called performerRadius and performerReach) to the MetadataWrapper class in the AgentManager file.
- Return the MCSController’s CapsuleCollider’s Radius in the output MetadataWrapper as the performerRadius (MCSController.generateMetadataWrapper).
- Return the MCSController’s maxVisibilityDistance (this variable sets the performer’s reach) in the output MetadataWrapper as the performerReach (MCSController.generateMetadataWrapper).

In Python:
- Add two new float properties (maybe called performer_radius and performer_reach) to the StepMetadata class, and update its documentation.
- Fill the two new properties in the MCS Controller’s wrap_output function with the radius and reach returned from Unity that you just implemented above (in the scene_event.metadata).
- Update the unit tests.

Slack links:
- https://machinecommonsense.slack.com/archives/CTR31V12M/p1605536698205900?thread_ts=1605536642.205800&cid=CTR31V12M
- https://machinecommonsense.slack.com/archives/CTR31V12M/p1605603290211400

TA2 internal JIRA ticket: https://nextcentury.atlassian.net/browse/MCS-441

rartiss55:
Closes #239.

Associated ai2thor change here: https://github.com/NextCenturyCorporation/ai2thor/pull/130

### DIFF
--- a/machine_common_sense/API.md
+++ b/machine_common_sense/API.md
@@ -384,7 +384,7 @@ Defines output metadata from an action step in the MCS 3D environment.
     * **performer_radius** (*float*) – The radius of the performer.
 
 
-    * **performer_reach** (*float*) – The max reach of the performer (equal to the max visibility distance).
+    * **performer_reach** (*float*) – The max reach of the performer (equal to the visibility distance).
 
 
     * **pose** (*string*) – Your current pose. Either “STANDING”, “CRAWLING”, or “LYING”.
--- a/machine_common_sense/API.md
+++ b/machine_common_sense/API.md
@@ -313,7 +313,7 @@ Defines metadata for an object in the MCS 3D environment.
 ## StepMetadata
 
 
-### class machine_common_sense.step_metadata.StepMetadata(action_list=None, camera_aspect_ratio=None, camera_clipping_planes=None, camera_field_of_view=0.0, camera_height=0.0, depth_map_list=None, goal=None, habituation_trial=None, head_tilt=0.0, image_list=None, object_list=None, object_mask_list=None, pose='UNDEFINED', position=None, return_status='UNDEFINED', reward=0, rotation=0.0, step_number=0, structural_object_list=None)
+### class machine_common_sense.step_metadata.StepMetadata(action_list=None, camera_aspect_ratio=None, camera_clipping_planes=None, camera_field_of_view=0.0, camera_height=0.0, depth_map_list=None, goal=None, habituation_trial=None, head_tilt=0.0, image_list=None, object_list=None, object_mask_list=None, performer_radius=0.0, performer_reach=0.0, pose='UNDEFINED', position=None, return_status='UNDEFINED', reward=0, rotation=0.0, step_number=0, structural_object_list=None)
 Defines output metadata from an action step in the MCS 3D environment.
 
 
@@ -379,6 +379,12 @@ Defines output metadata from an action step in the MCS 3D environment.
     scene with a scripted Previous Phase.
     The color of each object in the mask corresponds to the “color”
     property in its ObjectMetadata object.
+
+
+    * **performer_radius** (*float*) – The radius of the performer.
+
+
+    * **performer_reach** (*float*) – The max reach of the performer (equal to the max visibility distance).
 
 
     * **pose** (*string*) – Your current pose. Either “STANDING”, “CRAWLING”, or “LYING”.
--- a/machine_common_sense/API.md
+++ b/machine_common_sense/API.md
@@ -384,7 +384,7 @@ Defines output metadata from an action step in the MCS 3D environment.
     * **performer_radius** (*float*) – The radius of the performer.
 
 
-    * **performer_reach** (*float*) – The max reach of the performer (equal to the visibility distance).
+    * **performer_reach** (*float*) – The max reach of the performer.
 
 
     * **pose** (*string*) – Your current pose. Either “STANDING”, “CRAWLING”, or “LYING”.
--- a/machine_common_sense/controller.py
+++ b/machine_common_sense/controller.py
@@ -1305,6 +1305,8 @@ class Controller():
             object_mask_list=object_mask_list,
             pose=self.retrieve_pose(scene_event),
             position=self.retrieve_position(scene_event),
+            performer_radius=scene_event.metadata.get('performerRadius'),
+            performer_reach=scene_event.metadata.get('performerReach'),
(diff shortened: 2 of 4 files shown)


machine_common_sense/API.md
```diff
@@ -313,7 +313,7 @@ Defines metadata for an object in the MCS 3D environment.
 ## StepMetadata
 
 
-### class machine_common_sense.step_metadata.StepMetadata(action_list=None, camera_aspect_ratio=None, camera_clipping_planes=None, camera_field_of_view=0.0, camera_height=0.0, depth_map_list=None, goal=None, habituation_trial=None, head_tilt=0.0, image_list=None, object_list=None, object_mask_list=None, pose='UNDEFINED', position=None, return_status='UNDEFINED', reward=0, rotation=0.0, step_number=0, structural_object_list=None)
+### class machine_common_sense.step_metadata.StepMetadata(action_list=None, camera_aspect_ratio=None, camera_clipping_planes=None, camera_field_of_view=0.0, camera_height=0.0, depth_map_list=None, goal=None, habituation_trial=None, head_tilt=0.0, image_list=None, object_list=None, object_mask_list=None, performer_radius=0.0, performer_reach=0.0, pose='UNDEFINED', position=None, return_status='UNDEFINED', reward=0, rotation=0.0, step_number=0, structural_object_list=None)
```
deanwetherby, line 316: I struggle quite a bit with agent/robot/performer myself but do we need the 'performer' prefix here? Or can 'radius' and 'reach' stand on their own? For example, we have 'head_tilt' and not 'performer_head_tilt'.  I noticed in the step_metadata docs that we talk about 'player' so that's another one to add to the list.

rartiss55: I was more following the ticket description, but I do think its nice to specify here that we're referring to a performer/robot/agent attribute. 'head_tilt' seems a lot more obvious to me than the properties added -- 'reach' I could maybe see on its own, but 'radius' seems a lot more iffy to me on what we're referring to.

kdrumm: I think adding performer (or whatever the appropriate word is) is helpful here. Radius by itself can be almost anything.

I struggle with the word reach in this case.  To me, reach would be the distance from the performer an object can be while the performer can pick it up.  Isn't the visibility distance how far the performer can see?

ThomasSchellenbergNextCentury: > Isn't the visibility distance how far the performer can see?

@kdrumm This unfortunately gets a little confusing. In AI2-THOR, "reach" and "visibility distance" mean the same thing (you can only see objects you can reach, and visa-versa). For us, we don't want that; we've currently hard-coded "reach" to 1, and we've aligned the visibility distance to our camera's far clipping plane (15). However, some of the AI2-THOR code calls things "visibility_distance" when it's really talking about reach. Hopefully that clears some things up for you.

I have no opinion on the variable names.

machine_common_sense/API.md
```diff
@@ -381,6 +381,12 @@ Defines output metadata from an action step in the MCS 3D environment.
     property in its ObjectMetadata object.
 
 
+    * **performer_radius** (*float*) – The radius of the performer.
+
+
+    * **performer_reach** (*float*) – The max reach of the performer (equal to the visibility distance).
```
ThomasSchellenbergNextCentury, line 387: @rartiss55 We probably don't need to mention the "visibility distance" here, since that's just a term the AI2-THOR code uses internally, and is likely very confusing out of context (since you can see -- in both the images and the output metadata -- things that are out-of-reach).

machine_common_sense/controller.py
```diff
@@ -1305,6 +1305,8 @@ def wrap_output(self, scene_event):
             object_mask_list=object_mask_list,
             pose=self.retrieve_pose(scene_event),
             position=self.retrieve_position(scene_event),
+            performer_radius=scene_event.metadata.get('performerRadius'),
+            performer_reach=scene_event.metadata.get('performerReach'),
```
ThomasSchellenbergNextCentury, line 1309: I didn't actually think about it until now, but we're currently setting the MAX_REACH_DISTANCE in Python, sending that value to Unity via the wrap_step function, and getting it back via `scene_event.metadata.get('performerReach')`

I think the way you have it here is the best, so no change needed for my sake -- in fact, we may want to consider moving the MAX_REACH_DISTANCE from Python into Unity, since IMO it seems more appropriate to be defined there, but it's not a big deal.

Just pointing it out because it seemed funny, is all.

machine_common_sense/API.md
kdrumm: @ThomasSchellenbergNextCentury Thanks for the explaination.

machine_common_sense/API.md
rartiss55: Yeah I was debating that one. Updated!

machine_common_sense/controller.py
rartiss55: I looked again when you mentioned it, yes, kind of looks like a chicken and egg thing now. I left a TODO on the line above MAX_REACH_DISTANCE to address this later.

(Unrelated to this ticket, but there was another constant PERFORMER_CAMERA_Y that I just took out when addressing comments for this PR, since we're not using it anywhere and it had the old camera height).

ThomasSchellenbergNextCentury: I'm glad you mentioned the PERFORMER_CAMERA_Y, because I'm actually importing that in the scene generator! However, if it's not being used here, I should really just define it inside the scene generator, rather than import it from machine_common_sense. Easy for me to change, and probably cleaner that way as well.

ThomasSchellenbergNextCentury (approved):

kdrumm (approved):

machine_common_sense/controller.py
rartiss55: Ah sorry, missed that. I added it back in for now with the new height (0.762) and put in a ticket to move it to the scene generator later (MCS-627).

brianpippin (approved):

PR #300, opened 2021-04-08
Removing goal id restrictions

deanwetherby:


rartiss55:
LGTM, although I think this broke some unit tests (maybe we can just comment those out for now as well).

Did we need an additional ticket for circling back to this, or do we think https://nextcentury.atlassian.net/browse/MCS-24 covers it?

deanwetherby:
> LGTM, although I think this broke some unit tests (maybe just comment those out for now as well).
> 
> Did we need an additional ticket for circling back to this, or does https://nextcentury.atlassian.net/browse/MCS-24 cover it?

Thanks. I'll fix the tests.

MCS-24 is on the ingest side. I'll also write another ticket for circling back.

ThomasSchellenbergNextCentury (approved):
This looks good to me, and I can totally appreciate doing this simple solution to address the issue right now. However, if we decide we want to go back to removing the target IDs for the lower metadata tiers, maybe the following would work?

Current:
1. Create the GoalMetadata object in start_scene https://github.com/NextCenturyCorporation/MCS/blob/b3fe764ddc2d2a8991d8b7a83cd6a293d2937720/machine_common_sense/controller.py#L472
2. Set the `goal` property of the StepMetadata object in `wrap_output`
https://github.com/NextCenturyCorporation/MCS/blob/b3fe764ddc2d2a8991d8b7a83cd6a293d2937720/machine_common_sense/controller.py#L815-L816
https://github.com/NextCenturyCorporation/MCS/blob/b3fe764ddc2d2a8991d8b7a83cd6a293d2937720/machine_common_sense/controller.py#L1296
3. Remove the target data from the StepMetadata's `goal` property
https://github.com/NextCenturyCorporation/MCS/blob/b3fe764ddc2d2a8991d8b7a83cd6a293d2937720/machine_common_sense/controller.py#L830
https://github.com/NextCenturyCorporation/MCS/blob/b3fe764ddc2d2a8991d8b7a83cd6a293d2937720/machine_common_sense/controller.py#L951-L959
  
Changes:
- In step 2 above, `copy.deepcopy(self._goal)` when setting the `goal` property in the StepMetadata object, so that when step 3 happens, it affects a different object

deanwetherby:
@ThomasSchellenbergNextCentury Added your comment to the new ticket. Going to merge this one now.

PR #301, opened 2021-04-08
MCS-620 New config file property to open/close container

kdrumm:
Updates to test open and close from config file by adding isOpen and openable to metadata
Added integration test to test open/close
Added new scene to show open and close

Associated with: https://github.com/NextCenturyCorporation/ai2thor/pull/131

### DIFF
--- a/machine_common_sense/object_metadata.py
+++ b/machine_common_sense/object_metadata.py
@@ -52,7 +52,7 @@ class ObjectMetadata(object):
         This object's colors, derived from its textures, in plain English.
     visible : boolean
         Whether you can see this object in your camera viewport.
-    isOpen : boolean
+    is_open : boolean
         Whether the object is open or not
     openable : boolean
         Whether the object can be opened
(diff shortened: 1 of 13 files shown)


machine_common_sense/object_metadata.py
```diff
@@ -52,6 +52,10 @@ class ObjectMetadata(object):
         This object's colors, derived from its textures, in plain English.
     visible : boolean
         Whether you can see this object in your camera viewport.
+    isOpen : boolean
```
ThomasSchellenbergNextCentury, line 55: In Python, we try to keep all of the variable names in snake_case, since that's the commonly-used Python naming convention. (I know this is a little confusing since we use camelCase in Unity, since that's its naming convention for local variables, and thus we have to convert the names in scenarios like this.)

ThomasSchellenbergNextCentury (changes requested):
I really appreciate the integration test and example scene file. However, I'm getting some failures/errors in the unit tests.

kdrumm:
Updated.  Reran unit tests and integration tests with no failures on my side.

ThomasSchellenbergNextCentury (approved):

PR #304, opened 2021-04-13
MCS-473 Fix to picking contained objects when picking up a container

kdrumm:
Integration tests to correspond to https://github.com/NextCenturyCorporation/ai2thor/pull/132

Note: Integration test 46 and 48 will fail until pull request in MCS 303 is merged. NextCenturyCorporation/MCS#303

### DIFF
--- a/integration_tests/data/071.pickup_container.actions.txt
+++ b/integration_tests/data/071.pickup_container.actions.txt
@@ -1,0 +1,6 @@
+﻿LookDown
+MoveAhead
+PickupObject,objectId=bowl_a
+Crawl
+DropObject
+Pass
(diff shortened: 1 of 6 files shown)


rartiss55 (approved):

brianpippin (approved):
Is there an integration test for stacking objects?  If there isn't, I would potentially add one to make sure stacking in general still works.  Also I would wait for @ThomasSchellenbergNextCentury to chime in on this before merging just to be on the safe side.

kdrumm:
There are a bunch of block stacking tests.  I also agree that I was going to wait on Thomas before merging.

PR #308, opened 2021-04-16
Mcs 512 return frame rate

kdrumm:
Added frame rate to metadata and updated tests.

Integration tests fail if run sequentially until NextCenturyCorporation/MCS#281 is merged.

Associated with pull request: https://github.com/NextCenturyCorporation/ai2thor/pull/134


machine_common_sense/step_metadata.py
```diff
@@ -85,6 +85,9 @@ class StepMetadata:
     step_number : integer
         The step number of your last action, recorded since you started the
         current scene.
+    physics_frames_per_second : float
```
brianpippin, line 88: We need a corresponding ingest ticket for this, to add this to ingest, and then make sure it is visible in the UI.  Currently we pick and choose which parts of a "STEP" get added to the UI database because of how large the STEP_OUTPUT is, so we just need a line in ingest to add this in if we want to visualize it, which I assume we do.  A follow on ticket should be sufficient now, and if you would like to see how to work with the ingest code, i'd be happy to help with that.

brianpippin (approved):

machine_common_sense/step_metadata.py
kdrumm: Good idea.  I added a ticket, MCS-633.  I'd also like to see how to work with the ingest code.

machine_common_sense/step_metadata.py
```diff
@@ -85,6 +85,9 @@ class StepMetadata:
...
+        The frames per second that the physics engine has advanced during the
+        current step
```
ThomasSchellenbergNextCentury, line 90: I think this description is a little confusing. Could it just be "The frames per second of the physics engine" ?

kdrumm: updated

ThomasSchellenbergNextCentury (approved):

PR #311, opened 2021-04-22
sample logging config initialization

kdrumm:
This is a possible method for us to initialize logging for local scripts.  Comments and suggests are strongly encouraged.

My goals were:

- Minimal, easy changes to any script to add logging that could be consistent
- Allow for developer to make custom changes that wouldn't be checked into git without code changes
- A reasonable default for anyone who doesn't want to mess with logging configuration

One issue: it will errors without an obvious message if the ./logs folder doesn't exist.  We could use ./ but I'd rather not let it throw a couple log files into the root directory.

### DIFF
--- a/machine_common_sense/__init__.py
+++ b/machine_common_sense/__init__.py
@@ -1,4 +1,7 @@
 import logging
+import logging.config
+import ast
+from os.path import exists
 import json
 import signal
 
@@ -21,8 +24,9 @@ from .serializer import SerializerMsgPack, SerializerJson
 from ._version import __version__
 
 
+logger = logging.getLogger(__name__)
 # Set default logging handler to avoid "No handler found" warnings
-logging.getLogger(__name__).addHandler(logging.NullHandler())
+logger.addHandler(logging.NullHandler())
 
 # Timeout at 3 minutes (180 seconds).  It was 60 seconds but
 # this can cause timeouts on EC2 instances
@@ -96,3 +100,29 @@ def load_scene_json_file(scene_json_file_path):
     except IOError:
         return {}, "The given file '" + scene_json_file_path + \
             "' cannot be found."
+
+
+def init_logging():
+    """
+    Initializes logging system.  Attempts to read user file first,
+    which should not be checked in and each user can have their own.
+    If user file doesn't exist, then there is a base config file that
+    should be read.
+    """
+    log_config_base = "scripts/log.config.py"
+    log_config_user = "scripts/log.config.user.py"
+    log_config_file = None
+    if (exists(log_config_user)):
+        log_config_file = log_config_user
+    if (exists(log_config_base)):
+        log_config_file = log_config_base
+    if (log_config_file is not None):
+        with open(log_config_file, "r") as data:
+            logConfig = ast.literal_eval(data.read())
+        logging.config.dictConfig(logConfig)
+        logger.info(
+            "Loaded logging config from " + log_config_file)
+    else:
+        print(
+            "Error initializing logging.  No file found at " +
+            log_config_base + " or " + log_config_user)
--- a/machine_common_sense/util.py
+++ b/machine_common_sense/util.py
@@ -1,8 +1,5 @@
 import numpy
 import logging
-import logging.config
-import ast
-from os.path import exists
 
 from .action import Action
 from .material import Material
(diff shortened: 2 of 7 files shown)


scripts/run_human_input.py
```diff
@@ -209,6 +209,7 @@ def run_scene(controller, scene_data):
 
 
 def main():
+    mcs.Util.init_logging()
```
kdrumm, line 212: This is the only line (besides maybe an import of MCS) that needs to be added to any scripts that need logging.

machine_common_sense/util.py
```diff
@@ -314,3 +317,28 @@ def verify_material_enum_string(enum_string):
...
+            log_config_file = log_config_base
+        if (log_config_file is not None):
+            with open(log_config_file, "r") as data:
+                logConfig = ast.literal_eval(data.read())
+            logging.config.dictConfig(logConfig)
+            logger.info("Loaded logging config from " + log_config_file)
+        else:
+            print(
+                "Error initializing logging.  No file found at " +
+                log_config_base + " or " + log_config_user)
```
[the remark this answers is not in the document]
deanwetherby: Another way to handle this is to put the logging initialization for these run scripts as a function in `__init__.py`. Then usage would be something like:

```

mcs.init_logging('config/file/path')
```

That was my plan for initializing formal evaluation file logging anyway.

machine_common_sense/__init__.py
```diff
@@ -96,3 +100,29 @@ def load_scene_json_file(scene_json_file_path):
...
             "' cannot be found."
+
+
+def init_logging():
+    """
+    Initializes logging system.  Attempts to read user file first,
+    which should not be checked in and each user can have their own.
+    If user file doesn't exist, then there is a base config file that
+    should be read.
+    """
```
ThomasSchellenbergNextCentury, line 111: We should add some documentation to our README about the use case here, like:
- How to use this (code and config file instructions/example)
- Why to use this (specifically in comparison to importing the logger library and instantiating a logger variable yourself)
I'm assuming this feature makes some workflow easier, but it's not clear to me exactly how (probably due to my unfamiliarity with using logging config files).

machine_common_sense/__init__.py
```diff
@@ -96,3 +100,29 @@ def load_scene_json_file(scene_json_file_path):
...
+    log_config_base = "scripts/log.config.py"
+    log_config_user = "scripts/log.config.user.py"
+    log_config_file = None
+    if (exists(log_config_user)):
+        log_config_file = log_config_user
+    if (exists(log_config_base)):
+        log_config_file = log_config_base
+    if (log_config_file is not None):
```
ThomasSchellenbergNextCentury, line 119: Don't feel obligated to change this, but as an FYI, you can just do `if log_config_file:` here. When I see an if statement like this, I assume that there's a reason for the variable to sometimes be falsey but not `None` (like `0` or `""`), but I don't think that's the case here.

machine_common_sense/__init__.py
kdrumm: I'll write something up for the readme, but let me see if I can explain the why here.  

The main goal of a logger is logging configurability (I.E. turning different logs on and off, changing where logs go).  As a developer, I want to have a logging system that I can easily alter to me needs while not stepping on other developers toes.  This puts all the log configuration in 1 location (2 files) for any script that initializes this way.  One file doesn't get checked in so I can copy the default to the personal file and alter it without ever accidentally checking it in and stepping on other developers.  Additionally, setting up the log with the features in the default config file (rotating file logger and console logger, different log levels for MCS code and non-MCS code) would require a decently large block of code that would be copied around from file to file.

This will be more useful if we start using the log levels more.  Most of our logs appear to be on the DEBUG level with some WARN and ERROR.  If we used TRACE and INFO more, we can turn things on and off to help with debugging when we need or don't need the detail.  We can essentially turn the logging level up and down on a per file level.  (although this would be more useful if we didn't have massive files). If we just want to always print all our logs to the console, then why did we even switch to a logger?

Most of the time when I've used loggers, we used the same code for development, staging, and production, but we often wanted different levels of logging without needing to change the code.  This is particularly useful for compiled languages.  In our case, this particular points seem less useful.  However, the logging configuration may also be useful to debug TA1 issues - send them a config file, let the run their system, they send us the log file back - all without changing code.

ThomasSchellenbergNextCentury: That's a great explanation. Thank you!

kdrumm: I've updated some documentation.  Its more concise than this.

ThomasSchellenbergNextCentury: Thanks! That looks great.

ThomasSchellenbergNextCentury (approved):

Issue #307, opened 2021-04-15
Consistent action_list Format


ThomasSchellenbergNextCentury:
From: Falk

Priority: Eval 4

Components: Python API

Blockers: None

Issues:
- Both `GoalMetadata.action_list` and `StepMetadata.action_list` return action data in different formats (the former as strings like `action,param1=value1,param2=value2`, the latter as tuples like `(action_string, params_dict)`). Choose a format and make it consistent across both variables (and update the documentation).
- Update the developer scripts as needed.
- Consider adding public documentation on `mcs.Util.input_to_action_and_params()` to convert action strings into action/parameter tuples.
- Report our changes/expectations to TA1.

Slack links:
- https://machinecommonsense.slack.com/archives/CTR31V12M/p1618326327054700

ThomasSchellenbergNextCentury:
@fplk The fix for this issue is currently in our development branch, but we're planning on releasing a new version soon. We've changed the action_list to always return tuples, and have fixed our developer scripts like run_human_input. We know this is a change from Eval 2 and 3, but we believe that it is a useful one. We'll make sure to post all the details with our upcoming release.

PR #312, opened 2021-04-22
MCS-630 Unified action list format to always use tuples and updated human input mode script.

ThomasSchellenbergNextCentury:
Fixes https://github.com/NextCenturyCorporation/MCS/issues/307

Changes:
- Both GoalMetadata.action_list and StepMetadata.action_list now save actions and parameters as tuples rather than strings.
- Added "auto" mode to human input run script that automatically runs the next step if only one action is available.
- Reordered terminal output in human input run script to be more useful.
- Updated python run scripts and documentation.

TBD:
- Consider releasing a new python version.
- Post on Confluence about these changes.

### DIFF
--- a/machine_common_sense/API.md
+++ b/machine_common_sense/API.md
@@ -132,6 +132,11 @@ Make a prediction on the previously taken step/action.
 
 
 
+#### retrieve_action_list_at_step(goal, step_number)
+Return the action list from the given goal at the given step as a
+a list of actions tuples by default.
+
+
 #### retrieve_object_states(object_id)
 Return the state list at the current step for the object with the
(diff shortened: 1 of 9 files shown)


kdrumm (approved):

brianpippin:
Did you verify that the tuple change doesn't break the ability to write the history file?  I know we had issues with some dictionaries and tuples in the past?

ThomasSchellenbergNextCentury:
@brianpippin Good question. I just checked it, and the scene history files are valid JSON. The tuples are automatically converted into JSON arrays, so they should be just fine. Thanks for the reminder!

brianpippin (approved):

PR #319, opened 2021-04-27
adding incremental history file writing

kdrumm:


README.md
```diff
@@ -269,6 +269,12 @@ To use an MCS configuration file, you can either pass in a file path via the `co
 
 Whether to save the scene history output data in your local directory. Default: True
 
+#### history_write_step_interval
+
+(integer, optional)
+
+The history file can be written incrementally in case of a crash.  If this value is greater than 0, the history file will be rewritten every N steps where N is the value of the parameter, always writing before step 1 as well.  If zero or negative, incremental writes will be turned off and the history file will only be written on successful completion.  Default: 1
```
ThomasSchellenbergNextCentury, line 276: The precedent we had established (though maybe it will change with Dean's new documentation format) is for config file properties that seem more TA2-focused (stuff the TA1s probably won't use) to be documented in the DEV doc:
https://github.com/NextCenturyCorporation/MCS/blob/development/machine_common_sense/DEV.md#full-list-of-config-options

machine_common_sense/history_writer.py
```diff
@@ -48,7 +48,7 @@ def __init__(self, scene_config_data=None, hist_info={}, timestamp=''):
     def write_file(self):
         if self.scene_history_file:
             logger.info(f"Saving history file {self.scene_history_file}")
-            with open(self.scene_history_file, "a+") as history_file:
+            with open(self.scene_history_file, "w+") as history_file:
```
ThomasSchellenbergNextCentury, line 51: Won't this completely overwrite the scene history file each time?

kdrumm: Yeah, that's the point.  It seemed the best balance of performance, simplicity, and effectiveness.  I sacrificed performance for the other two, but I tried to mitigate that with the step interval property.

README.md
kdrumm: Got it, I'll move this.

machine_common_sense/history_writer.py
ThomasSchellenbergNextCentury: Yeah, but, we're only saving the output from the latest step each time, right? So instead of the file listing all the step outputs, we'll only ever have the last one saved? Or am I missing something here?

kdrumm: No, it writes the every step at that moment.  (Someone should verify that I'm not blind, even though I did test it)

kdrumm:
Closing the pull request as the fix is not performant enough and the issue may be OBE.

PR #322, opened 2021-04-28
MCS-575:  Headless Integration Tests

brianpippin:


machine_common_sense/controller.py
```diff
@@ -15,6 +15,8 @@
 import ai2thor.controller
 import ai2thor.server
 
+print(ai2thor.controller.__file__)
```
brianpippin, line 18: Was for debugging.

machine_common_sense/controller.py
```diff
@@ -195,7 +197,7 @@ def __init__(self, unity_app_file_path, config_file_path=None):
             quality='Medium',
             fullscreen=False,
             # The headless flag does not work for me
-            headless=False,
+            headless=True,
```
ThomasSchellenbergNextCentury, line 200: Not sure if we want this to default to True. Could we add a headless option to the config file?

machine_common_sense/controller.py
```diff
@@ -1186,7 +1188,9 @@ def save_images(self, scene_event, max_depth):
         object_mask_list = []
 
         for index, event in enumerate(scene_event.events):
-            scene_image = PIL.Image.fromarray(event.frame)
+            scene_image = None
+            if event.frame is not None:
```
ThomasSchellenbergNextCentury, line 1192: We may only want the `if` here if we're actually running headlessly, since otherwise I think it would be preferable to throw an error if an image isn't actually returned. So we could say something like `if event.frame and self._config.is_headless():` (same with the `if` statements for the depth and instance segmentation frames).

(You also don't strictly need the `is not None` here)

machine_common_sense/controller.py
brianpippin: All of this is just test code to get the integration tests working in Headless, I haven't gotten to adding configs, if statements or anything else, I am just trying to get it to work.  This wasn't meant to be a real code review yet, just for you to see what I was talking about.  I still want to get all the integration tests passing, or if we need to reconfigure some because we are lacking images, or figure out if we can still output some images, etc.

ThomasSchellenbergNextCentury: Oh, apologies -- I misunderstood. Thanks for the explanation.

PR #323, opened 2021-04-28
Moved getch to scripts

deanwetherby:


machine_common_sense/__init__.py
```diff
@@ -19,7 +19,6 @@
 from .history_writer import HistoryWriter
 from .step_metadata import StepMetadata
 from .util import Util
-from .getchHelper import getch
```
deanwetherby, line 22: There's no need for getch in the MCS package as it is only used for the human input script.

scripts/run_human_input.py
```diff
@@ -3,6 +3,8 @@
 
 import machine_common_sense as mcs
 
+from getchHelper import getch
```
deanwetherby, line 6: importing getch directly from the module rather than mcs namespaced.

ThomasSchellenbergNextCentury, line 6: While we're at it, should we rename this file to `getch_helper.py` to follow python conventions?

deanwetherby: Bam! Absolutely

rartiss55: I think getchHelper needs to be updated to getch_helper since it got renamed?

brianpippin:
Do we want to update the getch_helper file to use our python naming conventions as well or is that file just something we are using?

deanwetherby:
> Do we want to update the getch_helper file to use our python naming conventions as well or is that file just something we are using?

In what way?

brianpippin:
> > Do we want to update the getch_helper file to use our python naming conventions as well or is that file just something we are using?
> 
> In what way?

I think I was looking at class names, and forgot those can be camelcase, so feel free to just ignore my comment.

kdrumm (approved):

PR #324, opened 2021-04-30
MCS-638 Fixed links in python classes to API doc.

ThomasSchellenbergNextCentury:


deanwetherby (commented):
Thanks for the backup. I just got intersphinx working so it's now a question of standardization of :mod:`Action <machine_common_sense.Action>`  vs this approach. Sphinx does allow referencing modules, classes, functions and attributes which may be better to stick with.

PR #326, opened 2021-04-30
[DRAFT] Mcs 642 controller refactor part 1

kdrumm:
Note: This is not a final pull request for this issue.  There are a bunch of things that are known to be not quite right, but left to reduce the amount of things we're changing at once.

Goals of this pull request:

- Add listener pattern (I don't think its pub/sub since the controller only pushes specific events).
- Add abstract subscriber 
- Move video creation code to a single video creation listener (later to be separated)
- Move output logging to listener

Problems that still exist and won't be address in this part:
- payload of the listen is large data dumpster that needs to help
- Do we need image/video data in the output of step()?
- Need appropriate way (dependency injection?) initialize subscribers
- print output is somewhat weird when doing run_human_input
- Other smaller things that I forgot and hopefully have a code comment

### DIFF
--- a/machine_common_sense/controller.py
+++ b/machine_common_sense/controller.py
@@ -181,13 +181,14 @@ class Controller():
 
     def __init__(self, unity_app_file_path, config_file_path=None):
 
-        # Can we rearrange to use dependency injection?
-        self._subscribers = [ControllerLogger()]
+        self._subscribers = []
 
         self._end_scene_not_registered = True
 
         self._config = ConfigManager(config_file_path)
 
+        # Can we rearrange to use dependency injection?
+        self.subscribe(ControllerLogger())
         if (self._config.is_evaluation() or self._config.is_video_enabled()):
             self.subscribe(ControllerVideoManager())
 
--- a/machine_common_sense/controller.py
+++ b/machine_common_sense/controller.py
@@ -41,8 +41,7 @@ from .history_writer import HistoryWriter
 from .config_manager import ConfigManager
 from .controller_output_handler import ControllerOutputHandler
 from .controller_logger import ControllerLogger
-from .event_type import EventType
-from .controller_event_payload import ControllerEventPayload
+from .controller_events import ControllerEventPayload, EventType
 from .controller_video_manager import ControllerVideoManager
 
 
--- a/machine_common_sense/controller.py
+++ b/machine_common_sense/controller.py
@@ -30,12 +30,10 @@ DEFAULT_MOVE = 0.1
 from .action import Action
 from .goal_metadata import GoalMetadata
 from .object_metadata import ObjectMetadata
-from .plotter import TopDownPlotter
 from .pose import Pose
 from .return_status import ReturnStatus
 from .scene_history import SceneHistory
 from .step_metadata import StepMetadata
-from .recorder import VideoRecorder
 from .reward import Reward
 from .uploader import S3Uploader
 from .util import Util
@@ -45,6 +43,7 @@ from .controller_output_handler import ControllerOutputHandler
 from .controller_logger import ControllerLogger
 from .event_type import EventType
 from .controller_event_payload import ControllerEventPayload
+from .controller_video_manager import ControllerVideoManager
 
 
 def __reset_override(self, scene):
@@ -112,9 +111,6 @@ class Controller():
     # really matter because we set continuous to True in
     # the step input.)
     GRID_SIZE = 0.1
-    # Is there any way to get FPS from ai2thor?
-    # 20 FPS relies on the assumption that Time.deltaTime is 0.05 in Unity
-    FPS_FRAME_RATE = 20
 
     DEFAULT_CLIPPING_PLANE_FAR = 15.0
 
@@ -186,11 +182,15 @@ class Controller():
 
     def __init__(self, unity_app_file_path, config_file_path=None):
 
+        # Can we rearrange to use dependency injection?
         self._subscribers = [ControllerLogger()]
 
         self._end_scene_not_registered = True
 
         self._config = ConfigManager(config_file_path)
+
+        if (self._config.is_evaluation() or self._config.is_video_enabled()):
+            self.subscribe(ControllerVideoManager())
 
         self._update_screen_size()
 
@@ -223,23 +223,28 @@ class Controller():
 
     def _publish_event(self, event_type: EventType,
                        payload: ControllerEventPayload):
-        if (hasattr(self, '_subscribers')):
-            for subscriber in self._subscribers:
-                # TODO should we make a deep copy of the payload so subscribers
-                # cannot change source data?
-                # seems like performance vs safety
+        for subscriber in self._subscribers:
+            # TODO should we make a deep copy of the payload so subscribers
+            # cannot change source data?
+            # seems like performance vs safety
 
-                # TODO passing the whole controller is clearly a bad idea
-                # change later
-                subscriber.on_event(event_type, payload, self)
+            # TODO passing the whole controller is clearly a bad idea
+            # change later
+            subscriber.on_event(event_type, payload, self)
 
     def _create_event_payload(self):
         payload = ControllerEventPayload()
+        payload.output_folder = self.__output_folder
         payload.config = self._config
+        payload.timestamp = self.generate_time()
         payload.step_number = self.__step_number
         payload.scene_config = self.__scene_configuration
         payload.habituation_trial = self.__habituation_trial
         payload.goal = self._goal
+        payload.depth_maps_enabled = self.__depth_maps
+        payload.object_masks_enabled = self.__object_masks
+        payload.screen_width = self.__screen_width
+        payload.screen_height = self.__screen_height
         return payload
 
     # Pixel coordinates are expected to start at the top left, but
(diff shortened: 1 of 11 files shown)


machine_common_sense/controller.py
```diff
@@ -182,12 +181,21 @@ class Controller():
...
         self._config = ConfigManager(config_file_path)
 
+        # Can we rearrange to use dependency injection?
+        self.subscribe(ControllerLogger())
+        if (self._config.is_evaluation() or self._config.is_video_enabled()):
+            self.subscribe(ControllerVideoManager())
+
         self._update_screen_size()
 
+        self._output_handler = ControllerOutputHandler(self._config)
```
kdrumm, line 197: This is a WIP component that isn't being used yet.

machine_common_sense/controller.py
```diff
@@ -208,8 +216,40 @@ def __init__(self, unity_app_file_path, config_file_path=None):
...
+        payload.timestamp = self.generate_time()
+        payload.step_number = self.__step_number
+        payload.scene_config = self.__scene_configuration
+        payload.habituation_trial = self.__habituation_trial
+        payload.goal = self._goal
+        payload.depth_maps_enabled = self.__depth_maps
+        payload.object_masks_enabled = self.__object_masks
+        payload.screen_width = self.__screen_width
+        payload.screen_height = self.__screen_height
+        return payload
```
kdrumm, line 248: This is a mess to be fixed later.

machine_common_sense/controller.py
```diff
@@ -1184,7 +1169,7 @@ def retrieve_structural_object_list(self, scene_event):
                 key=lambda x: x.uuid
             )
 
-    def save_images(self, scene_event, max_depth):
+    def save_image_data(self, scene_event, max_depth):
```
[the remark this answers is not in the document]
deanwetherby: Yes. Used by TA1 through the step metadata dictionary.

machine_common_sense/controller_events.py
```diff
@@ -0,0 +1,56 @@
...
+
+import enum
+
+
+class EventType(enum.Enum):
+    ON_INIT = 1
+    ON_START_SCENE = 2
+    ON_BEFORE_STEP = 3
+    ON_AFTER_STEP = 4
+    ON_END_SCENE = 5
```
ThomasSchellenbergNextCentury, line 11: FYI If the specific numbers aren't important, you can use `auto`: https://docs.python.org/3/library/enum.html#using-automatic-values

machine_common_sense/controller_events.py
```diff
@@ -0,0 +1,56 @@
...
+class AbstractControllerSubscriber(ABC):
+
+    def __init__(self):
+        self._switcher = {
+            EventType.ON_INIT: self.on_init,
+            EventType.ON_START_SCENE: self.on_start_scene,
+            EventType.ON_BEFORE_STEP: self.on_before_step,
+            EventType.ON_AFTER_STEP: self.on_after_step,
+            EventType.ON_END_SCENE: self.on_end_scene,
+            "default": self.on_default
```
ThomasSchellenbergNextCentury, line 28: What is the purpose of "default"? If it's important, should it be its own EventType?

machine_common_sense/controller.py
ThomasSchellenbergNextCentury: Yeah, we might want separate classes for the different types of payloads? And we can put them in their own files too.

machine_common_sense/controller_video_manager.py
```diff
@@ -0,0 +1,146 @@
...
+        # strip prefix in scene_name
+        if '/' in scene_name:
+            scene_name = scene_name.rsplit('/', 1)[1]
+
+        basename_template = '_'.join(
+            [eval_name, payload.config.get_metadata_tier(), team, scene_name,
+             controller.PLACEHOLDER, timestamp]) + '.mp4'
+
+        visual_video_filename = basename_template.replace(
+            controller.PLACEHOLDER, controller.VISUAL)
```
ThomasSchellenbergNextCentury, line 43: If you `import` our `Controller` class, I think you can just say `Controller.VARIABLE` here. Since it looks like you're only using the `controller` argument in these functions to reference global variables, I believe you could remove the `controller` argument entirely this way (unless you need it for something else later).

If that doesn't work, you could just stick all these variables in the `payload`, if you want to eliminate the `controller` argument.

machine_common_sense/controller.py
```diff
@@ -208,8 +216,41 @@ def __init__(self, unity_app_file_path, config_file_path=None):
...
+    def _publish_event(self, event_type: EventType,
+                       payload: ControllerEventPayload):
+        for subscriber in self._subscribers:
+            # TODO should we make a deep copy of the payload so subscribers
+            # cannot change source data?
+            # seems like performance vs safety
+
+            # TODO passing the whole controller is clearly a bad idea
+            # change later
+            subscriber.on_event(event_type, payload, self)
```
ThomasSchellenbergNextCentury, line 233: Rather than each subscriber being automatically subscribed to ALL event types, and having no-op functions for some types, should the subscriber only be subscribed to event types it cares about?

machine_common_sense/controller.py
```diff
@@ -208,8 +216,41 @@ def __init__(self, unity_app_file_path, config_file_path=None):
...
+        if subscriber not in self._subscribers:
+            self._subscribers.append(subscriber)
+            # TODO consider sending init if init has alreayd happened?
+
+    def _publish_event(self, event_type: EventType,
+                       payload: ControllerEventPayload):
+        for subscriber in self._subscribers:
+            # TODO should we make a deep copy of the payload so subscribers
+            # cannot change source data?
+            # seems like performance vs safety
```
ThomasSchellenbergNextCentury, line 229: I want to say we should `copy.deepcopy`, but that will take a very long time on the depth data. Hmm...

machine_common_sense/controller.py
```diff
@@ -208,8 +216,41 @@ def __init__(self, unity_app_file_path, config_file_path=None):
 
         self._on_init(config_file_path)
 
+    def subscribe(self, subscriber):
+        if subscriber not in self._subscribers:
+            self._subscribers.append(subscriber)
+            # TODO consider sending init if init has alreayd happened?
```
ThomasSchellenbergNextCentury, line 222: I think it's OK to assume the subscriber has already been initialized.

machine_common_sense/controller.py
```diff
@@ -406,6 +387,7 @@ def end_scene(self, choice, confidence=1.0):
                                  '_' + history_filename)
                 )
 
+            # Need to figure out how to connect "uploader" and "video manager"
             topdown_filename = self._get_filename_without_timestamp(
```
ThomasSchellenbergNextCentury, line 391: At the risk of observable inception, could the video manager publish a separate set of events and the uploader subscribes to those events?

machine_common_sense/controller.py
kdrumm: This is meant to be that if we change how subscribers actually subscribe (aka outside of controller and thus completely decouple subscribers from the controller), it is theoretically possible they could miss onInit or even onStartScene events.  This also is more than likely only a theoretical problem.

machine_common_sense/controller.py
```diff
@@ -815,10 +795,16 @@ def step(self, action: str, **kwargs) -> StepMetadata:
...
-        output = self.restrict_step_output_metadata(
-            copy.deepcopy(pre_restrict_output))
+        self.write_debug_output_file(output)
 
-        self.write_debug_output(output)
+        payload = self._create_event_payload()
+        # do we need both outputs?
+        payload.pre_restrict_output = pre_restrict_output
+        payload.step_metadata = step_output
+        payload.step_output = output
```
ThomasSchellenbergNextCentury, line 804: FWIR one is needed for the scene history and debug output, and the other is needed for the TA1s. A reason why it may be useful to have separate payloads for separate subscribers.

machine_common_sense/controller.py
ThomasSchellenbergNextCentury: We'll probably want to identify exactly the different variables needed by the different subscribers, and have them defined as individual class properties on each Payload class.

machine_common_sense/controller.py
kdrumm: This is a "figure out later" part of this "rolling merge", but it is definitely worthwhile to discuss.  Your idea is certainly an option that seems reasonable to me.  It does mean we also need to figure out how does a subscriber find the components that it can subscribe to?

machine_common_sense/controller_events.py
kdrumm: Thanks!

machine_common_sense/controller_events.py
kdrumm: its really just a catch for someone doing something they shouldn't.

machine_common_sense/controller.py
kdrumm: This certainly needs work, but with this model we can't use different variables for different subscribers.  That breaks the decoupling.  If we remove the observer pattern, we could certainly pass individual payload classes to each component.  We would just be more coupled and lose the ease of adding new subscribers whenever needed.

ThomasSchellenbergNextCentury: Oh, you're right. I must have gotten myself mixed up -- my bad. I think this is what I meant: Should there be separate payload classes for each event type (OnInit, OnStartScene, etc.)? Maybe with a superclass for common variables?

machine_common_sense/controller.py
ThomasSchellenbergNextCentury: Oh, I see what you're saying. I thought you were talking about the subscriber class's `__init__` function (constructor). Your question makes a lot more sense now!

machine_common_sense/controller_events.py
ThomasSchellenbergNextCentury: How would this "catch" happen, though? Maybe I'm just being thick-headed, but I can't think of an example situation in which this would be relevant.

machine_common_sense/controller.py
kdrumm: That is an excellent question.  I can argue either way, but I went with the current method because i find it simpler to program and so far it seems the subscribers needed at least 3 of 5.  It is relatively easy to switch the code later as well.

machine_common_sense/controller_events.py
kdrumm: Ok, I agree.  While I can come up with some  contrived ways for this to be called, I removed it.  I can't think of a time we'd want to allow a call to this method where the parameter wasn't one of the events.

machine_common_sense/controller.py
kdrumm: I think you're right there.  I wanted to get more subscribers written before we nailed that down.  (This is the fun part of trying to merge this in multiple parts)

machine_common_sense/controller_video_manager.py
kdrumm: This certainly needs to be looked at.  I'm hoping we can probably move a bunch of the constants later.

machine_common_sense/controller.py
kdrumm: only the regular (restricted) output is sent to TA1.  They call the controller procedurally.

machine_common_sense/controller.py
kdrumm: I'll change the comment to be clear (it was admittantly pretty vague).

kdrumm (commented):
This PR mostly address video, image, json output coupling.  The goal is for the output to look identical to what it would before.  The main ways to test is to turn videos, images, and json on and off and verify they exist as appropriate and they look right.  

Additionally, I ran the same execution in the dev branch and this branch, saving the output folders in a separate locations.  From there I ran a diff on the two different folders and verified any differences make sense.  I also looked at the videos and images and made sure they looked identical.

ThomasSchellenbergNextCentury:
@kdrumm The `save_debug_images` config file property isn't working for me. It should save the `frame_<step>` images in the output folder.

kdrumm:
I think I found the bug, but it weirdly caused my images to ALWAYS be written.

kdrumm:
@ThomasSchellenbergNextCentury what config level were you using?

ThomasSchellenbergNextCentury:
@kdrumm I had tried with both `none` and `oracle`

machine_common_sense/controller.py
```diff
@@ -182,12 +181,21 @@ class Controller():
...
+        self._subscribers = []
+
         self._end_scene_not_registered = True
 
         self._config = ConfigManager(config_file_path)
 
+        # Can we rearrange to use dependency injection?
+        self.subscribe(ControllerLogger())
+        if (self._config.is_evaluation() or self._config.is_video_enabled()):
+            self.subscribe(ControllerVideoManager())
```
rartiss55, line 193: Regarding the images not being rendered when using the save_debug_images property, should the check for is_save_debug_images be included here + additional logic in the video manager to not necessarily render the videos in that case? Or perhaps that case means the debug images should go into its own handler, not sure what the right solution is.

kdrumm: They should be all separated into their own handler.  This was supposed to be a stepping stone solution, but you point out a valid reason to avoid merging this as it is.

PR #327, opened 2021-05-03
MCS-641 natural plot sizes

deanwetherby:
The video recorder now establishes a size (width, height) from the first provided frame. All future video frames are compared against that established size and a ValueError is raises if the size varies in any dimension. The plotter naturally produces images sizes of 640x480 for the default DPI setting which is different from our rendered frames of 600x400.

### DIFF
--- a/machine_common_sense/plotter.py
+++ b/machine_common_sense/plotter.py
@@ -41,14 +41,11 @@ class TopDownPlotter():
     MAXIMUM_ROOM_DIMENSION = 5
     BORDER = 0.05
 
-    def __init__(self, team: str, scene_name: str,
-                 plot_width: int, plot_height: int):
+    def __init__(self, team: str, scene_name: str):
         self._team = team
         if '/' in scene_name:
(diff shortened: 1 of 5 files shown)


ThomasSchellenbergNextCentury (approved):
LGTM. Nice to see the unit tests as well. Will this need a corresponding ticket for the UI? I'm not sure if we make any assumptions about the video size in the UI.

machine_common_sense/plotter.py
```diff
@@ -41,14 +41,11 @@ class TopDownPlotter():
     MAXIMUM_ROOM_DIMENSION = 5
     BORDER = 0.05
 
-    def __init__(self, team: str, scene_name: str,
-                 plot_width: int, plot_height: int):
+    def __init__(self, team: str, scene_name: str):
```
kdrumm, line 44: Mine crashes in controller.py line 516 because the argument list hasn't been updated.

deanwetherby: Missed that one

kdrumm (approved):

PR #328, opened 2021-05-03
[Python] Update master to 0.4.2

ThomasSchellenbergNextCentury:
Each commit should have been individually reviewed before, so no need to go through all of them now, unless you want to.


kdrumm:
I noticed today in other testing that our output is currently pushing:

    "image_list": [<PIL.Image.Image image mode=RGB size=600x400 at 0x7F68C80D2940>],
    "object_mask_list": [<PIL.Image.Image image mode=RGB size=600x400 at 0x7F68C80D2F10>],

I assume this is a bug.  Do we want to fix it before we release?

ThomasSchellenbergNextCentury:
@kdrumm Both of those properties are supposed to be lists of Pillow images as per our [docs](https://github.com/NextCenturyCorporation/MCS/blob/master/machine_common_sense/API.md#stepmetadata). Is there something specific concerning you?

deanwetherby (approved):

kdrumm:
@ThomasSchellenbergNextCentury It seems very odd to put that kind of data in a json file.  We aren't actually returning the image data, but instead how the class instance gets turned into a string by the programming language.  I think it also makes it invalid json.

ThomasSchellenbergNextCentury:
@kdrumm Well, it's not supposed to be going into JSON files -- you're right, that's invalid. It should be returned to the TA1 python code as output from the `start_scene` and `step` functions, but it's not supposed to be saved in any other JSON files. Which files are problematic? The scene history files?

kdrumm:
@ThomasSchellenbergNextCentury mcs_output_#.json where # is the step number.

ThomasSchellenbergNextCentury:
@kdrumm I made a PR to fix this issue here: https://github.com/NextCenturyCorporation/MCS/pull/329

Let me know if you think this is a good solution to the issue. If so, we can merge it and include it in this release. Thanks.

PR #329, opened 2021-05-03
Remove Pillow image lists and depth data from JSON debug output files.

ThomasSchellenbergNextCentury:
Fixes https://github.com/NextCenturyCorporation/MCS/pull/328#issuecomment-831392038

### DIFF
--- a/machine_common_sense/API.md
+++ b/machine_common_sense/API.md
@@ -349,7 +349,7 @@ Defines metadata for an object in the MCS 3D environment.
 ## StepMetadata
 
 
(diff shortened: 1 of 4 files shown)


kdrumm (approved):
Previously we had depth_map values (a huge array of floats that seemed to be 0 - 1).  Those are also gone.  It seems like that is a good thing, but I wanted to point that out to make sure.

ThomasSchellenbergNextCentury:
@kdrumm Fair observation. That's intentional. The depth data was already being deleted for the scene history files, and I did the same for the debug files because saving the depth data takes a significant amount of time and hard disk space. I think we (TA2) are the main consumers of the debug files anyway, but if anyone from TA1 has a concern, we can always reevaluate the behavior (IMO).

brianpippin (approved):

PR #331, opened 2021-05-11
Mcs 642 controller refactor part 2

kdrumm:
Note: there are a couple tickets to continue this later after this is merged:

- MCS-663 - Addresses the payload
- MCS-664 - Separate ControllerVideoManager
- MCS-665 - Continue rework on ControllerOutputHandler
- MCS-666 - Rework how controller initializes subscribers (dependency injection?)
- MCS-667 - Remove controller instance from controller events
- MCS-668 - Continue refactor of controller (handles input handling and moving more code to concrete classes)

Ways to test:
- Run development and this branch and verify history, json, output are the same (using diff is useful)
- Need to test an evaluation to verify when evaluation is set to true

I have a script that will run a test on two different branches, copy the output and scene history into folders and do a diff.  It'll run it one time per config file in a folder.  It is FAR from production ready, but may be nice to have for this or future tasks.  Let me know if you'd like me to send it to you.


machine_common_sense/config_manager.py
```diff
@@ -175,3 +201,56 @@ def is_video_enabled(self):
...
+    def get_screen_width(self) -> int:
+        return self.get_size()
+
+    def get_screen_height(self) -> int:
+        size = self.get_size()
+        return int(size / 3 * 2)
+
+
+# discrete class to be?
+class SceneConfiguration:
```
ThomasSchellenbergNextCentury, line 233: Would it be better to move `controller.retrieve_goal()` into here as well? I'm assuming this class is used for retrieving important data from the `controller.__scene_configuration`? (For some of your new classes, it might be helpful to add a class block comment, using `"""Comment"""` syntax under the `class Whatever():` line.)

machine_common_sense/config_manager.py
```diff
@@ -175,3 +201,56 @@ def is_video_enabled(self):
...
+    def __init__(self, scene_config_dict):
```
ThomasSchellenbergNextCentury, line 234: Will `scene_config_dict` be needed here, or will this class only ever have static methods? (I'm going to ask a lot of questions like this, and a lot of your answers may be "this is needed for a future PR", so please excuse me!)

machine_common_sense/controller.py
```diff
@@ -208,71 +191,61 @@ def __init__(self, unity_app_file_path, config_file_path=None):
 
         self._on_init(config_file_path)
 
+    def subscribe(self, subscriber):
+        if subscriber not in self._subscribers:
```
ThomasSchellenbergNextCentury, line 195: How does this work in Python? Is each property in `subscriber` compared to each property in each element of `self._subscribers`? Is there an equivalent of a `compareTo` function that we need to implement?

machine_common_sense/config_manager.py
kdrumm: I agree.  The whole Scene Configuration discrete class needs to be done, but decided to do it as a separate ticket.  Its actually in MCS-668 although that ticket probably needs a rename.  I'll add some class comments as well.

machine_common_sense/controller.py
```diff
@@ -359,82 +287,24 @@ def end_scene(self, choice, confidence=1.0):
...
+        payload = self._create_event_payload()
+        payload.choice = choice
+        payload.confidence = confidence
+        payload.uploader = uploader
+        payload.uploader_folder_prefix = self._config.get_s3_folder()
+        self._publish_event(EventType.ON_END_SCENE, payload)
 
-    def _get_filename_without_timestamp(self, filepath: pathlib.Path):
-        return filepath.stem[:-16] + filepath.suffix
+        if (self._end_scene_not_registered is False):
```
ThomasSchellenbergNextCentury, line 305: FYI This can just be `if not self._end_scene_not_registered:`

machine_common_sense/controller.py
```diff
@@ -457,69 +327,26 @@ def start_scene(self, config_data):
...
+        wrapped_step = self.wrap_step(
+            action='Initialize', sceneConfig=config_data)
+        step_output = self._controller.step(wrapped_step)
 
-        output = self.restrict_step_output_metadata(
-            copy.deepcopy(pre_restrict_output))
-
-        self.write_debug_output(debug_copy)
+        self._output_handler.set_scene_config(config_data)
+        (pre_restrict_output, output) = self._output_handler.handle_output(
```
ThomasSchellenbergNextCentury, line 347: FYI You don't need parentheses around the two output variables here.

machine_common_sense/controller.py
```diff
@@ -539,26 +366,38 @@ def start_scene(self, config_data):
...
-                ):
-                    self.__image_recorder.add(image_list[0])
-
                 output.image_list = image_list
                 output.depth_map_list = depth_map_list
                 output.object_mask_list = object_mask_list
 
             logger.debug('NO PREVIEW PHASE')
 
+            # Why is this in the if black?  should it be?
```
ThomasSchellenbergNextCentury, line 375: I'm not sure about your question, but I'd recommend adding a TODO or FIXME here so it's easy to search for or identify later.

machine_common_sense/controller_events.py
```diff
@@ -0,0 +1,111 @@
...
+                 internal_state: object = None):
+        self.config = config
+        self.choice = choice
+        self.confidence = confidence
+        self.violations_xy_list = violations_xy_list
+        self.heatmap_img = heatmap_img
+        self.internal_state = internal_state
+
+
+class AbstractControllerSubscriber(ABC):
```
ThomasSchellenbergNextCentury, line 75: It looks like you've made this an abstract class but aren't using the `@abstractmethod` decorator on any functions? Is that intentional?

machine_common_sense/config_manager.py
kdrumm: Ask away.  its better we are careful with this.

This is more or less a placeholder for future work.  MCS-668.  Of course, I could have left out the initialization for now.

machine_common_sense/controller_events.py
```diff
@@ -0,0 +1,111 @@
...
+class BaseEventPayload:
+
+    def __init__(self, step_number: int):
+        self.step_number = step_number
+
+
+class ControllerEventPayload(BaseEventPayload):
+    def __init__(self, output_folder: str, config: ConfigManager,
+                 step_number: int, scene_config: SceneConfiguration,
+                 habituation_trial, goal):
```
ThomasSchellenbergNextCentury, line 29: `habituation_trial` is an `Optional[int]` and `goal` is a `Dict`

machine_common_sense/controller.py
kdrumm: It verifies that we don't have duplicate instances of the same subscriber.  (I.E. the 'Sub' class could have instanace 'inst1' and 'inst2'.  inst1 and inst2 can both be subscribers, but inst1 can't subscribe twice).

machine_common_sense/controller_events.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def on_event(self, type: EventType,
+                 payload: ControllerEventPayload, controller):
+        self._switcher.get(
+            type,
+            "default")(
+            payload,
+            controller)
+
+    def on_init(self, payload, controller):
```
ThomasSchellenbergNextCentury, line 95: Is the `payload` in these functions intentionally untyped?

machine_common_sense/controller_logger.py
```diff
@@ -0,0 +1,110 @@
...
+
+from .controller_events import AbstractControllerSubscriber
+from .util import Util
+
+logger = logging.getLogger(__name__)
+
+
+class ControllerLogger(AbstractControllerSubscriber):
+
+    # def on_init(self, payload, controller):
```
ThomasSchellenbergNextCentury, line 12: Either delete these commented-out lines or add a TODO/FIXME comment above them (with the relevant JIRA ticket label, if there is one).

machine_common_sense/controller.py
kdrumm: fixed (in next commit)

machine_common_sense/controller.py
kdrumm: Is there a convention or best practice for this?  Or just personal preference?

machine_common_sense/controller.py
kdrumm: I'll change the comment to remove the typo and add TODO

machine_common_sense/controller_events.py
kdrumm: I want the methods to be optionally overridable (not forced), but the class itself cannot be instanced.  Hopefully my understanding of pythons abstract class/methods are accurate.

machine_common_sense/controller_events.py
kdrumm: somewhat.  payload currently is a single class, but there is a ticket to investigate payloads and split them into separate classes per event (MCS-663).    There is also a ticket to remove the controller from the events (MCS-667).  I can type them for now.

machine_common_sense/controller_logger.py
kdrumm: removed. thanks for catching

machine_common_sense/controller.py
ThomasSchellenbergNextCentury: I think it's just personal preference, so don't feel the need to change it. I just mentioned it in case you didn't know already!

machine_common_sense/controller.py
ThomasSchellenbergNextCentury: I meant: how does the comparison work between inst1 and inst2 when they are both classes? If the answer is python magic, then that's fair enough.

machine_common_sense/controller_events.py
ThomasSchellenbergNextCentury: Sounds good to me!

machine_common_sense/config_manager.py
ThomasSchellenbergNextCentury: That's fine. Sometimes people (myself included) accidentally leave stale code from when they tried implementing something one way, and then decide to do something completely different, so I figured I'd double check.

machine_common_sense/controller_output_handler.py
```diff
@@ -0,0 +1,402 @@
...
+from .step_metadata import StepMetadata
+from .pose import Pose
+from .return_status import ReturnStatus
+from .object_metadata import ObjectMetadata
+from .util import Util
+from .config_manager import ConfigManager, SceneConfiguration
+
+
+# from .reward import Reward
+# from .step_metadata import StepMetadata
```
ThomasSchellenbergNextCentury, line 17: Are these imports needed?

machine_common_sense/controller_output_handler.py
```diff
@@ -0,0 +1,402 @@
...
+
+    def set_scene_config(self, scene_config):
+        self._scene_config = scene_config
+
+    def handle_output(self, raw_output, goal, step_number, habituation_trial):
+        # output_copy = copy.deepcopy(raw_output)
+        # pre_restrict_output = self.wrap_output(
+        #    output_copy, goal, step_number, habituation_trial)
+        # output = self.restrict_step_output_metadata(pre_restrict_output)
+        # return (pre_restrict_output, output)
```
ThomasSchellenbergNextCentury, line 36: Is this code needed (or will it be needed in the future)?

machine_common_sense/controller_output_handler.py
```diff
@@ -0,0 +1,402 @@
...
+                    if 'id' in step_output.goal.metadata[target_name]:
+                        step_output.goal.metadata[target_name]['id'] = None
+                    if 'image_name' in step_output.goal.metadata[target_name]:
+                        step_output.goal.metadata[
+                            target_name]['image_name'] = None
+
+        return step_output
+
+
+class StepOutput:
```
ThomasSchellenbergNextCentury, line 89: We should clarify how this class differs from `StepMetadata`, since that's the output given to the TA1s from the `step` call. Maybe renaming the class and/or adding a class comment.

machine_common_sense/controller_video_manager.py
```diff
@@ -0,0 +1,219 @@
...
+            if payload.config.is_object_masks_enabled():
+                if config.is_evaluation() or config.is_video_enabled():
+                    self.__segmentation_recorder.add(object_mask)
+                if (payload.output_folder and
+                        payload.config.is_save_debug_images()):
+                    suffix = self._get_suffix(payload, index)
+                    object_mask.save(fp=payload.output_folder +
+                                     'object_mask' + suffix)
+
+# if payload.output_folder and payload.config.is_save_debug_images:
```
ThomasSchellenbergNextCentury, line 205: What's up with this line?

machine_common_sense/config_manager.py
```diff
@@ -3,12 +3,31 @@
...
 
+from .action import Action
 
 logger = logging.getLogger(__name__)
 
 
 class ConfigManager(object):
 
+    DEFAULT_CLIPPING_PLANE_FAR = 15.0
+    ACTION_LIST = [(item.value, {}) for item in Action]
```
ThomasSchellenbergNextCentury, line 14: Would `ACTION_LIST` be better moved to `action.py`? (I'm not sure myself.)

machine_common_sense/controller.py
```diff
@@ -182,20 +155,30 @@ class Controller():
...
         self._config = ConfigManager(config_file_path)
 
-        self._update_screen_size()
+        # Can we rearrange to use dependency injection?
+        if self._config.is_save_debug_json():
+            self.subscribe(ControllerDebugFileGenerator())
+            self.subscribe(ControllerAi2thorFileGenerator())
+        self.subscribe(ControllerVideoManager())
+        self.subscribe(ControllerLogger())
+        self.subscribe(HistoryEventHandler())
```
ThomasSchellenbergNextCentury, line 170: As our internal dependency tree grows within the MCS library, we should create and post a diagram somewhere for the development team (even if it's just a photo of a whiteboard drawing) that shows all of the different subscriber classes and how they each depend on one another.

machine_common_sense/controller.py
kdrumm: yeah.. python magic.  I just re-tested it too.

machine_common_sense/controller_output_handler.py
kdrumm: Nope, removing.  I was using it as reference and I think i got so used it it, i almost done see it anymore.

machine_common_sense/controller_output_handler.py
kdrumm: will add a comment to ensure it gets addressed in MCS-665 refactor, but I agree the name is somewhat misleading.

machine_common_sense/controller_video_manager.py
kdrumm: Removing.  I think it was from copying code that i turned into function and then commented out what I didn't need.

machine_common_sense/controller.py
kdrumm: The subscriber classes shouldn't depend on each other at all and shouldn't need to know anything about each other.  They especially cannot need another to be subscribed.  The goal here is to decouple all these pieces with a loose, but generic connection in the subscription.  The particular code that you highlighted is the last piece of that goal and should be addressed in MCS-666.  I'm not sure how to do dependency injection in python or with our current interface to TA1.

machine_common_sense/controller_logger.py
```diff
@@ -0,0 +1,114 @@
...
+        logger.debug(
+            "METADATA TIER: " +
+            payload.config.get_metadata_tier())
+        logger.debug(f"STEP: {payload.step_number}")
+        logger.debug("ACTION: Initialize")
+
+        self._write_debug_output(payload)
+
+    def on_before_step(self, payload, controller):
+        logger.info("before step")
```
ThomasSchellenbergNextCentury, line 30: Is this line needed?

machine_common_sense/controller_logger.py
```diff
@@ -0,0 +1,114 @@
...
+        elif payload.goal.habituation_total > 0:
+            logger.debug("HABITUATION TRIAL: DONE")
+        else:
+            logger.debug("HABITUATION TRIAL: NONE")
+
+    def on_after_step(self, payload, controller):
+        self._write_debug_output(payload)
+
+    def _write_debug_output(self, payload):
+        step_output = payload.step_output
```
ThomasSchellenbergNextCentury, line 48: @kdrumm This is supposed to be done on the pre-restricted output, but it isn't, for some reason. For example, when I use level 1 metadata, the object data shouldn't be printed to the terminal, but it is.

machine_common_sense/controller_logger.py
kdrumm: absolutely not!

machine_common_sense/controller_logger.py
kdrumm: You are correct.  It is now using the restricted output as it is supposed to (at least development is).  I renamed them to step_output and restricted_step_output for the event payload.  It seems those names made more sense with which one is actually modified.

machine_common_sense/controller_output_handler.py
kdrumm: removed

machine_common_sense/controller.py
ThomasSchellenbergNextCentury: Ah, OK, I misunderstood. Thanks for the clarification.

If I'm understanding now, could you do this: In `AbstractControllerSubscriber.__init__`, accept a `controller: Controller` constructor argument, and call `controller.subscribe(self)` in the constructor? Or are you talking about other dependencies here, besides subscription?

machine_common_sense/controller_logger.py
ThomasSchellenbergNextCentury: Yay! It works now.

ThomasSchellenbergNextCentury (approved):
LGTM. Thanks for fixing the saving-debug-images-to-file issue. I use that feature almost everyday to make example videos for our design team, so it was important for me that it continued working. This all looks like a promising start to the refactor!

machine_common_sense/controller.py
kdrumm: Forgive me if I misunderstand your question and this may be most appropriate to move to voice chat.  

With dependency injection, ideally, the subscribers do very little besides a direction to a response to an event.  They don't even need to know that the event came from the controller.  Also,  our controller doesn't need to know anything about the subscribers or even which ones are subscribed at any given time.  (Granted ares are set by the config and never changed).  Usually, some infrastructure code injects the appropriate subscribers to the controllers maintaining this isolation.  However, with how we interface with TA1, I don't believe we have control of a piece that can create and setup the controller.  (TA1 does that, right?).  We may be able to offload that task to a controller component to some some degree of separation and call it good enough.  To be clear, I am also NOT saying that our interface with TA1 is wrong.

In terms of passing a controller into the subscribers, that increases their coupling and makes the subscribers tied to our controller (which for us is probably only a theoretical problem).  I think (and this is certainly up for debate and more thought), that we would still need the controller to create the subscribers which already makes them more coupled than ideal.

ThomasSchellenbergNextCentury: Ah, you're right. I had reversed it in my head -- My bad. TA1 does create the Controller, but they do so using our [create_controller](https://github.com/NextCenturyCorporation/MCS/blob/development/machine_common_sense/__init__.py#L47) function, so maybe you can use that?

kdrumm: oh!  we can absolutely use that!  Thanks.  I'll update the ticket to mention that.

PR #333, opened 2021-05-13
Mcs 644 - Removed util module

kdrumm:
All unit functions were moved to one of the following:

- to_string.py
- numeric.py
- material.py
- action.py

The numeric and to_string are new.  

to_string:

The one to me that seems that might be most controversial is to_string as this is a java-ism that may not translate as a term to python.  I considered moving these to serializer, but they are functions that convert objects into human readable text and are not intended to be deserialized.  It would be easy to choose another term if desired.

numeric:

This module holds functions that works with managing and testing numbers.  I think its just is_in_range and is_number


scripts/kyle-test.ini
```diff
@@ -0,0 +1,6 @@
+[MCS]
+history_enabled: true
+metadata: level1
+save_debug_images: false
+save_debug_json: true
+video_enabled: false
```
ThomasSchellenbergNextCentury, line 6: Should this file have been committed?

kdrumm: What file?  (thanks for catching that)

machine_common_sense/numeric.py
```diff
@@ -0,0 +1,69 @@
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class Numeric():
```
ThomasSchellenbergNextCentury, line 6: Would Numeric be better renamed to something like Validation instead? The two functions are really just validating the user's input, and it seems like we might want to add more input validation code in the future.

machine_common_sense/to_string.py
```diff
@@ -1,16 +1,14 @@
...
 
-from .action import Action
-from .material import Material
-
-
 logger = logging.getLogger(__name__)
 
 
-class Util:
+class ToString:
```
ThomasSchellenbergNextCentury, line 7: This is super nitpicky, but should this class be named as a noun, like maybe "Stringifier", rather than a verb?

ThomasSchellenbergNextCentury (commented):
Remember to add the new classes to `docs/source/api.rst` and also remove the Util class.

machine_common_sense/numeric.py
deanwetherby: Rather than doing input validation, another (and more pythonic) approach would be to throw exceptions if the input argument is the wrong type or outside the expected range.

kdrumm:
I updated api.rst.  I also changed ToString to Stringifier and Numeric to Validation.  I didn't change the validation to throwing.  I do agree with that method but thought it was out of scope of the intention of this ticket.

deanwetherby (approved):

machine_common_sense/numeric.py
ThomasSchellenbergNextCentury: Then we should make sure to update our documentation for the `start_scene` and `step` functions to list the errors that might be thrown.

ThomasSchellenbergNextCentury (approved):

PR #335, opened 2021-05-13
MCS-666 moved subscriber initialization outside of the controller

kdrumm:
Blame @ThomasSchellenbergNextCentury for this being a smaller, easy PR.

Thanks @ThomasSchellenbergNextCentury for pointing out to me that the controller is constructed in __init__

This change mostly just sets a precedent for later for using dependency injection.

### DIFF
--- a/machine_common_sense/__init__.py
+++ b/machine_common_sense/__init__.py
@@ -8,6 +8,7 @@ import signal
 from contextlib import contextmanager
 
 from .action import Action
+from .config_manager import ConfigManager
 from .controller import Controller
 from .controller_logger import ControllerAi2thorFileGenerator
 from .controller_logger import ControllerDebugFileGenerator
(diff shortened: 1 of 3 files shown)


machine_common_sense/__init__.py
```diff
@@ -65,13 +69,30 @@ def create_controller(unity_app_file_path,
...
+        return controller
     except Exception as Msg:
         logger.error("Exception in create_controller()", exc_info=Msg)
         return None
 
 
+def _add_subscribers(controller):
+    if controller:
+        config = controller._config
+        if config.is_save_debug_json():
```
kdrumm, line 84: I believe this is a python no-no.  Do we want to make it a getter function and give TA1 explicit access?  (my answer no).  Alternatively, we could pull out the code to generate the config and pass it in, ala dependency injection. (my answer yes!).

kdrumm: comment above is OBE

machine_common_sense/__init__.py
```diff
@@ -8,15 +8,20 @@
 from contextlib import contextmanager
 
 from .action import Action
+from .config_manager import ConfigManager
 from .controller import Controller
+from .controller_logger import ControllerAi2thorFileGenerator
+from .controller_logger import ControllerDebugFileGenerator
+from .controller_logger import ControllerLogger
```
ThomasSchellenbergNextCentury, line 15: This can be:

```python
from .controller_logger import ControllerAi2thorFileGenerator, ControllerDebugFileGenerator, ControllerLogger
```

ThomasSchellenbergNextCentury (approved):
> Blame ThomasSchellenbergNextCentury for this being a smaller, easy PR.

Happy to be such a burden :)

PR #340, opened 2021-05-18
MCS-639 image recorder

deanwetherby:
Added an ImageRecorder class that uses threads and a queue to write images to disk. Pulled out common elements from the VideoRecorder into a base class which will make additional recorders fairly easy to implement if needed.


machine_common_sense/recorder.py
```diff
@@ -4,99 +4,93 @@
...
+        if self.recording:
+            logger.debug("Adding item to recorder queue")
+            self._queue.put(item)
+            logger.debug(f"Queue size is approximately {self._queue.qsize()}")
 
-    def finish(self) -> None:
-        '''Deactivate the recorder so that it does not accept more frames.
+    @abstractmethod
+    def _write(self):
+        pass
```
deanwetherby, line 71: The abstract method with an underscore suggests that the implementation not be called directly. This method is used as a callback within the base class.

machine_common_sense/recorder.py
```diff
@@ -4,99 +4,93 @@
...
-        '''
-        self.image_queue = None
-        self.thread = None
-        self.started = False
+    def __init__(self, timeout: float = 1.0):
         self.timeout = timeout
-        self._path = img_path
-        self._frames_written = 0
-
-        self.start()
```
deanwetherby: I'm leaving it up to the derived class implementer whether the thread starts or not for a particular recorder.

machine_common_sense/recorder.py
```diff
@@ -152,44 +132,26 @@ def add(self, frame: PIL.Image.Image) -> None:
...
-                logger.warning("Recorder not active")
-                return
-            if not self.frame_queue.empty():
-                frame = self.frame_queue.get()
-                self.writer.write(frame)
-                self._frames_written = self._frames_written + 1
-                logger.debug(f"Recorder wrote {self._frames_written} frames")
-            else:
-                time.sleep(self.timeout)
+        super().add(frame)
```
deanwetherby, line 148: Because the VideoWriter needs to see the first frame before establishing the size of the video writer, the add method is extended here to deal with those specifics before calling the super class' add method.

machine_common_sense/recorder.py
```diff
@@ -198,12 +160,6 @@ def finish(self) -> None:
         closed.
         '''
         logger.debug("Closing video recorder")
-        self.active = False
-        self.flush()
-        self.thread.join()
+        super().finish()
```
deanwetherby, line 163: Extending the finish method so we can properly close the opencv video writer and hence have to call the super class method here as well.

tests/test_recorder.py
```diff
@@ -5,18 +5,66 @@
...
+from machine_common_sense.recorder import BaseRecorder
 from machine_common_sense.recorder import VideoRecorder
 from machine_common_sense.recorder import ImageRecorder
 
 
+class ConcreteBaseRecorder(BaseRecorder):
+    '''Test implementation of the BaseRecorder'''
+
+    def _write(self):
+        raise NotImplementedError
```
deanwetherby, line 17: A concrete implementation of the base class so we can do testing on those methods. The _write method does nothing as we are not testing it here.

machine_common_sense/recorder.py
```diff
@@ -4,14 +4,93 @@
...
+        self.recording = False
+        self.flush()
+        self._thread.join()
+
+    def write(self):
+        '''Loop forever waiting for frames to enter the queue.'''
+        while True:
+            if not self.recording:
+                logger.warning("Recorder not active")
+                return
```
kdrumm, line 54: Kinda nitpicky: this appears to be the normal exit position of the loop when finish() is called, but it throws a warning.

kdrumm (commented):
Do we have a ticket for adding this to the image handlers?

machine_common_sense/recorder.py
deanwetherby: Recommendations? Remove the log statement or change it to info/debug?

deanwetherby:
> Do we have a ticket for adding this to the image handlers?

We do not. I wanted to get feedback on this approach before we signed off on further work.

machine_common_sense/recorder.py
kdrumm: My thoughts are add a debug there saying something along the lines of recorder going inactive and add another warning before the loop if the method is called while the recorder is not active.

kdrumm:
> We do not. I wanted to get feedback on this approach before we signed off on further work.

I think its good.  It'd be nice to know how it affects performance, but we may have to mostly implement it to actually find out.

kdrumm (approved):

ThomasSchellenbergNextCentury (approved):

PR #342, opened 2021-05-19
Mcs 648 downloader

kdrumm:
This PR removes the need for users to manually provide a MCS Unity executable.  Instead, it will be located and downloaded if necessary in the create_controller function.  All scripts should have an override.

Needs to be tested and updated on Mac

All launch developer launch scripts/commands will need to be updated.  The simple way should be adding "--mcs_unity_build_file_path" in front of the build path (also, verify that is the right argument for the given startup script)


deanwetherby:
So cool.

```python
>>> import machine_common_sense as mcs
>>> controller = mcs.create_controller()
MCS-AI2-THOR-Unity-App-v0.4.3-linux.zip: |####################################################|100%  24.2 MiB/s of 775.MB
```

machine_common_sense/__init__.py
```diff
@@ -71,15 +72,24 @@ def init_logging(log_config=None,
         log_config_file=log_config_file)
 
 
-def create_controller(unity_app_file_path,
+def create_controller(unity_app_file_path=None,
+                      unity_cache_version=None,
                       config_file_path=None):
```
kdrumm, line 77: Should we consider how this change affects TA1?  Should i remove the version?  move it to last so the original parameters are still in the same order?

machine_common_sense/unity_executable_provider.py
```diff
@@ -0,0 +1,269 @@
...
+    def download(self, url: str, filename: str,
+                 destination_folder: Path) -> Path:
+        file_data = self._do_download(url)
+        # todo probably change this
+        file = destination_folder.joinpath(filename)
+        file.write_bytes(file_data)
+        return file
+
+    # TODO should probably change how we download so we can write file
+    # sequentially
```
kdrumm, line 242: This seems like a good idea when downloading 750MB files, but if we're going to change it soon to a much smaller initial download (Addressables), maybe it doesn't matter.

tests/test_unity_executable_provider.py
```diff
@@ -0,0 +1,35 @@
...
+
+class MockExecutionCache(AbstractExecutionCache):
+    def has_version(self, version: str) -> bool:
+        return True
+
+    def _do_zip_to_cache(self, version: str, zip_file: Path):
+        return True
+
+
+class TestUnityExecutableProvider(unittest.TestCase):
```
kdrumm, line 16: Looking for ways to test this.  Most of the functions require it to either download the file or have a zip file

machine_common_sense/ai2thor_provider.py
```diff
@@ -0,0 +1,268 @@
+
+from abc import ABC, abstractmethod
+import logging
+from zipfile import ZipFile
+import tarfile
+import requests
+from progressbar import ProgressBar, Bar, Percentage, FileTransferSpeed
+import platform
+from machine_common_sense.logging_config import LoggingConfig
+from pathlib import Path
```
deanwetherby: Imports should be grouped in the following order:

1. Standard library imports.
2. Related third party imports.
3. Local application/library specific imports.

You should put a blank line between each group of imports.

https://www.python.org/dev/peps/pep-0008/#imports

I'm surprised flake8 or autopep8 didn't rearrange this.

machine_common_sense/__init__.py
deanwetherby: I was thinking the same thing. It might be more likely that a TA1 researcher loads the config path and let's us handle the unity stuff. `mcs.create_controller(config_path)` That won't be possible currently without providing two Nones before the config path. `mcs.create_controoller(None, None, config_path)`

deanwetherby: They could use the named argument though `mcs.create_controller(config_file_path=config_path)`

tests/test_unity_executable_provider.py
deanwetherby: In setUp you could create an empty zip file, run the cache check, and delete it in tearDown.

machine_common_sense/ai2thor_provider.py
kdrumm: Good to know.  Fixed.

deanwetherby:
You may need to run isort manually on those two files.

machine_common_sense/unity_executable_provider.py
```diff
@@ -0,0 +1,269 @@
...
+
+    def _do_zip_to_cache(self, version: str, zip_file: Path):
+        pass
+
+    def get_execution_location(self, version: str) -> Path:
+        ver_dir = self._get_version_dir(version)
+        return ver_dir.joinpath(self.APP_FILE.format(version))
+
+    # MAC appears to have MCS-AI2-THOR-Unity-App-v0.4.2.app folder, but need
+    # to verify later
```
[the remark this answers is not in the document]
rartiss55: The executable path is actually within in the .app directory, so it would be: 
`EXECUTABLE_FILE = "MCS-AI2-THOR-Unity-App-v{}.app/Contents/MacOS/MCS-AI2-THOR"`

Additionally, it looks like another `filepath.chmod(755)` call is needed after the download is unzipped. After that, I think this will work well for Mac users. Nice job!

tests/test_unity_executable_provider.py
```diff
@@ -59,8 +62,19 @@ def test_provider_init(self):
...
     def test_get_executable_empty(self):
-        result = self.provider.get_executable("0.0.0")
+        result = self.provider.get_executable("test")
+        self.assertTrue(result.exists())
+
+    def test_get_executable_cache(self):
+        self.assertEquals(self.provider._downloader.count, 0)
+        result = self.provider.get_executable("test2")
+        self.assertTrue(result.exists())
+        self.assertEquals(self.provider._downloader.count, 1)
```
deanwetherby, line 72: assertEquals is deprecated. Use assertEqual.

tests/test_unity_executable_provider.py
```diff
@@ -59,8 +62,19 @@ def test_provider_init(self):
...
+
+        # try again and verify the downloader isn't called again.
+        result = self.provider.get_executable("test2")
         self.assertTrue(result.exists())
+        self.assertEquals(self.provider._downloader.count, 1)
```
deanwetherby, line 77: replace with assertEqual

machine_common_sense/unity_executable_provider.py
```diff
@@ -75,15 +82,13 @@ def get_executable(self, version=current_version,
                 "unable to locate ai2thor for version={}".format(version))
 
 
-CACHE_LOCATION = "~/.mcs/"
-
-
 class AbstractExecutionCache(ABC):
+    CACHE_LOCATION = "~/.mcs/"
```
deanwetherby, line 86: Something to consider here is using pathlib.Path.home() rather than ~

tests/test_unity_executable_provider.py
kdrumm: fixed all assertEquals

machine_common_sense/unity_executable_provider.py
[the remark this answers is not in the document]
kdrumm: Thanks for testing.  Looking at it again, it seemed that i could reuse more code between the two platforms, so I rearranged quite a bit between the base and sub classes for the cache, but I believe I've implemented what Mac needs.

deanwetherby (approved):

deanwetherby:
We should make it clear in the docs that you can provide your own unity build file or continue doing things the way we've been doing it. This change is an easy-button option if researcher don't want to manage the downloads.

PR #345, opened 2021-05-24
MCS-685 removed error when ending scene without starting a scene

kdrumm:
Fixed error in history writer where if the writer isn't initiated (because the scene was never started) it would throw an error and crash.

The further question is should we alert the user if they do this?  Trying to end a scene when one has never started is a user error, should we detect it and alert them?  currently, there is no output and nothing happens when end_scene() is called.  stop_simulation() closes the unity window.

### DIFF
--- a/machine_common_sense/controller.py
+++ b/machine_common_sense/controller.py
@@ -245,7 +245,7 @@ class Controller():
 
     def end_scene(self, choice, confidence=1.0):
         """
-        Ends the current scene.  Calling end_scene() before calling 
+        Ends the current scene.  Calling end_scene() before calling
         start_scene() will do nothing.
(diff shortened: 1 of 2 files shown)


ThomasSchellenbergNextCentury (approved):
I don't have a preference on whether or not to throw an error here. Whatever we choose, we should update our documentation: either "Calling `end_scene` before calling `start_scene` does nothing" or "Calling `end_scene` before calling `start_scene` will throw an error"

rartiss55 (approved):

deanwetherby (approved):
Ending without an error seems fine to me.

brianpippin (approved):

PR #347, opened 2021-05-26
Mcs 686 start scene classes

kdrumm:
Converted SceneConfiguration into a concrete class

### DIFF
--- a/docs/source/schema.rst
+++ b/docs/source/schema.rst
@@ -95,7 +95,6 @@ Each **scene config** has the following properties:
 - `wallMaterial` (string, optional): The material (color/texture) for the room's four outer walls. See the :ref:`material list <Material List>` for options. Default (v0.0.3+): `"AI2-THOR/Materials/Walls/DrywallBeige"`
 - `roomMaterials` (:ref:`room material config <Room Material Config>`, optional): The materials for each individual wall.  For any individual wall not provided, or all outer walls if object is not provided, they will use 'wallMaterial' property.
 - `wallProperties` (:ref:`physics config <Physics Config>`, optional): Enable custom friction, bounciness, and/or drag on the walls. Default: see :ref:`physics config <Physics Config>`.
-- `observation` (bool, option): TBD
 
 Object Config
 *************
--- a/docs/source/schema.rst
+++ b/docs/source/schema.rst
@@ -95,6 +95,7 @@ Each **scene config** has the following properties:
 - `wallMaterial` (string, optional): The material (color/texture) for the room's four outer walls. See the :ref:`material list <Material List>` for options. Default (v0.0.3+): `"AI2-THOR/Materials/Walls/DrywallBeige"`
 - `roomMaterials` (:ref:`room material config <Room Material Config>`, optional): The materials for each individual wall.  For any individual wall not provided, or all outer walls if object is not provided, they will use 'wallMaterial' property.
 - `wallProperties` (:ref:`physics config <Physics Config>`, optional): Enable custom friction, bounciness, and/or drag on the walls. Default: see :ref:`physics config <Physics Config>`.
+- `observation` (bool, option): TBD
 
 Object Config
 *************
@@ -105,6 +106,7 @@ Each **object config** has the following properties:
 - `type` (string, required): The object's type from the :ref:`object list <Object List>`.
 - `changeMaterials` (:ref:`change_materials config <Change Materials Config>` array, optional): The steps on which to change the material(s) (colors/textures) used on the object, and the new materials to use. See the :ref:`material list <Material List>` for options. Default: `[]`
 - `forces` (:ref:`force config <Force Config>` array, optional): The steps on which to apply `force <https://docs.unity3d.com/ScriptReference/Rigidbody.AddForce.html>`_ to the object. The config `vector` describes the amount of force (in Newtons) to apply in each direction using the global coordinate system. Resets all existing forces on the object to 0 before applying the new force. Default: `[]`
+- `ghosts` (:ref:`step begin and end config config <Step Begin And End Config>` array, optional): TBD
 - `hides` (:ref:`single step config <Single Step Config>` array, optional): The steps on which to hide the object, completely removing its existence from the scene until it is shown again (see the `shows` property). Useful if you want to have impossible events (spontaneous disappearance). Default: `[]`
 - `kinematic` (boolean, optional): If true, the object will ignore all forces including gravity. See Unity's `isKinematic property <https://docs.unity3d.com/ScriptReference/Rigidbody-isKinematic.html>`_. Usually paired with `structure`. Default: `false`
 - `locationParent` (string, optional): The `id` of another object in the scene. If given, this object's `shows.position` and `shows.rotation` will both start from the position and rotation of the `locationParent` object rather than from `0`. Default: none
@@ -116,9 +118,11 @@ Each **object config** has the following properties:
 - `nullParent` (:ref:`transform config <Transform Config>`, optional): Whether to wrap the object in a null parent object. Useful if you want to rotate an object by a point other than its center point. Default: none
 - `openable` (boolean, optional): Whether the object should be openable, if it is not already openable based on its `type`. Default: depends on `type`
 - `opened` (boolean, optional): Whether the object should begin opened. Must also be `openable`. Default: `false`
+- `openClose` (:ref:`open close config <Open Close Config>`, optional): The steps where an object is opened or closed by the system.  Default: None
 - `physics` (boolean, optional): Whether to enable physics simulation on the object. Automatically `true` if `moveable`, `openable`, `pickupable`, or `receptacle` is `true`. Use `physics` if you want to enable physics but don't want to use any of those other properties. Default: `false`
 - `physicsProperties` (:ref:`physics config <Physics Config>`, optional): Enable custom friction, bounciness, and/or drag on the object. Default: see :ref:`physics config <Physics Config>`.
 - `pickupable` (boolean, optional): Whether the object should be pickupable, if it is not already openable based on its `type`. Pickupable objects are also automatically `moveable`. Default: depends on `type`
+- `receptacle` (boolean, optional): TBD
 - `resizes` (:ref:`size config <Size Config>` array, optional): The steps on which to resize the object. The config `size` is multiplied by the object's current size. Useful if you want to have impossible events (spontaneous resizing). Default: `[]`
 - `rotates` (:ref:`move config <Move Config>` array, optional): The steps on which to rotate the object. The config `vector` describes the amount of rotation (in degrees) to change, added to the object's current rotation. Useful if you want to rotate objects that are `kinematic`. A fifth of each move is made over each of the five substeps (five screenshots) during the step. Default: `[]`
 - `salientMaterials` (string array, optional)
@@ -237,6 +241,13 @@ Each **step begin and end config** has the following properties:
 
 - `stepBegin` (integer, required): The step on which the action should occur.  Must be non-negative.  A value of `0` means the action will occur during scene initialization.
 - `stepEnd` (integer, required): The step on which the action should end.  Must be equal to or greater than the `stepBegin`.
+
+Open Close Config
+*****************
+
+Each **Open Close Config** has the following properties:
+- `step` (integer, required): The step on which the action should occur.  Must be non-negative.  A value of `0` means the action will occur during scene initialization.
+- `open` (boolean, required): If true, the container will be opened, if false, the container will be closed
 
 Teleport Config
 ***************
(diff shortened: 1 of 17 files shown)


machine_common_sense/config_manager.py
```diff
@@ -255,3 +667,48 @@ def retrieve_object_states(
...
+            if (
+                target_name in goal_output.metadata and
+                'image' in goal_output.metadata[target_name] and
+                isinstance(goal_output.metadata[target_name]['image'], str)
+            ):
+                image_list_string = goal_output.metadata[target_name]['image']
+                goal_output.metadata[target_name]['image'] = np.array(
+                    ast.literal_eval(image_list_string)).tolist()
+
+        return goal_output
```
kdrumm, line 714: This function was mostly just moved from another file.

machine_common_sense/controller.py
```diff
@@ -298,27 +301,40 @@ def start_scene(self, config_data):
...
+            config_data['numpyArray'] = [4]
+
+            schema = SceneConfigurationSchema()
+            scene_config = schema.load(config_data)
+
+        self._scene_config = scene_config
         self.__habituation_trial = 1
         self.__step_number = 0
-        self._goal = self.retrieve_goal(self.__scene_configuration)
+        self._goal = self._scene_config.retrieve_goal()
```
deanwetherby, line 315: I much prefer a class telling me about itself rather than operating on a dict.

machine_common_sense/controller.py
```diff
@@ -298,27 +301,40 @@ def start_scene(self, config_data):
...
-        skip_preview_phase = (True if 'goal' in config_data and
-                              'skip_preview_phase' in config_data['goal']
-                              else False)
+        skip_preview_phase = (scene_config.goal is not None and
+                              scene_config.goal.skip_preview_phase)
 
         if (self.isFileWritingEnabled()):
-            os.makedirs('./' + config_data['name'], exist_ok=True)
-            self.__output_folder = './' + config_data['name'] + '/'
+            os.makedirs('./' + scene_config.name, exist_ok=True)
```
deanwetherby, line 321: Class attribute/property is better than dict[key] which is subject to KeyError and other problems.

requirements.txt
```diff
@@ -7,6 +7,7 @@ certifi==2020.6.20
 chardet==3.0.4
 click==7.1.2
 commonmark==0.9.1
+dataclasses==0.8; python_version<"3.7"
```
deanwetherby, line 10: For those that don't know, dataclasses became part of core python in 3.7. So for 3.6, we have to explicitly install the package. This is done conditionally with the semicolon.

deanwetherby (approved):

machine_common_sense/controller.py
```diff
@@ -683,71 +699,35 @@ def mcs_action_to_ai2thor_action(self, action):
...
+    # is there a better way to do this?
+    def _remove_none(self, d):
+        '''Remove all none's from dictionaries'''
+        for key, value in dict(d).items():
+            if isinstance(value, dict):
+                d[key] = self._remove_none(value)
+            if isinstance(value, list):
+                for index, val in enumerate(value):
+                    if isinstance(val, dict):
+                        value[index] = self._remove_none(val)
```
ThomasSchellenbergNextCentury, line 726: This doesn't actually remove None if it's in a list. I'm not sure if that's an important case.

machine_common_sense/controller.py
```diff
@@ -683,71 +699,35 @@ def mcs_action_to_ai2thor_action(self, action):
...
-            last_step=goal_config.get('last_step', None),
-            metadata=goal_config.get('metadata', {})
-        ))
-
     def stop_simulation(self):
         """Stop the 3D simulation environment. This controller won't work any
         more."""
         self._controller.stop()
 
+    # is there a better way to do this?
```
ThomasSchellenbergNextCentury, line 717: IMO we should either make a ticket and add a TODO with the ticket tag to the comment here, or remove the comment.

kdrumm: I'll remove it now.  It was more to make sure this was at least looked at during the PR.

machine_common_sense/config_manager.py
```diff
@@ -230,21 +235,428 @@ def get_screen_height(self) -> int:
...
-    def __init__(self, scene_config_dict):
-        pass
+class Vector3dSchema(Schema):
+    x = fields.Float()
+    y = fields.Float()
+    z = fields.Float()
+
+    @post_load
+    def make_position_3d(self, data, **kwargs):
+        return Vector3d(**data)
```
ThomasSchellenbergNextCentury, line 245: So we'll now need to make a new Python class whenever we add a new property to the JSON schema? In other words, we need to have a separate Python class for each corresponding Serializable class in our Unity code? That seems unfortunate, but I guess we'll make it work.

machine_common_sense/config_manager.py
```diff
@@ -255,3 +667,48 @@ def retrieve_object_states(
                     return [state_list]
                 return [str(state) for state in state_list]
         return []
+
+    def retrieve_goal(self):
```
ThomasSchellenbergNextCentury, line 671: If these are now public functions, we should add proper function documentation to them.

machine_common_sense/controller.py
```diff
@@ -683,71 +699,35 @@ def mcs_action_to_ai2thor_action(self, action):
...
+    def _remove_none(self, d):
```
ThomasSchellenbergNextCentury, line 717: Should we add some unit tests for this?

machine_common_sense/controller.py
```diff
@@ -298,27 +301,40 @@ def start_scene(self, config_data):
...
-        self.__scene_configuration = config_data
+        if isinstance(config_data, SceneConfiguration):
+            scene_config = config_data
+        else:
+            config_data['numpyArray'] = [4]
+
+            schema = SceneConfigurationSchema()
+            scene_config = schema.load(config_data)
+
+        self._scene_config = scene_config
```
ThomasSchellenbergNextCentury, line 312: Do we have a unit test in test_controller that verifies this conversion process?

ThomasSchellenbergNextCentury:
@kdrumm I know I already gave you some feedback on the numpy arrays, but we might want to reach out to the CORA team to make sure this change still works for them, since I'm not completely sure about their use case here. Maybe @deanwetherby could help? (We don't need to hold off on merging until we hear back from them, since it might take a while, but we should at least let them know that there's a change.)

machine_common_sense/config_manager.py
kdrumm: I can certainly add documentation, although I'm not exactly sure why the functions were originally needed.  They are no more or less public than they have been as well.

machine_common_sense/config_manager.py
kdrumm: Yes that is true and it is somewhat unfortunate.  I think this is a fair tradeoff to have a well defined and validated API (at least at this endpoint).  There are quite a few fields that were missing from our documentation (which I probably should note and update).

machine_common_sense/controller.py
kdrumm: Yes, good idea.

machine_common_sense/config_manager.py
deanwetherby: Why is it 'unfortunate'?

machine_common_sense/controller.py
kdrumm: No, but I can add one.

machine_common_sense/config_manager.py
kdrumm: we are essentially duplicating code between python and C# with the data objects.  Although I'm fairly sure that's standard practice when you have to send data between two languages.

kdrumm:
> @kdrumm I know I already gave you some feedback on the numpy arrays, but we might want to reach out to the CORA team to make sure this change still works for them, since I'm not completely sure about their use case here. Maybe @deanwetherby could help? (We don't need to hold off on merging until we hear back from them, since it might take a while, but we should at least let them know that there's a change.)

Honestly, I still don't understand how they used numpy arrays, where they added them and what the integration test was testing.  I personally think that test should fail because its adding a random field anyway.  I'm sure I'm missing something because I don't have much history here.

machine_common_sense/config_manager.py
deanwetherby: So if we make a change in Unity and the corresponding python side fails, that's a good thing.

machine_common_sense/config_manager.py
```diff
@@ -230,21 +235,428 @@ def get_screen_height(self) -> int:
...
+    isometric = fields.Bool()
+    answer = fields.Dict()
+    floorProperties = fields.Nested(PhysicsConfigSchema)
+    wallProperties = fields.Nested(PhysicsConfigSchema)
+    screenshot = fields.Bool()
+    observation = fields.Bool()
+    isometric = fields.Bool()
+    # this just allows the integration test until we figure out what to do
+    # with the test.
+    numpyArray = fields.Raw()
```
ThomasSchellenbergNextCentury, line 468: Might want to make a ticket for this. We'll need some feedback from the CORA team.

machine_common_sense/config_manager.py
ThomasSchellenbergNextCentury: That's fine, and you make a fair point that it adds clarity and validation. Just wanted to make sure that I understood all the ramifications of this change.

ThomasSchellenbergNextCentury:
> Honestly, I still don't understand how they used numpy arrays, where they added them and what the integration test was testing. I personally think that test should fail because its adding a random field anyway. I'm sure I'm missing something because I don't have much history here.

The use case was: CORA is using numpy arrays somewhere in their scene data, I don't know where, so I put it in a random place. I think the solution is to find out from CORA where they're using numpy arrays and update the tests accordingly.

kdrumm:
> > Honestly, I still don't understand how they used numpy arrays, where they added them and what the integration test was testing. I personally think that test should fail because its adding a random field anyway. I'm sure I'm missing something because I don't have much history here.
> 
> The use case was: CORA is using numpy arrays somewhere in their scene data, I don't know where, so I put it in a random place. I think the solution is to find out from CORA where they're using numpy arrays and update the tests accordingly.

Created a ticket to handle this later: https://nextcentury.atlassian.net/browse/MCS-689

integration_tests/additional_integration_tests.py
```diff
@@ -230,9 +230,7 @@ def run_position_by_step_test(controller, metadata_tier):
 
 
 def run_public_sample_scenes_test(controller, metadata_tier):
-    scene_filename_list = glob.glob(SAMPLE_SCENES_FOLDER + '*.json')
-    scene_filename_list.sort()
-
+    scene_filename_list = sorted(glob.glob(SAMPLE_SCENES_FOLDER + '*.json'))
```
kdrumm, line 233: My IDE forced this change for some reason.  Any objections?  I assume the IDE is just trying to comply with best practices.

deanwetherby: What IDE? Do you have any interesting plugins?

https://sourcery.ai/

kdrumm: i have a couple (pylance, python, python extension pack, python for vscode, python test explorer and now sourcery).

kdrumm:
I've added some more unit tests, added some documentation, updated the numpy integration test.

integration_tests/additional_integration_tests.py
ThomasSchellenbergNextCentury: Yeah this looks good

docs/source/schema.rst
```diff
@@ -95,6 +95,7 @@ Each **scene config** has the following properties:
 - `wallMaterial` (string, optional): The material (color/texture) for the room's four outer walls. See the :ref:`material list <Material List>` for options. Default (v0.0.3+): `"AI2-THOR/Materials/Walls/DrywallBeige"`
 - `roomMaterials` (:ref:`room material config <Room Material Config>`, optional): The materials for each individual wall.  For any individual wall not provided, or all outer walls if object is not provided, they will use 'wallMaterial' property.
 - `wallProperties` (:ref:`physics config <Physics Config>`, optional): Enable custom friction, bounciness, and/or drag on the walls. Default: see :ref:`physics config <Physics Config>`.
+- `observation` (bool, option): TBD
```
ThomasSchellenbergNextCentury, line 98: The `observation` property is deprecated, so maybe we don't need it in the docs?

machine_common_sense/config_manager.py
```diff
@@ -460,7 +460,6 @@ class SceneConfigurationSchema(Schema):
     wallProperties = fields.Nested(PhysicsConfigSchema)
     screenshot = fields.Bool()
     observation = fields.Bool()
-    isometric = fields.Bool()
```
ThomasSchellenbergNextCentury: We're definitely using `isometric` for all the passive agent scenes, so why is it being deleted?

ThomasSchellenbergNextCentury: Ooops, I see it was repeated. Never mind.

ThomasSchellenbergNextCentury (approved):
I started running the integration tests earlier and then forgot about them... But they all pass for me. Looks good.

PR #350, opened 2021-06-01
Mcs 663 payload classes

kdrumm:
All controller events have their own specific data payload class.

I'm not sure this is a huge improvement, but at least subscribers should know what data is available.

### DIFF
--- a/machine_common_sense/controller.py
+++ b/machine_common_sense/controller.py
@@ -191,7 +191,6 @@ class Controller():
             # change later
             subscriber.on_event(event_type, payload)
 
-    # TODO this is a huge data dumpster and needs to be thought about
     def _create_event_payload_kwargs(self):
         return {"step_number": self.__step_number,
                 "config": self._config,
--- a/machine_common_sense/controller.py
+++ b/machine_common_sense/controller.py
@@ -191,14 +191,14 @@ class Controller():
             # change later
             subscriber.on_event(event_type, payload)
 
-    def _create_event_payload_kwargs(self):
+    def _create_event_payload_kwargs(self) -> dict:
         return {"step_number": self.__step_number,
                 "config": self._config,
                 "scene_config": self._scene_config}
 
     def _create_post_step_event_payload_kwargs(
             self, wrapped_step, step_metadata, step_output: StepMetadata,
-            restricted_step_output: StepMetadata):
+            restricted_step_output: StepMetadata) -> dict:
         args = self._create_event_payload_kwargs()
         args['output_folder'] = self.__output_folder
         args['timestamp'] = datetime.datetime.now().strftime("%Y%m%d-%H%M%S")
--- a/machine_common_sense/controller.py
+++ b/machine_common_sense/controller.py
@@ -1,11 +1,11 @@
-
 import atexit
+import datetime
 import glob
 import json
 import logging
 import os
 import random
-from typing import Dict, List
+from typing import Dict, List, Union
 
 import ai2thor.controller
 import ai2thor.server
@@ -27,8 +27,9 @@ DEFAULT_MOVE = 0.1
 from .action import Action
 from .config_manager import (ConfigManager, SceneConfiguration,
                              SceneConfigurationSchema)
-from .controller_events import (ControllerEventPayload, EventType,
-                                PredictionPayload)
+from .controller_events import (AfterStepPayload, BeforeStepPayload,
+                                EndScenePayload, EventType, PredictionPayload,
+                                StartScenePayload)
 from .controller_output_handler import ControllerOutputHandler
 from .goal_metadata import GoalMetadata
 from .step_metadata import StepMetadata
@@ -178,7 +179,9 @@ class Controller():
             self._subscribers.append(subscriber)
 
     def _publish_event(self, event_type: EventType,
-                       payload: ControllerEventPayload):
+                       payload: Union[StartScenePayload, BeforeStepPayload,
+                                      AfterStepPayload, PredictionPayload,
+                                      EndScenePayload]):
         for subscriber in self._subscribers:
             # TODO should we make a deep copy of the payload so subscribers
             # cannot change source data?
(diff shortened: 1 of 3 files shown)


machine_common_sense/controller.py
```diff
@@ -273,12 +283,15 @@ def end_scene(self, choice, confidence=1.0):
...
-        self._publish_event(EventType.ON_END_SCENE, payload)
+        payloadArgs = self._create_event_payload_kwargs()
+        payloadArgs['choice'] = choice
+        payloadArgs['confidence'] = confidence
+        payloadArgs['uploader'] = uploader
+        payloadArgs['uploader_folder_prefix'] = self._config.get_s3_folder()
+        self._publish_event(
+            EventType.ON_END_SCENE,
+            EndScenePayload(
+                **payloadArgs))
```
ThomasSchellenbergNextCentury, line 294: Would it be better to move this behavior into the Payload class (e.g. EndScenePayload in this case), as either a static function that returns a new Payload from the arguments, or in the Payload's constructor? I know that would involve passing more arguments (e.g. here it would be: self.__step_number, self._config, self._scene_config, choice, confidence, uploader), but maybe it would be cleaner (i.e. the _create_* functions would be defined in the corresponding Payload classes and not in the Controller)? Same with the other Payload classes.

machine_common_sense/controller_events.py
```diff
@@ -1,12 +1,17 @@
 import datetime
 import enum
 from abc import ABC
+from dataclasses import dataclass
 from typing import Dict, List
 
 import PIL
+from ai2thor.server import Event
+from marshmallow import Schema, fields  # , post_load
```
ThomasSchellenbergNextCentury, line 9: Unnecessary comment

machine_common_sense/controller_events.py
```diff
@@ -20,10 +25,70 @@ class EventType(enum.Enum):
...
 
 
-class BaseEventPayload:
+class BaseEventPayloadSchema(Schema):
+    step_number = fields.Int()
 
-    def __init__(self, step_number: int):
-        self.step_number = step_number
+
+@dataclass
```
ThomasSchellenbergNextCentury, line 32: I was totally looking for something like this to use in the scene generator. Thanks for the tip!

machine_common_sense/controller_events.py
```diff
@@ -20,10 +25,70 @@ class EventType(enum.Enum):
...
+    habitation_trial = fields.Int()
+    restricted_step_output
+    wrapped_step = wrapped_step
+    step_metadata = step_output
+    step_output = pre_restrict_output
+    restricted_step_output = output
+
+    @post_load
+    def make_scene_configuration(self, data, **kwargs):
+        return SceneConfiguration(**data)'''
```
ThomasSchellenbergNextCentury, line 91: Should this be deleted? Or will it be needed in the future?

machine_common_sense/controller_events.py
kdrumm: @deanwetherby pointed it out to me.

machine_common_sense/controller.py
kdrumm: Maybe?  I think the best options are this or just call the constructor with the arguments directly.  The only reasons I did it this way was to reuse the create kwargs functions and I don't like passing massive amounts of parameters on one line.  (Note: end scene is one of the simpler examples).

ThomasSchellenbergNextCentury: That's fine. I don't have an issue with doing it this way. I figured I'd start a conversation on the topic, just in case. :)

machine_common_sense/controller_events.py
deanwetherby: @ThomasSchellenbergNextCentury dataclasses were introduced as of python 3.7 and then back-ported to 3.6. Since the scene generator requires a later python version, you would not have to include it as a dependency since it's already built in to core python.

machine_common_sense/controller.py
```diff
@@ -188,16 +191,23 @@ def _publish_event(self, event_type: EventType,
...
-    def _create_event_payload(self):
-        payload = ControllerEventPayload(
-            self.__output_folder,
-            self._config,
-            self.__step_number,
-            self._scene_config,
-            self.__habituation_trial,
-            self._goal)
-        return payload
+    def _create_event_payload_kwargs(self):
```
deanwetherby, line 194: Consider adding type hint for the returned object.

machine_common_sense/controller.py
```diff
@@ -188,16 +191,23 @@ def _publish_event(self, event_type: EventType,
...
+        return {"step_number": self.__step_number,
+                "config": self._config,
+                "scene_config": self._scene_config}
+
+    def _create_post_step_event_payload_kwargs(
+            self, wrapped_step, step_metadata, step_output: StepMetadata,
+            restricted_step_output: StepMetadata):
```
deanwetherby, line 201: Return type hint

machine_common_sense/controller.py
kdrumm: Its a good thing to bring up.  I'm not sure I like it.  I just don't like anything else better (yet).

machine_common_sense/controller_events.py
kdrumm: This is the stuff I get so used to (saving it for later) that I don't see it in my own review.  Thanks for pointing it out.

machine_common_sense/controller.py
deanwetherby: Good use of dictionary unpacking. I stare at this a bit too but don't really see a better way.

machine_common_sense/controller.py
```diff
@@ -191,14 +191,14 @@ def _publish_event(self, event_type: EventType,
             # change later
             subscriber.on_event(event_type, payload)
 
-    def _create_event_payload_kwargs(self):
+    def _create_event_payload_kwargs(self) -> dict:
```
deanwetherby, line 194: Interesting. Python 3.9+ allows the lower-case version as the type hint (as in 'dict'). Below version 3.9, you usually import the upper-case Dict from the typing package. I always done the latter although I just tested both in 3.6 and nothing went wrong.

kdrumm:
I removed the commented code that I missed and Thomas found.  Also added type hints where Dean pointed it out.

deanwetherby (approved):

PR #352, opened 2021-06-02
Mcs 409 scene event class

kdrumm:
See MCS-409 ticket for motivation here.  

Also, this ticket is somewhat hard to review as the diff doesn't put changes near each other very well.

### DIFF
--- a/machine_common_sense/controller_output_handler.py
+++ b/machine_common_sense/controller_output_handler.py
@@ -1,10 +1,12 @@
 import copy
 import logging
+from dataclasses import dataclass
 
 import numpy as np
 import PIL
+from ai2thor.server import Event
 
-from .config_manager import ConfigManager
+from .config_manager import ConfigManager, SceneConfiguration
 from .controller import DEFAULT_MOVE
 from .material import Material
 from .object_metadata import ObjectMetadata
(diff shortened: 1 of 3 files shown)


machine_common_sense/controller_output_handler.py
```diff
@@ -16,202 +18,149 @@
...
+    def rotation(self) -> float:
+        return self._raw_output.metadata['agent']['rotation']['y']
 
-    def process_image_data(self):
-        self._image_list, self._depth_map_list, self._object_mask_list = \
-            self.save_image_data(
-                self._raw_output.metadata.get(
-                    'clippingPlaneFar',
-                    ConfigManager.DEFAULT_CLIPPING_PLANE_FAR
-                )
```
kdrumm, line 105: Note clipping plane far default

machine_common_sense/controller_output_handler.py
```diff
@@ -16,202 +18,149 @@
...
-        step_output = StepMetadata(
-            action_list=goal.retrieve_action_list_at_step(
-                self._step_number
-            ),
-            camera_aspect_ratio=(
-                self._config.get_screen_width(),
-                self._config.get_screen_height()),
-            camera_clipping_planes=(
-                self._raw_output.metadata.get('clippingPlaneNear', 0.0),
-                self._raw_output.metadata.get('clippingPlaneFar', 0.0),
```
kdrumm, line 147: Note clippingPlaneFar default

machine_common_sense/controller_output_handler.py
```diff
@@ -16,202 +18,149 @@
...
+    @property
+    def clipping_plane_far(self):
+        # Review: The clipping plane is retrieved in two different places.
+        # The Step Output constructor had the default as 0, the image
+        # processing had the default as the constant
+        # from the ConfigManager.  This seems wrong so I put it as the
+        # constant.  We should discuss before final approval.
+        # self._raw_output.metadata.get('clippingPlaneFar', 0.0)
+        return self._raw_output.metadata.get(
+            'clippingPlaneFar', ConfigManager.DEFAULT_CLIPPING_PLANE_FAR)
```
ThomasSchellenbergNextCentury, line 93: I think this is correct (we probably never had to fall back to the default, which is why we didn't notice).

machine_common_sense/controller_output_handler.py
```diff
@@ -16,202 +18,149 @@
...
+        # if no config specified, return visible objects (for now)
 
-        restrict_non_oracle = (
-            restricted and
-            (metadata_tier == ConfigManager.CONFIG_METADATA_TIER_NONE or
-             metadata_tier == ConfigManager.CONFIG_METADATA_TIER_LEVEL_1 or
-             metadata_tier == ConfigManager.CONFIG_METADATA_TIER_LEVEL_2)
-        )
+        # Reviewer: Take a hard look at this change to make sure my logic is
+        # valid then we should remove this comment
```
ThomasSchellenbergNextCentury, line 129: I don't see anything obviously incorrect around here. Which change here specifically concerned you? The return value?

machine_common_sense/controller_output_handler.py
```diff
@@ -16,202 +18,149 @@
...
-
-    def retrieve_head_tilt(self):
-        return self._raw_output.metadata['agent']['cameraHorizon']
-
-    def retrieve_rotation(self):
-        return self._raw_output.metadata['agent']['rotation']['y']
+    @property
+    def return_status(self):
+        # TODO MCS-47 Need to implement all proper step statuses on the Unity
+        # side
```
ThomasSchellenbergNextCentury, line 147: I don't think this TODO is needed anymore.

machine_common_sense/controller_output_handler.py
```diff
@@ -375,30 +244,146 @@ def retrieve_object_output(
...
+                    object_metadata, self.object_colors
                 )
-                depth_map_list.append(np.array(depth_float_array))
+                for object_metadata in self._raw_output.metadata[
+                    'structuralObjects'
+                ]
+                if object_metadata['visibleInCamera'] or show_all
+            ],
+            key=lambda x: x.uuid
+        )
```
ThomasSchellenbergNextCentury, line 267: Would it be smart to combine this similar code with the returned data for the object_list function?

machine_common_sense/controller_output_handler.py
```diff
@@ -375,30 +244,146 @@ def retrieve_object_output(
...
+            restricted and
+            (metadata_tier == ConfigManager.CONFIG_METADATA_TIER_NONE or
+             metadata_tier == ConfigManager.CONFIG_METADATA_TIER_LEVEL_1 or
+             metadata_tier == ConfigManager.CONFIG_METADATA_TIER_LEVEL_2)
+        )
+
+        depth_map_list = [] if restrict_depth_map else (
+            self._scene_event.depth_map_list)
+        # image_list = [] if restrict_non_oracle else
+        # self._scene_event.image_list
```
ThomasSchellenbergNextCentury, line 318: Are these lines needed?

machine_common_sense/controller_output_handler.py
```diff
@@ -375,30 +244,146 @@ def retrieve_object_output(
...
+            goal, habituation_trial, True)
+        return (unrestricted, restricted)
 
-            if self._config.is_object_masks_enabled():
-                object_mask = PIL.Image.fromarray(
-                    event.instance_segmentation_frame)
-                object_mask_list.append(object_mask)
+    # used to be wrap output
+    def _get_step_metadata(self, goal, habituation_trial,
+                           restricted=True) -> StepMetadata:
```
ThomasSchellenbergNextCentury, line 295: Looking much nicer with this change!

machine_common_sense/controller_output_handler.py
kdrumm: the original retrieve function is at 233 to 262 in the old code.

machine_common_sense/controller_output_handler.py
kdrumm: Good idea.  I combined them.  The only weirdness is that now the structural objects will be tested for 'isPickedUp'.  I assume that will always be false.

kdrumm:
Removed 2 reviewer comments, commented code, and todo comment.  Will push once I verify tests.

machine_common_sense/controller_output_handler.py
```diff
@@ -1,10 +1,12 @@
 import copy
 import logging
+from dataclasses import dataclass
 
 import numpy as np
 import PIL
+from ai2thor.server import Event
```
deanwetherby, line 7: I wish ai2thor had a better way to import this Event so that it was namespaced.

deanwetherby (approved):
Good use of the property decorator in the controller_output_handler. Converting to classes is starting to pay dividends as the code appears to be more organized.

ThomasSchellenbergNextCentury (approved):

PR #354, opened 2021-06-04
MCS-640 moved ai2thor json writing to recorder

kdrumm:
This is a small update to make some of the json writing threaded.

It be nice to thread the debug log file writer, but it creates a json string in a custom way.  We could create yet another record to just write text?


deanwetherby:
Any unit tests?

kdrumm:
Added a bunch of unit tests

deanwetherby (approved):

PR #356, opened 2021-06-07
Mcs 552 change config

kdrumm:
Added MCS fucntion to change the config and changed integration tests to run with a single Unity instance

Also, added a copy for integration test 11 at #74.  The goal here is to verify we are properly resetting floor properties after intuitive physics is set and then unset by other tests.

Note: you will need to use a recent development version of our ai2thor (6/7/21)


kdrumm:
Should I add some kind of checking to ensure the change config isn't called in the middle of a run?

integration_tests/run_handmade_tests.py
```diff
@@ -324,7 +328,7 @@ def start_handmade_tests(
                 successful_test_list.append((test_name, metadata_tier))
             else:
                 failed_test_list.append((test_name, metadata_tier, status))
-        controller.stop_simulation()
+        # controller.stop_simulation()
```
ThomasSchellenbergNextCentury, line 331: I think you can remove this line now

machine_common_sense/controller.py
```diff
@@ -219,23 +219,29 @@ def _convert_y_image_coord_for_unity(self, y_coord):
...
-        # Whether or not to show depth maps or object masks is based on
-        # metadata tier (the default for these if no metadata level is
-        # set or metadata is set to `none`, is `False`)
+    def set_config(self, config):
+        '''Allows config to be changed without changing the controller and
+        attached Unity process.  This typically should only be called by the
+        MCS package itself.
+
+        For users, call machine_common_sense.change_config()
+        '''
```
ThomasSchellenbergNextCentury, line 235: Should we prefix this function name with an underscore to discourage users from calling it directly?

integration_tests/run_handmade_tests.py
kdrumm: I realized we should still probably stop the simulation, but just at the very end.  Minor change income once i finish my tests.

machine_common_sense/controller.py
kdrumm: I can see arguments either way.  I usually treat _*** as private and try not to violate it (except tests).  Of course, things will break if users try to use this.  Maybe practicality is better than purity.  (also, what I can purity may not be pythonic purity).  @deanwetherby

deanwetherby: I would prefix with an underscore.

brianpippin (commented):
LGTM

deanwetherby (approved):

kdrumm:
I'm waiting to merge this due to integration test numbering contention.  @jaudick has 6 or so tests that would have to be shifted and it'll be easier to change this one.  This fix may also cause 0.4.3 ai2thor to no longer work.

We may want to come up with a new way to organize tests as more people are writing integration tests in this branch.

brianpippin (approved):

PR #357, opened 2021-06-09
MCS-700 Added debug scene and object properties to the schema classes.

ThomasSchellenbergNextCentury:
- Added debug and UI properties to the schema classes
- Marked debug and UI properties with comments

Note that this change is NOT backwards compatible, but since the issue only affects debug scene files, hopefully that limitation won't matter very much.

You can generate example scenes on your own, but here's a set that I used for my testing: https://ta2-examples.s3.amazonaws.com/mcs700_scenes.zip

Run example (from the MCS repo's root folder):
```
python scripts/run_passive_scenes.py --mcs_unity_build_file <path_to_unity_app> <path>/mcs700_0001_
```

Corresponding python PR: https://github.com/NextCenturyCorporation/mcs-private/pull/32
Corresponding ingest PR: TODO (probably when I return from vacation)


kdrumm (approved):

ThomasSchellenbergNextCentury:
> We should probably consider making a new ticket for the UI properties and turning them into concrete classes.

I think that task will be part of [this ticket](https://nextcentury.atlassian.net/browse/MCS-696) and [this one](https://nextcentury.atlassian.net/browse/MCS-697), but it's a good point to keep in mind! Thanks!

brianpippin (approved):

PR #358, opened 2021-06-09
Adding script for launching caching addressables build

erick-u:


scripts/cache_addressables.py
```diff
@@ -0,0 +1,8 @@
+import sys
+import os
```
ThomasSchellenbergNextCentury, line 2: iSort is failing - you have to reorder your imports

ThomasSchellenbergNextCentury:
Did we want to update the README with a step to run this script, or will that be done in a separate PR?

deanwetherby (commented):
This might be good to add to the entry_points in the package setup.py.

PR #360, opened 2021-06-11
MCS-682 add wheel file

kdrumm:
This PR will add a master-merge.yaml workflow.  This workflow will kick off only when code is merged into master.  It will perform one last run of tests, but those tests are only on ubuntu, python version 3.6.  This choice is rather arbitrary, but I wanted only 1 instance and all the code should have already been tested when moved to development.  After the tests, a wheel file is created and a new pre-release is created.  The pre release includes the wheel file and a source tar.gz file and is tagged as latest.  

We still need to manually add the unity build, edit the version, and change the tag if necessary.  

There is an alternative method to do this where we trigger on a new tag that is semantically versioned.  I mistakenly told standup that we were not semantically versioned, but I believe we are.

### DIFF
--- a/.github/master-merge.yaml
+++ b/.github/master-merge.yaml
@@ -1,0 +1,47 @@
+name: Pre-release to latest
+
+on:
+  push:
+    branches:
+      master
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        # we are building a pure python wheel so it should work on any platform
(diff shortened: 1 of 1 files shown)


kdrumm:
This is also a test for doing pull requests from outside repositories

deanwetherby (approved):

PR #361, opened 2021-06-14
Picking up static containers when switched to a pickupable object throws nre exception fixed integration testing

jaudick:
https://github.com/NextCenturyCorporation/ai2thor/pull/148

The test is labeled 80 because I made a bunch of tests for ramp ascension that stopped at 79 that have not been merged yet

### DIFF
--- a/integration_tests/data/080.pickup_chest.actions.txt
+++ b/integration_tests/data/080.pickup_chest.actions.txt
@@ -1,0 +1,3 @@
+LookDown
+PickupObject,objectId=treasure_chest
+MoveAhead
--- a/integration_tests/data/080.pickup_chest.level1.outputs.json
+++ b/integration_tests/data/080.pickup_chest.level1.outputs.json
@@ -1,0 +1,52 @@
+﻿[
+    {
+        "head_tilt": 40,
+        "objects_count": 0,
+        "position_x": null,
+        "position_z": null,
+        "return_status": "SUCCESSFUL",
+        "rotation_y": null,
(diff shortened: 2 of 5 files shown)


kdrumm:
I'm trying to replicate the problem using the 0.4.3 build of ai2thor to compare the difference.  However, I am unable to.  Do you know exactly how to replicate?  The integration test should fail on an older build of ai2thor like 0.4.3.

jaudick:
> I'm trying to replicate the problem using the 0.4.3 build of ai2thor to compare the difference. However, I am unable to. Do you know exactly how to replicate? The integration test should fail on an older build of ai2thor like 0.4.3.

I left this comment of the unity pull "You can still pick up the box in the old executable. It only throws the error inside the editor console. It does not interfere with any actions, it just appears as an error. I'm not sure how to test an NRE in the integration tests but I made one that sets up the scenario"

kdrumm:
Ok, that makes more sense.  I didn't realize it was only in the editor.  If that is the case, we should probably close this PR without merge.

rartiss55:
I'm actually leaning towards keeping the test in -- it doesn't hurt anything/nice to have confirmation that there aren't any other issues with chest_1?

brianpippin:
I always feel like an extra test can't hurt so long as it is stable.

brianpippin (approved):

kdrumm (approved):
You guys are right.  It certainly doesn't hurt.

PR #364, opened 2021-06-30
Fixed null parent typing in object schema class.

ThomasSchellenbergNextCentury:


deanwetherby:
Is there a unit test you could add that exhibits this problem?

ThomasSchellenbergNextCentury:
> Is there a unit test you could add that exhibits this problem?

We should probably be testing each schema property in the unit tests. I'll work on that.

PR #369, opened 2021-07-07
Added the optional prefix argument to the developer run scripts that appends a prefix to the output filenames

ThomasSchellenbergNextCentury:
By default, scripts like `run_last_action.py` save output data named using the input scene's `name` property. For example, a scene with `"name": "thomas_1"` would generate an output folder named `thomas_1/` and, with the `--save-videos` flag, an output video named `thomas_1.mp4`

The `--prefix` flag lets you add a prefix to these output files, so, in the above example, `--prefix foobar` would generate an output folder named `foobar_thomas_1` and an output video named `foobar_thomas_1.mp4`

This is useful in the pipeline to append evaluation names and team names to the output videos uploaded to S3.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -39,7 +39,8 @@ Use the following command line options:
 
 - `--debug` Saves output images of each action/frame/step to file.
 - `--save-videos` Creates a video from all the output images after running each scene.
-- `--rename <prefix>` Renames the output files to use the given file prefix.
+- `--prefix <prefix>` Uses the given prefix with each scene's name as the name for all the output files.
+- `--rename <name>` Ignores each scene's original name and uses the given name with each scene's ID (if available) as the name for all the output files.
 
 Example:
 
--- a/scripts/runner_script.py
+++ b/scripts/runner_script.py
@@ -15,12 +15,7 @@ BLACK_IMAGE_PATH = SCRIPT_FOLDER + '/black_image.png'
 
 
 class AbstractRunnerScript():
-    def __init__(
-        self,
-        name,
-        action_callback,
-        rename=False
-    ):
+    def __init__(self, name, action_callback):
         self._name = name
         args, filename_list = self._read_args()
         self.args = args
@@ -48,7 +43,8 @@ class AbstractRunnerScript():
                 filename,
                 action_callback,
                 args.last_step,
-                args.rename if args.rename else ('' if rename else None)
+                args.prefix,
+                args.rename
             )
             if args.save_videos or args.save_gifs:
                 # Copy the black image into the debug folder as the last frame.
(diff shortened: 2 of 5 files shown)


scripts/runner_script.py
```diff
@@ -79,9 +75,10 @@ def _append_subclass_args_to_parser(self, parser):
     def _read_args(self):
```
ThomasSchellenbergNextCentury, line 75: I reordered the `add_argument` calls in this function. The main new thing here is the `prefix` argument.

scripts/README.md
```diff
@@ -39,7 +39,8 @@ Use the following command line options:
 
 - `--debug` Saves output images of each action/frame/step to file.
 - `--save-videos` Creates a video from all the output images after running each scene.
-- `--rename <prefix>` Renames the output files to use the given file prefix.
+- `--prefix <prefix>` Uses the given prefix with each scene's name as the name for all the output files.
+- `--rename <name>` Ignores each scene's original name and uses the given name with each scene's ID (if available) as the name for all the output files.
```
ThomasSchellenbergNextCentury, line 43: I use `rename` when generating examples for the psych team. I plan to use `prefix` for generating videos in my new AMI.

kdrumm (commented):
I ran this twice.  

first with the command: `python scripts/run_just_rotate.py --config_file scripts/kyle-test.ini --prefix test1 docs/source/scenes/playroom.json`

This resulted in videos in the test1_playroom folder with names like `_oracle__test1_playroom_depth_20210713-101603.mp4`

The second time I ran with the command: `python scripts/run_just_rotate.py --level2 --save-videos --prefix test2 docs/source/scenes/playroom.json`

This time the video was placed outside the `test2_playroom` folder and named `test2_playroom.mp4`

This may not be a new problem related to the changes, but I think the names and locations should consistent between config file vs no config file.

ThomasSchellenbergNextCentury:
@kdrumm What is in `scripts/kyle-test.ini`? Does it use `videos_enabled: true`?

I think `_oracle__test1_playroom_depth_20210713-101603.mp4` is being generated by the MCS python library (outside of the script) while `test2_playroom.mp4` is being generated by the `run_just_rotate` script, which would explain the naming inconsistency. The former are needed by the eval pipeline while the latter are needed by people running the script.

kdrumm:
Here is my config:
```
[MCS]
history_enabled: true
metadata: oracle
save_debug_images: true
save_debug_json: true
video_enabled: true
```

So to clarify, does --save-videos not set the same value as the config file flag?  Does --level1 change the metadata used the same as the config?

Again, this mostly sounds like it is not related to the changes, but this is just when i discovered this.

ThomasSchellenbergNextCentury:
@kdrumm Yeah no worries. The `--save-videos` flag is completely different from `video_enabled: true`. The latter does a lot of extra things that I don't normally need (creating top-down videos, heatmap videos, etc.), and I wanted to customize the video making for my purposes (distributing videos, as opposed to posting them in the UI). The `--level1`, `--level2`, and `--oracle` flags specify which `scripts/config_*.ini` file to use.

kdrumm (approved):

brianpippin (approved):
LGTM

PR #371, opened 2021-07-13
Automatically Create Log Folder inside /scripts if Needed

jaudick:
I think the solution is supposed to be this straightforward, link to the issue:
https://nextcentury.atlassian.net/browse/MCS-753?atlOrigin=eyJpIjoiZWQwMzAzMDFlYjNiNGVmNTliZWFiOTcwYzM0ZjA4NmIiLCJwIjoiaiJ9

### DIFF
--- a/machine_common_sense/logging_config.py
+++ b/machine_common_sense/logging_config.py
@@ -39,8 +39,7 @@
 
 import ast
 import logging
-from os import mkdir
-from os.path import exists
+import os
 
 logger = logging.getLogger(__name__)
 
@@ -86,7 +85,7 @@ class LoggingConfig():
         should be read.
         """
         init_message = ""
-        if (exists(log_config_file)):
+        if (os.path.exists(log_config_file)):
             with open(log_config_file, "r") as data:
                 log_config = ast.literal_eval(data.read())
                 init_message = "Loaded logging config from " + log_config_file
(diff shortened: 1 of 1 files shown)


kdrumm:
Just a little background/soapboxing.  I consider this a glaring bug of the logging framework.  I think this fix is a good idea (even though I actively did not include it originally) if only to make everyone's lives easier.  We should be aware that if anyone uses a different logging config and changes where the logging resides, this error could still occur.  I don't think there is anything reasonable we can do about that.  Soapbox over.

kdrumm (approved):

ThomasSchellenbergNextCentury (approved):

machine_common_sense/logging_config.py
```diff
@@ -39,6 +39,7 @@
 
 import ast
 import logging
+from os import mkdir
```
deanwetherby, line 42: As a senior dev, I prefer importing the package once `import os` and then using dot-notation to call the various functions in that namespaced package. Technically your import still works but when I see the `mkdir` in the code, I don't know if it's something custom we wrote or from a package without having to spend a cognitive load investigating.

jaudick: @deanwetherby I agree, that would definitely help with clarity. This has been updated

deanwetherby (approved):

PR #377, opened 2021-07-14
Mcs 797 auto download dev

kdrumm:
Added develop option (or dev or development) for version for auto-downloader
Added ability to set mcs-unity-version for auto-downloader for integration test scripts and AbstractScriptRunner scripts.  More can be added as needed.

### DIFF
--- a/machine_common_sense/unity_executable_provider.py
+++ b/machine_common_sense/unity_executable_provider.py
@@ -16,6 +16,8 @@ logger = logging.getLogger(__name__)
 
 LINUX_URL = "https://github.com/NextCenturyCorporation/MCS/releases/download/{ver}/MCS-AI2-THOR-Unity-App-v{ver}-linux.zip"  # noqa
 MAC_URL = "https://github.com/NextCenturyCorporation/MCS/releases/download/{ver}/MCS-AI2-THOR-Unity-App-v{ver}-mac.zip"  # noqa
+LINUX_DEV_URL = "https://ai2thor-unity-releases.s3.amazonaws.com/MCS-AI2-THOR-Unity-App-vdevelop-linux.zip"  # noqa
+MAC_DEV_URL = "https://ai2thor-unity-releases.s3.amazonaws.com/MCS-AI2-THOR-Unity-App-vdevelop-mac.zip"  # noqa
 
 
 class UnityExecutableProvider():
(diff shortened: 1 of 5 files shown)


kdrumm:
Note: This will cache the Dev version and never update it which requires the user to know they need to manually clear.  I don't like this, but I'm not sure how to check.  We could download the zip each time, but the download takes a while.  Maybe that's more viable once we move to addressables.  I'd like some opinions on this.

kdrumm:
Also, purposely left this as an undocumented feature at the moment.  If we can solve the caching issue, maybe we should change that.

scripts/runner_script.py
```diff
@@ -39,6 +39,7 @@ def __init__(
             config_file_path = args.config_file
         controller = mcs.create_controller(
             unity_app_file_path=args.mcs_unity_build_file,
+            unity_cache_version=args.mcs_unity_version,
```
ThomasSchellenbergNextCentury, line 42: Thanks for doing this!

machine_common_sense/unity_executable_provider.py
```diff
@@ -16,6 +16,8 @@
 
 LINUX_URL = "https://github.com/NextCenturyCorporation/MCS/releases/download/{ver}/MCS-AI2-THOR-Unity-App-v{ver}-linux.zip"  # noqa
 MAC_URL = "https://github.com/NextCenturyCorporation/MCS/releases/download/{ver}/MCS-AI2-THOR-Unity-App-v{ver}-mac.zip"  # noqa
+LINUX_DEV_URL = "https://ai2thor-unity-releases.s3.amazonaws.com/MCS-AI2-THOR-Unity-App-vdevelop-linux.zip"  # noqa
+MAC_DEV_URL = "https://ai2thor-unity-releases.s3.amazonaws.com/MCS-AI2-THOR-Unity-App-vdevelop-mac.zip"  # noqa
```
ThomasSchellenbergNextCentury, line 20: In response to the caching question: Is there any way you can easily retrieve (and cache) the "last modified" date from S3? Maybe that would work?

kdrumm: hmm.. good idea!  That'll probably work, only because we're storing this one in S3.  I don't think you can typically do that with  normal file downloads unless you download the file.

kdrumm: I added a bunch of code that uses the last-modified header of the request.

machine_common_sense/unity_executable_provider.py
```diff
@@ -292,3 +308,21 @@ def _do_download(self, url):
...
+            r.close()
+            dt = datetime.datetime.strptime(
+                last_mod,
+                "%a, %d %b %Y %H:%M:%S %Z")
+            return dt > date
+        except Exception as e:
+            logger.warn(
+                "Error checking last modified of development build",
+                exc_info=e)
+        return updated
```
deanwetherby, line 328: Does it make sense for this to be in a finally block? Then the try section could set the 'updated' variable and allow finally to return either the set variable or the default in the case of an exception.

JuanRMarquez-Ortiz (approved):

deanwetherby (approved):

PR #380, opened 2021-07-15
Added tutorial videos to the Adding 3D Models doc

ThomasSchellenbergNextCentury:
Pros:
- Now the tutorial videos are in the appropriate doc, where they belong, and not hidden on S3

Cons:
- Now anyone who clones the repo gets to download the videos
- Had to add a new python/sphinx dependency that's not on pypy
- ~~Had to manually copy over the tutorial videos since sphinx wasn't doing it automatically (maybe it ignores all video files?)~~


scripts/generate_sphinx_docs.sh
```diff
@@ -13,8 +13,10 @@ then
 fi
 echo "moving to docs"
 cd docs/
-echo "calling make'"
+echo "calling make"
 make html
+echo "copying tutorial videos into build folder"
+cp -r source/tutorial _build/html/_tutorial
```
deanwetherby, line 19: GH Actions isn't using this shell script so we have a discrepancy. Publishing the documentation out to GH Pages would likely break the video links.

ThomasSchellenbergNextCentury: That's good to know. Thanks for the info. I'll look into the GitHub action and some alternative solutions. (I had tried putting the videos in the `images` folder, but sphinx seems to ignore all video files? Or I'm missing something)

deanwetherby: You're not missing anything. This is awkward. I'm looking into a few things as well.

ThomasSchellenbergNextCentury: I imagine the "right" solution would be to get sphinx to copy over the videos itself, but, in lieu of that, maybe we could add an action to copy the files in our "doc" GitHub workflow (or an action to run a script that does so)? I'm assuming that's possible?

ThomasSchellenbergNextCentury:
Dean figured out how to copy over the tutorial videos using sphinx itself, rather than manually copying them using the script file. I've implemented and committed the change. Thanks Dean!

deanwetherby (approved):

PR #381, opened 2021-07-20
fixed autodownloader.  Added version to human input

kdrumm:
Tiny PR to fix date handling in auto-downloader.
Also added to human input.

### DIFF
--- a/machine_common_sense/unity_executable_provider.py
+++ b/machine_common_sense/unity_executable_provider.py
@@ -317,7 +317,7 @@ class Downloader():
             dt = datetime.datetime.strptime(
                 last_mod,
                 "%a, %d %b %Y %H:%M:%S %Z")
-            return dt > date
+            updated = dt > date
         except Exception as e:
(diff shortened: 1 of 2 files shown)


machine_common_sense/unity_executable_provider.py
```diff
@@ -317,7 +317,7 @@ def is_updated(self, url, date: datetime.datetime):
             dt = datetime.datetime.strptime(
                 last_mod,
                 "%a, %d %b %Y %H:%M:%S %Z")
-            return dt > date
+            updated = dt > date
```
ThomasSchellenbergNextCentury, line 320: Do we still need the `updated = True` line above?

ThomasSchellenbergNextCentury: Oh, is that there in case there's an exception before you set it again? I got it now.

ThomasSchellenbergNextCentury (approved):

PR #382, opened 2021-07-22
MCS-674 Removing evaluation-related code

deanwetherby:
AWS credentials, S3 buckets, evaluation mode, team name are all removed from the python api. This draft PR opens the door for debate on the merits of moving file uploads (history, logs, video recordings) to the pipeline. The python api will be focused on running scenes at various metadata levels and providing the json or imagery recordings for researchers to use.


kdrumm (commented):
I really like this change.  The only minor issue to work out is how to name the files properly.  Of course we can rename files in pipeline fairly easily.  Hopefully adding all this to the pipeline isn't that hard since it already has some upload capabilities.

deanwetherby:
@kdrumm File renaming can happen, as you suggested, in the pipeline itself. Since the pipeline is the authority on team and evaluations names, it seems apt to be handled there.

brianpippin:
So this code looks good to me, but we probably can't merge this until we have coverage in the pipeline for everything we are taking out correct?  Or at the least a way to track all of it?

deanwetherby:
@brianpippin I don't think so. It's all in git somewhere if we need to copy code to the pipeline.

kdrumm:
I would think this would be another case of parallel PRs in both repos (mcs and mcs-pipeline) that are both reviewed before either are merged.  Just like when we have changes in MCS and AI2Thor that are dependent on each other.

machine_common_sense/controller.py
```diff
@@ -391,9 +360,11 @@ def start_scene(self, config_data):
 
             # TODO Should this be in the if block?  Now that we are using
             # subscribers, we may want to always register
+            # if(self._end_scene_not_registered is True and
+            #         (self._config.is_history_enabled() or
+            #          self._config.is_evaluation())):
```
ThomasSchellenbergNextCentury, line 365: Is this code block still needed?

machine_common_sense/history_writer.py
```diff
@@ -143,16 +122,18 @@ def filter_history_output(
...
             history.output.structural_object_list = None
+            targets = ['target', 'target_1', 'target2']
             for target in targets:
-                if target in history.output.goal.metadata.keys():
-                    if history.output.goal.metadata[target].get(
-                            'image', None) is not None:
-                        del history.output.goal.metadata[target]['image']
+                if (
+                    target in history.output.goal.metadata.keys() and
+                    history.output.goal.metadata[target].get('image', None)
```
ThomasSchellenbergNextCentury, line 133: Isn't `None` the default for the 2nd argument?

ThomasSchellenbergNextCentury:
I am also in favor of this change. The code looks good -- you did a nice job cleaning up some things too.

deanwetherby:
> I am also in favor of this change. The code looks good -- you did a nice job cleaning up some things too.

I snuck some stuff in there.

machine_common_sense/history_writer.py
deanwetherby: It is. I didn't change anything from the previous code although I prefer explicit over implicit.

machine_common_sense/controller.py
deanwetherby: No it isn't

machine_common_sense/history_writer.py
```diff
@@ -26,19 +26,11 @@ def on_start_scene(self, payload):
             if self.__history_writer:
                 self.__history_writer.check_file_written()
 
-            hist_info = {}
-            hist_info[
-                payload.config.CONFIG_EVALUATION_NAME
-            ] = payload.config.get_evaluation_name()
-            hist_info[
-                payload.config.CONFIG_EVALUATION
```
rartiss55, line 34: Since I kind of forgot about this one and was mainly thinking of uploads + changing filenames later: 
We need to make sure evaluation_name and team name either get appended later to the history file in the pipeline somehow, or that we use some other mechanism to inform ingest about them, since its currently expecting them in the history file.

deanwetherby: Another thought it to remove those attributes (eval=true and eval_name) from the history log and provide those values elsewhere. I know that would change ingest and downstream but something to consider.

brianpippin (approved):

PR #383, opened 2021-07-26
MCS-819 plotter dimensions

deanwetherby:


kdrumm (commented):
Code looks good.  Need to run it now.

tests/test_plotter.py
```diff
@@ -11,7 +12,8 @@ class TestTopDownPlotter(unittest.TestCase):
     def setUp(self):
         self.plotter = TopDownPlotter(
             team="test",
-            scene_name="scene"
+            scene_name="scene",
+            room_size=Vector3d(x=10, y=4, z=10)
```
ThomasSchellenbergNextCentury, line 16: This would probably be rather excessive, but if you really want to test that the dimensions are being set correctly (again, definitely _not necessary_), you could create a mock plot object, inject the mock plot into the plotter, and assert that the mock plot's "xlim" and "ylim" functions are called with the expected parameters.

machine_common_sense/controller_media.py
```diff
@@ -186,9 +186,10 @@ class TopdownVideoEventHandler(AbstractVideoEventHandler):
     def on_start_scene(self, payload: ControllerEventPayload):
         self.__recorder = self.create_video_recorder(
             payload, AbstractVideoEventHandler.TOPDOWN)
-        team = payload.config.get_team()
-        scene = payload.scene_config.name.replace('json', '')
-        self.__plotter = TopDownPlotter(team, scene)
+        self.__plotter = TopDownPlotter(
+            team=payload.config.get_team(),
+            scene=payload.scene_config.name.replace('json', ''),
```
kdrumm, line 191: I don't know how we got here, but this doesn't work for me until scene=... is changed to scene_name=...

ThomasSchellenbergNextCentury, line 191: I'm getting an error: `unexpected keyword argument 'scene'`. Should this be `scene_name`?

kdrumm (approved):

ThomasSchellenbergNextCentury (approved):
Beautiful

PR #386, opened 2021-07-30
Hole Generation

jaudick:
Please run these tests. They appear to work now but I was getting 39 and 50 failing on oracle. I did some minor tweaks on the unity side and that seemed to fix it but they were troubling me for a day. https://github.com/NextCenturyCorporation/ai2thor/pull/156


deanwetherby (approved):

machine_common_sense/config_manager.py
```diff
@@ -359,6 +359,15 @@ def make_physics_config(self, data, **kwargs):
         return PhysicsConfig(**data)
 
 
+class HolesConfigSchema(Schema):
```
deanwetherby, line 362: Is there a hole depth to consider here?

kdrumm (approved):

machine_common_sense/config_manager.py
brianpippin: Also is the hole always the same size?

jaudick: The hole is always 1x1 is width dimensions. This is because the floor is divided into sections of 1x1. As of now the depth is constant and set up in unity. It is -6 units deep.

ThomasSchellenbergNextCentury (approved):

PR #387, opened 2021-08-03
Textures on specific areas of the floor

jaudick:
Unity: https://github.com/NextCenturyCorporation/ai2thor/pull/157
Screenshot from 2021-08-03 14-35-18


machine_common_sense/config_manager.py
```diff
@@ -359,13 +359,22 @@ def make_physics_config(self, data, **kwargs):
...
     x = fields.Int()
     z = fields.Int()
 
     @post_load
     def make_holes_config(self, data, **kwargs):
-        return HolesConfig(**data)
+        return FloorHolesAndTexturesXZConfig(**data)
+
+
+class FloorTexturesConfigSchema(Schema):
```
deanwetherby: Would it be better to leave holes as its own simple schema and then add x,z to the floor texture?

kdrumm (commented):
This looks good, but needs to reflect changes from the corresponding AI2Thor PR for FloorHolesAndTexturesConfig.

brianpippin (approved):

machine_common_sense/config_manager.py
jaudick: I dont think so because both the floorTextures and holes property within the scene jsons needs a list of xz coordinates. The only difference with textures is that those points have a reference to a string material name The process of setting up holes and textures also happens within the same method for generating the grid on the unity side. Separating them would be easy but I figured that anything grid coordinate related should be together

kdrumm (approved):

ThomasSchellenbergNextCentury (approved):

PR #389, opened 2021-08-05
MCS-821 - Publish PyPI release Github Action

rartiss55:
I had an additional test branch to test pushing up to Test PyPI. I didn't want to bother anyone else with it since testing Test PyPI is kind of a pain (I can delete that branch though when this is merged).

Tacked the PyPI stuff onto the existing docs.yaml. Keep in mind: I'm not 100% sure this is correct, since this version will only trigger when we tag a release.


.github/workflows/docs-publish-pypi.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Install Python 3
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-versions }}
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip setuptools wheel
+          python -m pip install -r requirements.txt
+      - name: create wheel file
+        run: python setup.py bdist_wheel sdist clean --all
```
rartiss55, line 42: Hopefully I pruned this down enough -- realized we don't really need to run the tests/linter stuff again at this point since it already would've failed.

.github/workflows/docs-publish-pypi.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip setuptools wheel
+          python -m pip install -r requirements.txt
+      - name: create wheel file
+        run: python setup.py bdist_wheel sdist clean --all 
+      - name: Publish distribution to Test PyPI
+        uses: pypa/gh-action-pypi-publish@master
+        with:
+          password: ${{ secrets.TEST_PYPI_API_TOKEN }}
```
rartiss55, line 46: I added the secrets to the github project settings, so this part will magically work.

deanwetherby (approved):

PR #390, opened 2021-08-06
MCS-751 plotter performance

deanwetherby:
Increase plotter performance by replacing slow plot libraries such as matplotlib with direct image manipulation through scikit-image drawing.


deanwetherby:
Suggest using 'Files Changed' tab to review the final state as some of these commits/pushes were work in progress (WIP). When testing, ensure that the 'roomDimensions' field is present in your scene file as seen below.

```
  "roomDimensions": {
    "x": 10,
    "y": 4,
    "z": 10
  },
  ```

File: machine_common_sense/plotter.py [hunk 1]
```diff
@@ -1,32 +1,34 @@
...
 import math
-
-import ai2thor
-import matplotlib
-import PIL
+from dataclasses import dataclass
+from typing import Dict, List
+
+import ai2thor.server
+import colour
```
deanwetherby, line 6: Using this package to convert string colors like 'magenta' into RGB values. A couple of thoughts on that:

1) We could include RGB in the metadata and get that information from Unity
2) Rather than having `colour` as a package dependency, the package is small enough that we could vendorize it into the MCS code base

machine_common_sense/plotter.py
```diff
@@ -1,32 +1,34 @@
...
 
-matplotlib.use('Agg')
-from typing import Dict, List, NamedTuple
-
-import matplotlib.pyplot as plt
-from shapely import geometry
 
-
-class XZHeading(NamedTuple):
+@dataclass
```
deanwetherby, line 16: Switched to using dataclasses now that we're using it elsewhere. The NamedTuple approach was the poor man's dataclasses in <3.6.

machine_common_sense/plotter.py
```diff
@@ -36,30 +38,184 @@ class Object(NamedTuple):
...
+    ROBOT_COLOR = colour.COLOR_NAME_TO_RGB['red']
+    GOAL_COLOR = colour.COLOR_NAME_TO_RGB['gold']
+
     ROBOT_PLOT_WIDTH = 0.2
-    ROBOT_PLOT_LABEL = "robot"
-    ROBOT_COLOR = 'xkcd:gray'
-    DEFAULT_COLOR = "xkcd:black"
     HEADING_LENGTH = 0.4
-    BORDER = 0.05
+    PLOT_IMAGE_SIZE = 512
```
deanwetherby, line 51: Fixing the plotter image size to 512. This could be configuration-driven but changes a bit of the underlying plot drawing. Although most of it works using different dimensions, the text scale is a bit more difficult to deal with. Recommend we keep this fixed and unexposed until later.

kdrumm (approved):
LGTM.

deanwetherby:
https://user-images.githubusercontent.com/26032066/128550285-39cf9701-3c9b-4a89-8e3e-8dd09c918abb.mp4

PR #392, opened 2021-08-11
MCS-817 remove heatmap

deanwetherby:


tests/test_mcs.py
```diff
@@ -38,9 +37,12 @@ def return_false():
         cfg.is_video_enabled = return_false
         add_subscribers(ctrl, cfg)
         self.assertEqual(len(ctrl._subscribers), 2)
-        classes = (ControllerLogger, HistoryEventHandler)
-        self.assertIsInstance(ctrl._subscribers[0], classes)
-        self.assertIsInstance(ctrl._subscribers[1], classes)
+        self.assertTrue(
+            any([isinstance(subscriber, ControllerLogger)
+                 for subscriber in ctrl._subscribers]))
```
deanwetherby, line 42: The new test goes like this:
1. list comprehension returns a list of booleans resulting from the isinstance call on each element of the list
2. If any of the list elements are instances of the class, then one element in the new list will be True
3. The any() call returns true if any element is True.
4. We assert that the condition from any is indeed True and the test passes.

deanwetherby: This means we don't have check specific indices for classes and documents that order of event handlers shouldn't matter.

kdrumm (approved):
LGTM

PR #393, opened 2021-08-13
Mcs 844 cache addressables docs

kdrumm:
Adding some documentation and instructions to cache addressables script.

### DIFF
--- a/scripts/cache_addressables.py
+++ b/scripts/cache_addressables.py
@@ -1,5 +1,5 @@
+import argparse
 import os
-import sys
 
 '''
 Script will start Unity build with a flag to clear the addressables cache,
@@ -18,9 +18,21 @@ executables.
 '''
 
 
+def parse_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('build_path',
+                        help="Unity build path")
+    return parser.parse_args()
+
+
 def run_caching_build(build_path):
     os.execl(build_path, build_path, "CACHEADDRESSABLES")
 
 
-build_path = sys.argv[1]
-run_caching_build(build_path)
+def main():
+    args = parse_args()
+    run_caching_build(args.build_path)
+
+
+if __name__ == '__main__':
+    main()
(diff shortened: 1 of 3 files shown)


JuanRMarquez-Ortiz (approved):

scripts/cache_addressables.py
```diff
@@ -1,6 +1,22 @@
 import os
 import sys
 
+'''
+Script will start Unity build with a flag to clear the addressables cache,
+cache all the data in catalog and then close the application.
+
+Usage: python cache_addressables.py <unity-build-path>
```
deanwetherby, line 8: If a researcher/developer installs MCS from PyPI, do they have access to this script?

deanwetherby (commented):
I think the cache_addressables script needs to be added to console_scripts in setup.py that way it's available in venv/bin/ to run from the command line.

scripts/cache_addressables.py
deanwetherby: Confirmed that the only scripts in the venv when installing from PyPI are those specified in the console_scripts. Those scripts are installed in venv/bin/.

setup.py
```diff
@@ -47,7 +47,8 @@ def get_version():
     entry_points={
         'console_scripts': [
             'run_in_human_input_mode=scripts.run_human_input:main',
-            'run_scene_timer=scripts.run_scene_timer:main'
+            'run_scene_timer=scripts.run_scene_timer:main',
+            'cache_addressables=scripts/cache_addressables:main'
```
deanwetherby, line 51: Adding the script to entry_points will put the script in venv/bin when installed from PyPI. With an activate python virtual environment, the developer could call the script from the command line via:

`cache_addressables <MCS_path>`

Without explicitly adding this script to entry_points, an MCS installation from PyPI will not install the script.

rartiss55: Good catch!

rartiss55: If this is the case, the line about this script in install.rst should be updated too.

JuanRMarquez-Ortiz (approved):

rartiss55 (approved):

PR #394, opened 2021-08-13
Providing scene default room dimensions

deanwetherby:


kdrumm (approved):
LGTM - Of course we already discussed that we need to eventually resolve having defaults in 2 different places to avoid changing one and not the other.

machine_common_sense/config_manager.py
```diff
@@ -681,7 +683,8 @@ class SceneConfiguration:
     objects: List[SceneObject] = field(default_factory=list)
     observation: bool = False  # deprecated; please use intuitivePhysics
     performerStart: PerformerStart = None
-    roomDimensions: Vector3d = None
+    roomDimensions: Vector3d = field(
+        default=ConfigManager.DEFAULT_ROOM_DIMENSIONS)
```
ThomasSchellenbergNextCentury, line 687: Should the `DEFAULT_ROOM_DIMENSIONS` be wrapped in a `copy.deepcopy()` here? I'm assuming `field()` doesn't automatically copy the default value, which means we're setting the `roomDimensions` of each `SceneConfiguration` object to the same `Vector3d` object. Therefore if our code ever changes the inner properties of a `roomDimensions`, it's changing the `DEFAULT_ROOM_DIMENSIONS` as well, which will affect each current and future `SceneConfiguration`.

deanwetherby: Why would the room dimensions change mid-scene?

ThomasSchellenbergNextCentury: Who knows what tasks the psych team has planned for future evals? All I was trying to say is that wrapping this in a `copy.deepcopy` seems like it could save a lot of debugging headache later down the line.

deanwetherby: YAGNI

PR #395, opened 2021-08-20
MCS-863 Ensure outer walls are shown properly in top-down videos.

ThomasSchellenbergNextCentury:
Changes:
- Previously, the room's walls would not always appear in the top-down videos because they're positioned outside the room bounds. For example, in a default room (X=[-5, 5], Z=[-5, 5]), the front wall's Z position is between 5 and 6. Now, I added a buffer to the X and Z scales so the walls would be shown properly.
- I reduced the size of the room walls so they don't take up as much real estate in the top-down videos.

Here are some examples of an 8x12 room with three blue walls and one red wall.

OLD (you can barely see the right wall, and can't even see the left wall!):
MCS-863-old

NEW:
MCS-863-new

### DIFF
--- a/machine_common_sense/controller_media.py
+++ b/machine_common_sense/controller_media.py
@@ -5,7 +5,7 @@ from abc import abstractmethod
 import numpy as np
 import PIL
 
-from .config_manager import ConfigManager
+from .config_manager import ConfigManager, Vector3d
 from .controller_events import (AbstractControllerSubscriber,
                                 ControllerEventPayload)
 from .plotter import TopDownPlotter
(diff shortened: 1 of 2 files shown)


deanwetherby (commented):
Have you tried a square room? I don't think it's working in all cases. Wish I had more time to delve into this.

Screenshot from 2021-08-20 14-56-36

ThomasSchellenbergNextCentury:
@deanwetherby Square rooms should work, but intuitive physics scenes (with `"intuitivePhysics": true`) automatically resize the room, which is why you're not seeing all of the walls. I'm not sure if it's terribly important to see all the walls in intuitive physics scenes (just the wall you're stuck facing)?

brianpippin:
The code looks good to me, I don't know if we can make the plot slightly larger for intuitive physics scenes so we can see the outside walls?  But everything coded LGTM.

ThomasSchellenbergNextCentury:
@brianpippin Yeah, I was considering that too, but I actually think Unity is moving the side walls out a bit too farther than is necessary, so I wanted to play around with it some more (now that the builds are working).

rartiss55 (approved):

deanwetherby:
I'm taking a closer look at this today. I think there's a better way to solve this.

deanwetherby:
I'm going to assume that this PR inherits the approval from the other PR and merge into the development branch.

PR #397, opened 2021-09-01
MCS-868 Update Python API Doc to Use Eval 4 Plan Terminology

ThomasSchellenbergNextCentury:
- In `end_scene` and `make_step_prediction`, renamed `choice` to `rating` and `confidence` to `score`. Updated corresponding docstrings with expected variable types, required vs. optional arguments, and valid input. Did not rename any scene history file properties, so ingest should not be affected.
- Updated Agents and Intuitive Physics GoalCategory documentation with the same information.

Would probably be good to merge this before releasing 0.4.4, unless there are any major concerns.

Eval 4 Plan: https://nextcentury.atlassian.net/wiki/spaces/MCSC/pages/1874985098/Eval+4+Plan+Final#2.1-Passive-Observational-Physics-(VOE)-Tasks

Fixes GitHub Issue: https://github.com/NextCenturyCorporation/MCS/issues/237

Issue #237, opened 2021-02-08
Changes to Classification (Choice) and Confidence


ThomasSchellenbergNextCentury:
From: TA2

Priority: Eval 4

Components: Python API (`make_step_prediction`, `end_scene`)

Blockers: Needs design (TA2) and approval (TA1)

Issues:
- The term "confidence" has a specific meaning in machine learning, which has lead to some confusion
- We may want to encourage teams to provide a confidence value rather than rely on a default value
- We may want to provide feedback (like an error message) if an incorrectly-formatted confidence value is passed
- We want the API to support both VoE and future non-binary classification tasks in a clear way

Possible Solutions:
- Rename `confidence` to `score` (or `plausibility_score`, or `continuous_score`), to mimic the terminology used in the Eval 4 Plan
- Rename `choice` to `rating` (or `plausibility_rating`, or `binary_rating`), to mimic the terminology used in the Eval 4 Plan
- Update the documentation in GoalMetadata to describe the expectations for intuitive physics versus agents as detailed in the Eval 4 Plan
- Reject function calls with incorrectly-formatted data based on the goal category

TA2 internal JIRA ticket: https://nextcentury.atlassian.net/browse/MCS-513

deanwetherby:
https://machinecommonsense.slack.com/archives/CTR31V12M/p1615651333007600

Ben Zinberg Yesterday at 11:02 AM
Confirming that the confidence kwarg to Controller.end_scene has the meaning of 0 = sure implausible, 1 = sure plausible, 0.5 = totally uncertain (and continuous partial uncertainty in between)?  (If so, perhaps we should rename that kwarg to something like soft_choice) (edited)

ThomasSchellenbergNextCentury:
Related to: https://github.com/NextCenturyCorporation/MCS/issues/263

rartiss55:
Resolved in the previously linked PR + included in the 0.4.4 release.

PR #398, opened 2021-09-02
MCS-513 (WIP for retrospective reporting)

rartiss55:
Need to wait for input from @deanwetherby on the TODOs (whether or not to get rid of make_step_prediction/if we'd like to maintain backwards compatibility, as well as anything we'd like to rename that could affect ingest). 

Spent the most time with unit tests here/trying to mock things with events. The test_history_event_handler.py may have been overkill in hindsight.

### DIFF
--- a/machine_common_sense/controller.py
+++ b/machine_common_sense/controller.py
@@ -135,12 +135,6 @@ class Controller():
     OBJECT_MOVE_ACTIONS = ["CloseObject", "OpenObject"]
     MOVE_ACTIONS = ["MoveAhead", "MoveLeft", "MoveRight", "MoveBack"]
 
-    AWS_CREDENTIALS_FOLDER = os.path.expanduser('~') + '/.aws/'
-    AWS_CREDENTIALS_FILE = os.path.expanduser('~') + '/.aws/credentials'
-
-    AWS_ACCESS_KEY_ID = 'aws_access_key_id'
-    AWS_SECRET_ACCESS_KEY = 'aws_secret_access_key'
-
     def __init__(self, unity_app_file_path, config):
 
         self._subscribers = []
--- a/machine_common_sense/controller.py
+++ b/machine_common_sense/controller.py
@@ -241,7 +241,7 @@ class Controller():
         if self.__seed:
             random.seed(self.__seed)
 
-    def end_scene(self, choice, confidence=1.0, report={}):
+    def end_scene(self, choice, confidence=1.0, report=None):
         """
         Ends the current scene.  Calling end_scene() before calling
         start_scene() will do nothing.
@@ -261,27 +261,38 @@ class Controller():
             end_scene isn't properly called but history_enabled is true,
             this value will be written to file as -1.
         report : Dict[integer, object], optional
-            Variable for retrospective per frame reporting.
+            Variable for retrospective per frame reporting. (default None)
+
             Key is frame number take from step metadata (step number
             starts at 1). Value or payload contains:
-            - choice : string, optional
-            The selected choice for per frame prediction with
-            violation-of-expectation or classification goals.
-            Is not required for other goals. (default None)
-            - confidence : float, optional
-            The choice confidence between 0 and 1 required by the end
-            of scenes with violation-of-expectation or classification
-            goals. Is not required for other goals. (default None)
-            - violations_xy_list : List[Dict[str, float]], optional
-            A list of one or more (x, y) locations
-            (ex: [{"x": 1, "y": 3.4}]),
-            each representing a potential violation-of-expectation.
-            Required on each step for passive tasks. (default None)
-            - internal_state : object, optional
-            A properly formatted json object representing various
-            kinds of internal states at a particular moment. Examples
-            include the estimated position of the agent, current map
-            of the world, etc. (default None)
+
+                * choice : string, optional -
+                  The selected choice for per frame prediction with
+                  violation-of-expectation or classification goals.
+                  Is not required for other goals.
+                * confidence : float, optional -
+                  The choice confidence between 0 and 1 required by the end
+                  of scenes with violation-of-expectation or classification
+                  goals. Is not required for other goals.
+                * violations_xy_list : List[Dict[str, float]], optional -
+                  A list of one or more (x, y) locations
+                  (ex: ```[{"x": 1, "y": 3.4}]```),
+                  each representing a potential violation-of-expectation.
+                  Required on each step for passive tasks.
+                * internal_state : object, optional -
+                  A properly formatted json object representing various
+                  kinds of internal states at a particular moment. Examples
+                  include the estimated position of the agent, current map
+                  of the world, etc.
+
+            Example report:
+            ```{1: {
+            "choice": "plausible",
+            "confidence": .75,
+            "violations_xy_list": [{"x": 1,"y": 1}],
+            "internal_state": {"test": "some state"}
+            }}
+            ```
 
         """
         payloadArgs = self._create_event_payload_kwargs()
--- a/machine_common_sense/controller.py
+++ b/machine_common_sense/controller.py
@@ -241,6 +241,8 @@ class Controller():
         if self.__seed:
             random.seed(self.__seed)
 
+    # TODO: MCS-513: do we need to update docs here + make_step_prediction/
+    # specify scores are for Agent tasks as well?
     def end_scene(self, choice, confidence=1.0, report=None):
         """
         Ends the current scene.  Calling end_scene() before calling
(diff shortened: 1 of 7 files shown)


machine_common_sense/controller.py
```diff
@@ -266,10 +260,45 @@ def end_scene(self, choice, confidence=1.0):
...
+                  Is not required for other goals.
+                * confidence : float, optional -
+                  The choice confidence between 0 and 1 required by the end
+                  of scenes with violation-of-expectation or classification
+                  goals. Is not required for other goals.
+                * violations_xy_list : List[Dict[str, float]], optional -
+                  A list of one or more (x, y) locations
+                  (ex: ```[{"x": 1, "y": 3.4}]```),
+                  each representing a potential violation-of-expectation.
+                  Required on each step for passive tasks.
```
ThomasSchellenbergNextCentury, line 281: If/when we merge https://github.com/NextCenturyCorporation/MCS/pull/397, we should update this section to use the new wording. (Also, sorry for the potential merge conflicts -- I wasn't thinking.)

machine_common_sense/controller.py
```diff
@@ -266,10 +260,45 @@ def end_scene(self, choice, confidence=1.0):
...
+                  include the estimated position of the agent, current map
+                  of the world, etc.
+
+            Example report:
+            ```{1: {
+            "choice": "plausible",
+            "confidence": .75,
+            "violations_xy_list": [{"x": 1,"y": 1}],
+            "internal_state": {"test": "some state"}
+            }}
```
ThomasSchellenbergNextCentury, line 294: This is super nitpicky, but can we make the indentation here prettier?

machine_common_sense/controller.py
```diff
@@ -619,6 +648,8 @@ def step(self, action: str, **kwargs) -> StepMetadata:
 
         return output
 
+    # TODO: MCS-513: Do we remove this or keep for backwards
+    # compatability/teams that are fine reporting frame by frame?
```
ThomasSchellenbergNextCentury, line 652: I think backwards compatibility is good. Also, I know that the Eval 4 Plan specifically [says](https://nextcentury.atlassian.net/wiki/spaces/MCSC/pages/1874985098/Eval+4+Plan+Final#2.1-Passive-Observational-Physics-(VOE)-Tasks) (under bullet point 1):

_"But the scores will be reported only after the end of the scene (i.e., there will be no option to report a frame-by-frame score immediately after seeing a given frame)."_

However, do we really need to prohibit the use of `make_step_prediction`? Would it be hard to have both `make_step_prediction` and the new `report` variable save the data the same way on our end?

machine_common_sense/controller.py
```diff
@@ -266,10 +260,45 @@ def end_scene(self, choice, confidence=1.0):
             Note: when an issue causes the program to exit prematurely or
             end_scene isn't properly called but history_enabled is true,
             this value will be written to file as -1.
+        report : Dict[integer, object], optional
+            Variable for retrospective per frame reporting. (default None)
```
ThomasSchellenbergNextCentury, line 264: I recommend that we be abundantly clear here that the `report` is only needed for non-agent passive / VoE scenes.

brianpippin: We also need to Brendan about what exactly he wants for the Agent tasks, I think those should definitely have a confidence score at the end, so we need to make sure they actually deliver that unlike in Eval 3.

machine_common_sense/controller.py
brianpippin: I good go either way on this one, I think the path of least resistance from TA1 and what makes sense for keeping our code clean is the best way to go.

brianpippin:
I think all the code looks great, it's more of the other unknowns to be figured out that we need to button up.

machine_common_sense/controller.py
rartiss55: So this is why I had the other TO DO, and I was unaware of that github issue. I am still struggling with the terminology score + rating tbh, but its in the plan and if it works for TA1 that's fine :)

Additionally, depending on what we rename these things to/what values we accept, we may want to update their titles in the history file and therefore ingest, so something to consider.

machine_common_sense/controller.py
rartiss55: I updated the description but didn't add "non-agent" since it sounds like we may want scores for those (added a TODO to check about that).

machine_common_sense/controller.py
rartiss55: Hah I had it prettier and then sphinx yelled at me. I updated it -- turns out I needed a bunch of spaces.

machine_common_sense/controller.py
ThomasSchellenbergNextCentury: @brianpippin I'm pretty sure we already clarified this as a group when we wrote the [Eval 4 Plan](https://nextcentury.atlassian.net/wiki/spaces/MCSC/pages/1874985098/Eval+4+Plan+Final#2.1-Passive-Observational-Physics-(VOE)-Tasks), since we're specifically only asking for the rating (in other words, a single value):

_In the Agency tasks (see section 2.2, below), the AI systems will only be required to return **continuous plausibility ratings**..._

Mostly pointing this out because, if we decide we now need them to return TWO values, that's different what we've told them before, and we'll need to communicate the new requirement appropriately.

brianpippin: Doesn't "continuous plausibility ratings" mean reporting per frame?   If it only means at the end then that is fine by me.  I know they need to have a final plausibility score because in Agency they get it correct if they say which one is more plausible than the other.

ThomasSchellenbergNextCentury: @brianpippin It's only at the end (sorry, I cut off the rest of the quote):

_In the Agency tasks (see section 2.2, below), the AI systems will only be required to return continuous plausibility ratings **evaluating the event as a whole after it has concluded**_

So they don't need to report ANYTHING frame-by-frame on the agent scenes.

brianpippin: I am fine with that solution, that word "continuous" is very confusing to me though.  But I am okay with only getting one final plausibility score.

ThomasSchellenbergNextCentury:
I've merged my changes from https://github.com/NextCenturyCorporation/MCS/pull/397:

- In `end_scene` and `make_step_prediction`, renamed `choice` to `rating` and `confidence` to `score`. Updated corresponding docstrings with expected variable types, required vs. optional arguments, and valid input. Did not rename any scene history file properties, so ingest should not be affected.
- Updated Agents and Intuitive Physics GoalCategory documentation with the same information.

Would probably be good to merge this before releasing 0.4.4, unless there are any major concerns.

Eval 4 Plan: https://nextcentury.atlassian.net/wiki/spaces/MCSC/pages/1874985098/Eval+4+Plan+Final#2.1-Passive-Observational-Physics-(VOE)-Tasks

Fixes GitHub Issue: https://github.com/NextCenturyCorporation/MCS/issues/237

machine_common_sense/controller.py
```diff
@@ -247,29 +241,97 @@ def _set_config(self, config):
...
+                    (ex: [{"x": 1, "y": 3.4}]), each representing a potential
+                    violation-of-expectation. These locations are required for
+                    all passive / VoE scenes except agent scenes. Not used for
+                    any interactive scenes or passive agent scenes.
+                    (default None)
+                * internal_state : object, optional
+                    A properly formatted json object representing various kinds
+                    of internal states at a particular moment. Examples include
+                    the estimated position of the agent, current map of the
+                    world, etc. (default None)
```
ThomasSchellenbergNextCentury, line 310: This part is directly copied from the docstring of `make_step_prediction`. Seemed better to copy it (and, unfortunately, have to maintain it) than to leave it vague and possibly confusing, but I'm open to feedback.

machine_common_sense/history_writer.py
```diff
@@ -76,8 +76,28 @@ def on_end_scene(self, payload):
             payload.config.is_history_enabled()
         ):
             self.__history_writer.add_step(self.__history_item)
+
+            # Loop back and fill out previous steps with
+            # retrospective report
+            # TODO: MCS-513: Kept the same property names as before.
+            # Did we want to rename any of these properties?
+            # If so we will need to update ingest.
```
ThomasSchellenbergNextCentury, line 84: Doing this seems like it would complicate backwards compatibility, so I'm leaning "no" on this, but could be persuaded otherwise.

ThomasSchellenbergNextCentury (approved):

machine_common_sense/history_writer.py
rartiss55: Looking at this again, I think we should leave it alone in ingest. Also, in the UI, we call "score" whether or not they got it correct, so that would be more confusing if we had two fields we called "score".

ThomasSchellenbergNextCentury: Good point: we'd need to differentiate between "TA1 score" and "TA2 score"... Seems like this would cause a domino effect lol

rartiss55:
Looks like Thomas already approved, but I took out make_step_prediction (which also means we don't need the on_prediction event related code either) and accounted for steps as string keys in "report" for the end_scene call. I think everything should be resolved now.

deanwetherby:
LGTM. Removing the step prediction cleaned up some of the code too. bonus.

brianpippin (approved):
LGTM - also as long as the end result is the same as the current history file (which it looks like it is), we have no updates to ingest/ui to make until we add the old plausibility graph back in.

PR #400, opened 2021-09-02
MCS-857 config improvements

rartiss55:
So we can get on the same page internally before putting something in front of TA1:

Here I'm proposing an optional config_dict that can be passed in, while keeping the config environment variable. Added this to the docs -- this would be the order of precedence:

        1. MCS_CONFIG_FILE_PATH environment variable (meant for internal
           TA2 use)
        2. If no environment variable given, use config_file_path property
           passed to create_controller
        3. config_dict property passed to create_controller, if the above
           filepaths are not specified or do not exist


machine_common_sense/config_manager.py
```diff
@@ -47,7 +47,6 @@ class ConfigManager(object):
 
     CONFIG_FILE_ENV_VAR = 'MCS_CONFIG_FILE_PATH'
     METADATA_ENV_VAR = 'MCS_METADATA_LEVEL'
-    DEFAULT_CONFIG_FILE = './mcs_config.ini'
```
rartiss55, line 50: We used to have this default option here -- I took it out since I think it makes things more confusing.

kdrumm:
This looks good to me.  A stretch goal may be for property by property overriding through the precedence, although I don't think its necessary and may end up being more confusing.

machine_common_sense/config_manager.py
```diff
@@ -67,26 +66,28 @@ class ConfigManager(object):
     SCREEN_WIDTH_DEFAULT = 600
     SCREEN_WIDTH_MIN = 450
 
-    def __init__(self, config_file_path=None):
+    def __init__(self, config_file_path=None, config_dict=None):
```
deanwetherby, line 69: Part of me is wondering if we can't have a single config parameter. Internally, we can check if it's an instance of string or a dictionary and handle appropriately.

machine_common_sense/config_manager.py
```diff
@@ -67,26 +66,28 @@ class ConfigManager(object):
...
         self._config = configparser.ConfigParser()
-        if os.path.exists(self._config_file):
+        if self._config_file is not None and os.path.exists(self._config_file):
             self._config.read(self._config_file)
             logger.info('Config File Path: ' + self._config_file)
+        elif config_dict is not None:
+            self._config[self.CONFIG_DEFAULT_SECTION] = config_dict
+            logger.info('No config file given or file path does not exist,'
+                        ' using config dictionary')
+            logger.info('Read in config dictionary: ' + str(config_dict))
```
ThomasSchellenbergNextCentury, line 88: Is there a chance this could print out secret information, like passwords? Since we're not using the config file to hold AWS credentials for eval runs any more, I think the answer is "No", but just double checking.

machine_common_sense/config_manager.py
ThomasSchellenbergNextCentury: I have no opinion, but whatever we choose, we should add a docstring here, since there isn't one.

machine_common_sense/config_manager.py
rartiss55: Good point -- I think the way things are now, this isn't a concern, but something to keep in the back of our minds as we move forward with logging.

tests/test_config_manager.py
```diff
@@ -33,24 +33,98 @@ def tearDownClass(cls):
...
+    @patch.multiple(ConfigManager, _read_in_config_dict=DEFAULT,
+                    _read_in_config_file=DEFAULT)
+    def test_init_with_arg(self, _read_in_config_dict, _read_in_config_file):
         file_path = './arg-test.ini'
-        config_mngr = ConfigManager(file_path)
+        _ = ConfigManager(file_path)
 
-        self.assertEqual(config_mngr._config_file, file_path)
+        _read_in_config_file.assert_called_once_with(file_path)
+        _read_in_config_dict.assert_not_called()
```
ThomasSchellenbergNextCentury, line 46: Neat mocking!

ThomasSchellenbergNextCentury (dismissed):

tests/test_config_manager.py
deanwetherby: @rartiss55 taught me that. In general I'm not a fan of mocking but this particular case is useful.

brianpippin (dismissed):

rartiss55:
LGTM!

machine_common_sense/config_manager.py
```diff
@@ -67,26 +66,41 @@ class ConfigManager(object):
...
+        # then look at config_file_or_dict from constructor
+        try:
+            if (os.getenv(self.CONFIG_FILE_ENV_VAR) is not None):
+                self._read_in_config_file(os.getenv(self.CONFIG_FILE_ENV_VAR))
+            elif (isinstance(config_file_or_dict, dict)):
+                self._read_in_config_dict(config_file_or_dict)
+            elif(isinstance(config_file_or_dict, str)):
+                self._read_in_config_file(config_file_or_dict)
+            else:
+                logger.warning("No config options given.")
```
rartiss55, line 85: Looking at this again -- I think there was talk last week about possibly throwing an exception if there are no config options given? Now that there is a config dict option, I'm leaning more towards the "error out in this case" scenario.

scripts/run_human_input.py
```diff
@@ -29,9 +29,10 @@ def parse_args():
         '--config_file_path',
         type=str,
         default=None,
+        required=True,
```
deanwetherby, line 32: Since the configuration is required (either env, file, or dict), making this argument required seems like the easiest way to do the same in this script. We could have an internal dictionary and remove the arg as another option.

scripts/runner_script.py
```diff
@@ -79,6 +79,7 @@ def _read_args(self):
             '--config_file',
             type=str,
             default=None,
+            required=True,
```
deanwetherby, line 82: config file is required for all scripts leveraged by the runner script.

machine_common_sense/__init__.py
```diff
@@ -70,25 +70,15 @@ def init_logging(log_config=None,
         log_config_file=log_config_file)
 
 
-def create_controller(unity_app_file_path=None,
-                      unity_cache_version=None,
-                      config_file_or_dict=None):
+def create_controller(config_file_or_dict=None,
+                      unity_app_file_path=None,
+                      unity_cache_version=None):
```
deanwetherby, line 75: changed the ordering so that config_file_or_dict came first since it is the only required argument.

machine_common_sense/controller.py
```diff
@@ -86,18 +86,8 @@ class Controller():
...
-
-        * Note the **order of precedence for config options**, in case more
-          than one is given:
-
-        1. **MCS_CONFIG_FILE_PATH** environment variable (meant for internal
-           TA2 use during evaluation)
-        2. If no environment variable given, use **config_file_or_dict**
-           parameter.
+    unity_app_file: str
+    config: ConfigManager
```
deanwetherby, line 89: Worried about developers creating a Controller from it's constructor, I realized that the documentation was not correct here. The config argument is actually a ConfigManager not a file or dict. That's good because all of the logic to check that variable is in the package init create_controller.

machine_common_sense/controller.py
```diff
@@ -149,7 +139,7 @@ class Controller():
     AWS_ACCESS_KEY_ID = 'aws_access_key_id'
     AWS_SECRET_ACCESS_KEY = 'aws_secret_access_key'
 
-    def __init__(self, unity_app_file_path, config):
+    def __init__(self, unity_app_file_path: str, config: ConfigManager):
```
deanwetherby, line 135: Added type hints here to be clear(er).

docs/source/examples.rst
```diff
@@ -3,18 +3,18 @@ Example Usage
...
+Please note that for the below example, make sure `MCS_CONFIG_FILE_PATH` is not set, since it will override the `config_file_or_dict` specified below and is meant more for internal use by TA2 during evaluation:
 
 .. code-block:: python
 
     import machine_common_sense as mcs
 
     # Specify a location for the Unity app as opposed to downloading it automatically
     controller = mcs.create_controller(unity_app_file_path='./some-path/unity-app', 
-                                       config_file_path='./some-path/config.ini')
+                                       config_file_or_dict='./some-path/config.ini')
```
ThomasSchellenbergNextCentury, line 17: We'll want to note this change in next version's changelog, since anyone using a keyword argument will need to update.

ThomasSchellenbergNextCentury (approved):
Looks good to me. Also, now I know that you can dismiss stale reviews!

rartiss55:
@deanwetherby Pushed a tiny commit just to update various scripts/examples with an empty config parameter in create_controller, and on second glance, I think the docs are fine as is. LGTM!

PR #402, opened 2021-09-08
Alternate wall buffering

deanwetherby:


brianpippin (approved):

ThomasSchellenbergNextCentury:
Dean -- thanks for putting this together. It's always good to review all the options.

Since we're (I'm) trying to be better at defining ticket goals, I think the goals for this ticket are:
- Primarily, show the walls in the top-down plots
- Secondarily, make it pretty (which is obviously a matter of personal opinion)

I did a comparison of my original branch with Dean's new branch (this PR) using the recently released [interactive object permanence](https://github.com/NextCenturyCorporation/MCS/blob/development/docs/source/scenes/eval_4_intobjperm_validation_01.json) and [reorientation](https://github.com/NextCenturyCorporation/MCS/blob/development/docs/source/scenes/eval_4_reor_validation_01.json) validation scenes.

Here's interactive object permanence:
| Original PR | This PR |
| ------ | ---- |
| thomas_intobjperm | dean_intobjperm |

And reorientation:
| Original PR | This PR |
| ------ | ---- |
| thomas_reor | dean_reor |

So both solutions achieve the primary goal, which is the important part. I think my solution is a bit prettier (YMMV), but requires that unfortunate `_reduce_bounds_size_if_wall` function. I'd be interested in the rest of the team's thoughts on this output/tradeoff, because I could see us going either way.

kdrumm:
I think both are sufficient.  I don't like the "glitchiness" of the original PR one.  By that I mean the inconsistent buffer between the white and gray lines along the walls and the boxes(object) that covers that in certain parts.  I don't like the wall thickness inconsistency for this PR.  Is there an easy way to adjust either?  Overall, I prefer this PR.

deanwetherby:
Thanks for doing that comparison @ThomasSchellenbergNextCentury.  While I do agree that the original PR looks better, I do worry that the walls will overlap with objects in the room. I'm not a fan of the function required to move the walls either. From a code maintenance and readability perspective, I much prefer the simpler solution presented here.

@kdrumm The inconsistent walls could be handled by increasing the size of the wall buffer. The thickness of the walls is a direct reflection of those assets in the Unity environment.

kdrumm:
If its inconsistent due to being accurate, then we should let it be accurate.

deanwetherby:
I increased the wall buffer to 50.

Screenshot from _none__wall_ahead_topdown_20210909-083420 mp4

ThomasSchellenbergNextCentury:
@deanwetherby @kdrumm The walls in the environment are always the same thickness. The inconsistency in my above examples is because the top-down-plotter's buffer cuts off part of the walls in rectangular rooms, so make sure you're testing with rectangular rooms as well.

deanwetherby:
@ThomasSchellenbergNextCentury The wall buffer value should account for that regardless of the room dimensions. I tested on some rectangular rooms just in case to confirm that is indeed true. It is.

ThomasSchellenbergNextCentury (commented):
Looking good.

| Int Obj Perm | Reor |
| --- | --- |
| dean_v2_intobjperm | dean_v2_reor |

I guess my only quibble now is that the wall is taking up a lot of real estate, but you can also argue it's the real size of the walls, and at this point it's really just me nitpicking anyway :)

So, approved!

ThomasSchellenbergNextCentury (approved):

deanwetherby:
I have the same issue but just tried to make the plotter a reflection of the scene. If we want thinner walls, then we should change then in the scenes and then the plotter will be good. Thanks for the additional examination.

PR #404, opened 2021-09-14
Enum and env var fixes

rartiss55:
Removed environment variables for metadata tier and references to an outdated debug level one, so now we just have an environment variable for the config file (MCS-839). 

Also made enums for metadata tier values (MCS-476). Still questioning the value add at this point for the metadata tier enums (its still a string value anyway in the config file itself + represented as so in the logging and get_metadata_level() controller function), but let me know what you think.


deanwetherby (approved):
I like this change

kdrumm (approved):

ThomasSchellenbergNextCentury (approved):
I like this too!

brianpippin (approved):

PR #408, opened 2021-09-16
Exit the developer run script immediately if no scene files were found.

ThomasSchellenbergNextCentury:
Example incorrect usage:

```
python3 scripts/run_passive_scenes.py --mcs_unity_build_file ~/Development/darpa/ai2thor/unity/MCS-AI2-THOR-Unity-App-v0.4.4.x86_64 docs/source/scenes/agents_preference_expected.json 
Loaded provided logging config dictionary
Found 0 files matching docs/source/scenes/agents_preference_expected.json*_debug.json
No matching files found... trying non-debug files
Found 0 files matching docs/source/scenes/agents_preference_expected.json*.json
No matching files found... Exiting
```

Example correct usage:
```
python3 scripts/run_passive_scenes.py --mcs_unity_build_file ~/Development/darpa/ai2thor/unity/MCS-AI2-THOR-Unity-App-v0.4.4.x86_64 docs/source/scenes/agents_preference_
Loaded provided logging config dictionary
Found 0 files matching docs/source/scenes/agents_preference_*_debug.json
No matching files found... trying non-debug files
Found 2 files matching docs/source/scenes/agents_preference_*.json
========================================
```

(etc.)

### DIFF
--- a/scripts/runner_script.py
+++ b/scripts/runner_script.py
@@ -1,5 +1,3 @@
-from __future__ import annotations
-
 import argparse
 import glob
 import logging
@@ -22,7 +20,7 @@ class AbstractRunnerScript():
         self,
         name: str,
         action_callback: Callable[
-            [Dict, mcs.StepMetadata, AbstractRunnerScript],
+            [Dict, mcs.StepMetadata, 'AbstractRunnerScript'],
             Tuple[str, Dict]
         ]
     ):
@@ -179,7 +177,7 @@ class AbstractRunnerScript():
         controller: mcs.Controller,
         filename: str,
         action_callback: Callable[
-            [Dict, mcs.StepMetadata, AbstractRunnerScript],
+            [Dict, mcs.StepMetadata, 'AbstractRunnerScript'],
             Tuple[str, Dict]
         ],
         last_step: int,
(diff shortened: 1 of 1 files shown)


rartiss55 (approved):

brianpippin (approved):
Code looks good, my only question and it might not apply to this is I know Dean wanted to move away print statements to a logger, but I also don't feel like initializing a logger for this made much sense as well, just trying to make sure on my own I understand when we want to use one or the other.

ThomasSchellenbergNextCentury:
@brianpippin 

> when we want to use one or the other

This is a good question, and Rachel and I were actually talking about it on Teams yesterday too. My opinion is that you should use prints for critical error messages that you want to be sure are always seen (since you can disable log messages), and loggers for everything else. This seemed like the former, which is why I went with prints.

deanwetherby:
Personal preferences aside
https://docs.quantifiedcode.com/python-anti-patterns/readability/putting_type_information_in_a_variable_name.html

ThomasSchellenbergNextCentury:
Not all anti-patterns are meant to be applied universally, and I disagree with that particular one in some cases. I've worked on software in which we've defined dicts, lists, numpy arrays, and pandas dataframes, and pass them around to lots of different functions. It can be very confusing figuring out which variables are which type without some sort of reference information, like the variable name. (Also, that page seems to recommend using single-letter variable names, which I see as an anti-pattern in itself, outside from for-loop index variables.)

Now, that being said, we've started doing a much better job using type hints throughout our code, which can provide the needed reference information, as long as we remain diligent with the type hinting. I've updated this script to add more type hints and rename `filename_list` as you requested.

PR #413, opened 2021-09-27
MCS-318 colormap

### LINKED ISSUE #240: Add Color Map to Step Metadata for Object (Instance Segmentation) Mask Colors
From: Ben, Falk https://machinecommonsense.slack.com/archives/CTR31V12M/p1594750155073000

Priority: Backlog

Components: Python API

Blockers: None

Steps:
- Remove “color” from ObjectMetadata
- Add “color_map” to StepMetadata
  - A Dict with object ID string keys and color pixel array values
  - Use scene_event.metadata['colors'] but reformat it
  - Should only be used in Oracle Metadata Mode
- Update the unit tests

TA2 internal JIRA ticket: https://nextcentury.atlassian.net/browse/MCS-318

deanwetherby:
Added dictionary ```segment_map``` to ```StepMetadata``` where key is the index into object_list. The ```rgb``` dictionary (as value) is the segmentation color for that object.

example segment map:
```
{0: {'r': 218, 'g': 65, 'b': 21},
1: {'r': 239, 'g': 4, 'b': 237},
2: {'r': 126, 'g': 205, 'b': 46},
3: {'r': 36, 'g': 139, 'b': 166},
4: {'r': 37, 'g': 216, 'b': 187}}
```

Removed misleading color attribute from ```ObjectMetadata```.

Ran into a few problems implementing this change:
1. Stringifier failed with integer dictionary keys. Will be making a ticket to address that.
2. Fixing ```test.msgpack``` for class modifications is more cumbersome than one would think.
3. Had to add ```strict_map_key=False``` to unpackb for deserialization as msgpack also does not like integer keys.

Closes #240


machine_common_sense/controller_output_handler.py
```diff
@@ -246,25 +239,25 @@ class ControllerOutputHandler():
...
     def __init__(self, config: ConfigManager):
-        # do nothing
         self._config = config
 
     def set_scene_config(self, scene_config):
         self._scene_config = scene_config
 
     def handle_output(self, raw_output, goal, step_number, habituation_trial):
-        self._scene_event = SceneEvent(
+        scene_event = SceneEvent(
```
deanwetherby, line 248: There's no need for the scene event to be a class attribute.

machine_common_sense/controller_output_handler.py
```diff
@@ -246,25 +239,25 @@ class ControllerOutputHandler():
...
             self._config, self._scene_config, raw_output, step_number)
-        self._step_number = step_number
         unrestricted = self._get_step_metadata(
-            goal, habituation_trial, False)
+            scene_event, step_number, goal, habituation_trial, False)
```
deanwetherby, line 251: Remove self from step number as well

machine_common_sense/controller_output_handler.py
```diff
@@ -246,25 +239,25 @@ class ControllerOutputHandler():
...
         return (unrestricted, restricted)
 
-    # used to be wrap output
-    def _get_step_metadata(self, goal, habituation_trial,
-                           restricted=True) -> StepMetadata:
+    def _get_step_metadata(self, scene_event: SceneEvent,
+                           step_number: int,
+                           goal: GoalMetadata,
+                           habituation_trial: int,
+                           restricted: bool = True) -> StepMetadata:
```
deanwetherby, line 260: Added type hinting and passed in the scene event and step number as arguments rather than using self.

machine_common_sense/controller_output_handler.py
```diff
@@ -285,63 +278,69 @@ def _get_step_metadata(self, goal, habituation_trial,
...
             reward=Reward.calculate_reward(
-                goal, objects, agent, self._step_number,
-                self._scene_event.performer_reach),
+                goal, objects, agent, step_number,
+                scene_event.performer_reach),
             rotation=(
-                None if restrict_non_oracle else self._scene_event.rotation),
-            step_number=self._step_number,
+                None if restrict_non_oracle else scene_event.rotation),
+            segment_map=segment_map,
```
deanwetherby, line 332: Adding segment_map to step metadata

machine_common_sense/controller_output_handler.py
```diff
@@ -285,63 +278,69 @@ def _get_step_metadata(self, goal, habituation_trial,
...
-                                    self._scene_event.structural_object_list)
+                                    scene_event.structural_object_list)
         )
 
-        # This is here to retain the exact outputs as before
-        # If None is passed into the constructor, it is turned into
-        # {}, but restricted mode turns it to None.
         if (restrict_non_oracle):
             step_output.position = None
+            step_output.segment_map = None
```
deanwetherby, line 342: Removing segment_map when using non-oracle metadata

machine_common_sense/serializer.py
```diff
@@ -64,7 +64,7 @@ def _ext_pack(x):
                     x.camera_height, x.depth_map_list, x.goal, x.head_tilt,
                     x.image_list, x.object_list, x.object_mask_list, x.pose,
                     x.position, x.return_status, x.reward, x.rotation,
-                    x.step_number, x.structural_object_list
+                    x.segment_map, x.step_number, x.structural_object_list
```
deanwetherby, line 67: Adding segment_map to serializer

machine_common_sense/serializer.py
```diff
@@ -119,9 +119,10 @@ def _ext_unpack(code, data):
             action_list, camera_aspect_ratio, camera_clipping_planes, \
                 camera_field_of_view, camera_height, depth_map_list, goal, \
                 head_tilt, image_list, object_list, object_mask_list, pose, \
-                position, return_status, reward, rotation, step_number, \
-                structural_object_list = \
-                msgpack.unpackb(data, ext_hook=SerializerMsgPack._ext_unpack)
+                position, return_status, reward, rotation, segment_map, \
+                step_number, structural_object_list = \
+                msgpack.unpackb(data, ext_hook=SerializerMsgPack._ext_unpack,
+                                strict_map_key=False)
```
deanwetherby, line 125: Had to add ```strict_map_key=False``` in order to unpack the segment_map with integer dictionary keys.

machine_common_sense/serializer.py
```diff
@@ -283,40 +283,48 @@ def convert_object_list(raw_list):
...
+                object_raw['distance_in_world'],
+                object_raw['held'],
+                object_raw['mass'],
+                object_raw['material_list'],
+                object_raw['position'],
+                object_raw['rotation'],
+                object_raw['shape'],
+                object_raw['state_list'],
+                object_raw['texture_color_list'],
+                object_raw['visible'])
```
deanwetherby, line 300: formatting changes to make it easier to read

machine_common_sense/serializer.py
```diff
@@ -283,40 +283,48 @@ def convert_object_list(raw_list):
...
+        ]
 
-        image_list = []
-        for img_raw in input_json['image_list']:  # PIL
-            image_list.append(Image.fromarray(np.array(img_raw,
-                                                       dtype='uint8')))
+        image_list = [
+            Image.fromarray(np.array(img_raw, dtype='uint8'))
+            for img_raw in input_json['image_list']
+        ]
```
deanwetherby, line 327: using list comprehension to build image list rather than for loop

rartiss55:
A few things...
1.) Are we certain that integer keys is preferable over using the uuid?
2.) Semi related, but should we also leave the "color" as part of the ObjectMetadata in addition to this map? To alleviate confusion, would renaming it "segmentation_color" or something be sufficient?
3.) If we're sticking with integer keys, for the serializer, would just doing `str(key)` be okay, or do we need something more complex?

ThomasSchellenbergNextCentury (approved):
Thanks for all the helpful PR comments. Looks like a lot of great changes here. Though I am interested in a followup to Rachel's questions (posted above).

tests/test_controller_output_handler.py
```diff
@@ -1128,6 +1097,28 @@ def test_retrieve_object_list_with_config_metadata_none(self):
...
+        coh = ControllerOutputHandler(self._config)
+        coh.set_scene_config(SceneConfiguration((mock_scene_event_data)))
+        (res, actual) = coh.handle_output(mock_event, GoalMetadata(), 0, 1)
+        print(actual.segment_map)
+        # scene_event = SceneEvent(
+        #     self._config,
+        #     SceneConfiguration(),
+        #     mock_event,
+        #     0)
+
```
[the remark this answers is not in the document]
deanwetherby: I meant to finish that test and hopefully add a few more

integration_tests/additional_integration_tests.py
```diff
@@ -87,26 +87,6 @@ def run_depth_and_segmentation_test(controller, metadata_tier):
...
-        if (
-            step_metadata_0.object_list[1].color !=
-            step_metadata_1.object_list[1].color
-        ):
-            return (
-                False,
-                f'Step 1 {metadata_tier} failed: object_list[1].color '
-                f'{step_metadata_0.object_list[1].color} != '
-                f'{step_metadata_1.object_list[1].color}'
-            )
```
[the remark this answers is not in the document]
deanwetherby: It would

deanwetherby:
> A few things... 1.) Are we certain that integer keys is preferable over using the uuid?
I assume so since that was the request. Although the GH issue description stated using uuid, that was an interpretation of the work that needed to be done. I imagine that TA1 will use the segment map keys (integer indices) to associate the segmentation color value with rather than using the uuid to do a more expensive list filter or lookup. That approach makes sense to me.

> 2.) Semi related, but should we also leave the "color" as part of the ObjectMetadata in addition to this map? To alleviate confusion, would renaming it "segmentation_color" or something be sufficient? 
Since TA1 requested this change, I'd rather not duplicate this information for maintainability and metadata filtering reasons.

> 3.) If we're sticking with integer keys, for the serializer, would just doing `str(key)` be okay, or do we need something more complex?
Since the serializer reads directly in to and out of python classes, I'd prefer keeping the integers and addressing our internal assumptions of the dictionary key always needing to be a string.

rartiss55:
I compared the outputs on dev and this branch, and it looks like we're now missing the structural object segmentation colors with this approach. Perhaps leaving the object metadata the way it was and renaming the "color" field to "segmentation_color/id" for clarity would make the most sense.

deanwetherby:
> I compared the outputs on dev and this branch, and it looks like we're now missing the structural object segmentation colors with this approach. Perhaps leaving the object metadata the way it was and renaming the "color" field to "segment_color/id" for clarity would make the most sense.

That is a good point @rartiss55. This PR may be null and void. As you mentioned, a less confusingly named attribute in ObjectMetadata may be the better approach here.

bzinberg:
Hi folks - was pointed here by https://github.com/NextCenturyCorporation/MCS/issues/240#issuecomment-929272818.  Thanks for this work!

My first question is, what do the colors in the color map mean?  I.e., same question as in second paragraph of https://github.com/NextCenturyCorporation/MCS/issues/240#issuecomment-920347128.

rartiss55:
@bzinberg Due to the earlier comments about some segmentation color information being missing using this approach, we will likely go with the alternative you mentioned with simply renaming the color field. See PR here: https://github.com/NextCenturyCorporation/MCS/pull/415

I'll go ahead and link the issue as well. Sorry for any confusion!

deanwetherby:
> My first question is, what do the colors in the color map mean?

@bzinberg These are the rgb values we get back from Unity representing instance segmentation regions in the segmentation image.

deanwetherby:
Closing in favor of #415

bzinberg:
Agreed, https://github.com/NextCenturyCorporation/MCS/pull/415 looks like the right fix for our situation.  Here are some more comments in case useful:

From the above I'm gathering that the upstream AI2Thor API for computing segmentation masks requires the user to think about colors (in particular, to compare the pixels of an image to the value of the "segment color").  I think, at least in hindsight, that probably wasn't the best choice: if a segmentation image gets passed around and at any point is converted lossily, e.g. to a GIF, then the pipeline (or the downstream viewer of a dashboard, etc.) breaks.  Just as a point of comparison, PyBullet's segmentation maps are integer arrays, of shape `(height_px, width_px)` where each entry is PyBullet's integer ID for the object, consistent with other methods for manipulating the object.

AI2Thor does seem to maintain a [mapping](https://github.com/allenai/ai2thor/blob/3ce9fe15a850b0a67e6ccd8fa92f317589caec83/ai2thor/server.py#L213-L219) from "name" (whatever that is) to "color"; though I don't know whether they expose these "names" in the public API.

PR #414, opened 2021-09-28
MCS-907 Fix Stringifier

kdrumm:
Requires MCS-318-colormap to be merged before this is merged into development.

Stringifier is used when we create debug output files as well as anyone in code tries to print step_metadata.  We now handle integer keys by quoting them such that we create valid JSON.  There is an argument that we may not want to quote the keys in the output, but then we need another mechanism to write the debug output.  There is commented out code that will handle this without quoting the keys, but it is commented out.  We should remove that during this PR.

### DIFF
--- a/tests/test_step_metadata.py
+++ b/tests/test_step_metadata.py
@@ -1,3 +1,4 @@
+import copy
 import textwrap
 import unittest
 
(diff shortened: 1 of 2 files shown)


tests/test_step_metadata.py
```diff
@@ -137,6 +178,19 @@ def test_str(self):
...
+    def test_str_int_segment_map(self):
+        metadata = copy.deepcopy(self.step_metadata)
+        metadata.segment_map = {
+            0: {
+                'r': 218,
+                'g': 65,
+                'b': 21
+            }
+        }
+        # test = str(metadata)
```
rartiss55, line 190: Can this comment be removed as well?

kdrumm: whoops!

deanwetherby (approved):

rartiss55 (approved):
That makes sense about the debug json files/leaving the quotes in is best.

kdrumm:
Closing this because it was based off MCS-318-colormap.  Made a new branch and PR with the changes.

PR #415, opened 2021-09-28
MCS-318 Rename color field

deanwetherby:
Simply renaming 'color' to 'segment_id' in ObjectMetadata. Contrasts with other branch/PR which moves segmentation color to higher level StepMetadata.

### DIFF
--- a/machine_common_sense/object_metadata.py
+++ b/machine_common_sense/object_metadata.py
@@ -40,7 +40,7 @@ class ObjectMetadata(object):
     rotation : dict
         This object's rotation angles around the "x", "y", and "z" axes
         in degrees.
-    segment_id : dict
+    segment_color : dict
         The "r", "g", and "b" pixel values of this object in images from the
         StepMetadata's "object_mask_list".
     shape : string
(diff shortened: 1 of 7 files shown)


rartiss55:
The run_depth_and_segmentation_test needs to get updated with the new name, otherwise LGTM.

deanwetherby:
I'm not sure `segment_id` is the correct attribute name for instance segmentation color. It's definitely less confusing than color though.

rartiss55:
> I'm not sure `segment_id` is the correct attribute name for instance segmentation color. It's definitely less confusing than color though.

Maybe segment_color or segmentation_color?

kdrumm (approved):

rartiss55 (approved):

machine_common_sense/object_metadata.py
```diff
@@ -43,6 +40,9 @@ class ObjectMetadata(object):
     rotation : dict
         This object's rotation angles around the "x", "y", and "z" axes
         in degrees.
+    segment_color : dict
+        The "r", "g", and "b" pixel values of this object in images from the
+        StepMetadata's "object_mask_list".
```
bzinberg, line 45: This doc is super helpful and clarifying.  Thanks!

bzinberg (commented):
Awesome, thanks folks

PR #417, opened 2021-09-29
Mcs 907 fix stringifier dev

kdrumm:
New branch for stringifier fix that is based off development instead of a feature branch since we want to keep this and not the other feature branch.  It should be identical except I needed to change the unit test string to make it pass.  For some reason the field order changed.


deanwetherby (approved):
Strange that the ordering changed. Segment_map shouldn't be in step_metadata yet though. I'll have to re-examine when one of the color map/segment_id PRs are merged.

PR #418, opened 2021-09-29
Mcs 909 write target visibility

cedorman:
For one of the scorecard values, we need to know whether or not the target is visible.  Add this to the step metadata output.  This means a change to the step metadata object, _init__, and __iter__, and associated unit test.   

The determination of whether or not the target is visible is done in history_writer.  It's not clear that this is the right place for the logic, but there are other manipulations for of the step metadata and it was the easiest place to put it.

### DIFF
--- a/machine_common_sense/history_writer.py
+++ b/machine_common_sense/history_writer.py
@@ -32,7 +32,7 @@ class HistoryEventHandler(AbstractControllerSubscriber):
             ] = payload.config.get_evaluation_name()
             hist_info[
                 payload.config.CONFIG_METADATA_TIER
-            ] = payload.config.get_metadata_tier()
+            ] = payload.config.get_metadata_tier().value
             hist_info[
                 payload.config.CONFIG_TEAM
             ] = payload.config.get_team()
--- a/machine_common_sense/history_writer.py
+++ b/machine_common_sense/history_writer.py
@@ -63,8 +63,8 @@ class HistoryEventHandler(AbstractControllerSubscriber):
 
     def on_end_scene(self, payload):
         if (
-                self.__history_writer is not None and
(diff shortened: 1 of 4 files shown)


deanwetherby:
How about moving the visibility flag to SceneHistory? It's still ingested and we wouldn't have the serialization issues.

machine_common_sense/history_writer.py
```diff
@@ -162,13 +162,31 @@ def add_step(self, step_obj: Dict):
         """Add a new step to the array of history steps"""
         current_time = perf_counter() * 1000
         if step_obj is not None:
-            step_obj.delta_time_millis = current_time - \
-                self.last_step_time_millis
+            step_obj.delta_time_millis = \
+                current_time - self.last_step_time_millis
             self.last_step_time_millis = current_time
+            if step_obj.output:
+                step_obj.output.target_visible = \
```
brianpippin, line 169: @deanwetherby  - do you mean just making this step_obj.target_visible here so we just add a value to the dictionary here and do nothing anywhere else?  I think that would give us the output we wanted and not have to worry about the metadata objects, right?

deanwetherby: Yes. You have the gist of it

deanwetherby: I think `step_obj` is a SceneHistory class though. We may have to add the attribute to that class but we won't run into serialization issues or confusion over what the field means in the metadata classes.

deanwetherby: That particular method argument has step_obj mistyped as Dict.

deanwetherby (approved):
Feels weird approving my own branch merge into this PR but it is what it is.

brianpippin (approved):

PR #420, opened 2021-09-30
Moving scripts folder inside package folder

deanwetherby:


brianpippin (approved):
LGTM

deanwetherby:
Verified that the scripts exposed through entry_points work as intended in a new/fresh virtual environment. Moved every script that relied on the runner into the package folder. I'm not sure we need them all.

Just noticed that the ini config files are not installed. They probably have to be handled with package_data if we need them.

Left sphinx docs scripts and other items in the root scripts folder.

rartiss55:
Sounds good/run_in_human_input_mode works!

Dunno if we want to split the scripts up later, but I don't think its a big deal right now. I guess the only other thing is if we directly reference the scripts/ folder in any docs, we want to update those paths.

rartiss55 (approved):

brianpippin (approved):

deanwetherby:
Most importantly, does cache_addressables work?

rartiss55:
Re cache_addressables, seems like it (I deleted what was in my Application.persistentDataPath and ran that script).

ThomasSchellenbergNextCentury (approved):
Agreed that there are probably some scripts that don't need to be entry points, but we can worry about reorganizing them later (when we have time).

deanwetherby:
Thanks for the prompt feedback. I'll merge as soon as GH Actions finishes with this last push.

PR #435, opened 2021-11-26
Mcs 1008 multiple end scene

deanwetherby:


machine_common_sense/controller.py
```diff
@@ -136,7 +136,8 @@ def __init__(self, unity_app_file_path: str, config: ConfigManager):
 
         self._subscribers = []
 
-        self._end_scene_not_registered = True
+        self._failure_handler_registered = False
```
deanwetherby, line 139: Renamed to remove the double negative cognitive load; `if not not_registered`

kdrumm (approved):
LGTM

brianpippin (approved):

PR #436, opened 2021-11-30
Added trace logging level.  Added call to init_logging such that it s…

kdrumm:
To test backup init_logging:

- Run your favorite entry point  (I.E. 'run_human_input.py') and verify default logging
- Remove init_logging from your favorite entry point (be sure to revert this when you are done)
- Run your favorite entry point and verify no logging (besides print statements. Note: run_human_input has a lot of print statements which will circumvent logging configurations.)
- Create a 'log.config.user.py' file in your working directory and add a config (copy and modify from 'Logging_config.py').  I recommend adding a prefix string to the format like this: '"format": "TEST%(message)s"' to make it obvious when this file is used.
- Run your favorite entry point.  Verify your logging works. 
- Remove or rename 'log.config.user.py' so it doesn't get used continuously

To test trace:
- add a logger.trace("message") somewhere
- Run and verify no message (assuming default logging config)
- add a 'log.config.user.py' file that supports trace logging
- Run and verify your log message now appears
- Remove or rename 'log.config.user.py' so it doesn't get used continuously

Here is an override 'log.config.user.py' file that I used:
`
{
    "version": 1,
    "root": {
        "level": "TRACE",
        "handlers": ["console"],
        "propagate": False
    },
    "loggers": {
        "machine_common_sense": {
            "level": "TRACE",
            "handlers": ["console"],
            "propagate": False
        }
    },
    "handlers": {
        "console": {
            "class": "logging.StreamHandler",
            "formatter": "brief",
            "level": "TRACE",
            "stream": "ext://sys.stdout"
        }
    },
    "formatters": {
        "brief": {
            "format": "TEST%(message)s"
        }
    }
}
`

### DIFF
--- a/machine_common_sense/logging_config.py
+++ b/machine_common_sense/logging_config.py
@@ -217,4 +217,4 @@ class LoggingConfig():
 
 # Call init logging early with no config to ensure it is called.  Logging
 # config can always be overwritten any time MCS is used.
-LoggingConfig.init_logging(LoggingConfig.get_no_logging_config())
+LoggingConfig.init_logging(LoggingConfig.get_errors_only_console_config())
--- a/machine_common_sense/logging_config.py
+++ b/machine_common_sense/logging_config.py
@@ -43,6 +43,8 @@ import os
 
 logger = logging.getLogger(__name__)
 
+TRACE = 5
+
 
 class LoggingConfig():
 
@@ -84,6 +86,7 @@ class LoggingConfig():
         If user file doesn't exist, then there is a base config that
         should be read.
         """
+        LoggingConfig._add_trace()
         init_message = ""
         if (os.path.exists(log_config_file)):
             with open(log_config_file, "r") as data:
@@ -99,6 +102,15 @@ class LoggingConfig():
             os.mkdir("logs")
         logging.config.dictConfig(log_config)
         logger.info(init_message)
+
+    @staticmethod
+    def _add_trace():
+        logging.addLevelName(TRACE, "TRACE")
+
+        def trace(self, message, *args, **kws):
+            if self.isEnabledFor(TRACE):
+                self._log(TRACE, message, args, **kws)
+        logging.Logger.trace = trace
 
     @staticmethod
     def get_default_console_config():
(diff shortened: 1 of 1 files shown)


machine_common_sense/logging_config.py
```diff
@@ -193,3 +213,8 @@ def get_dev_logging_config():
                 }
             }
         }
+
+
+# Call init logging early with no config to ensure it is called.  Logging
+# config can always be overwritten any time MCS is used.
+LoggingConfig.init_logging(LoggingConfig.get_no_logging_config())
```
kdrumm, line 220: The only weirdness with this is if init_logging() is called by the performer and we use the 'log.config.user.py' file to override their logging config, the 'Loaded logging config from ...' message will be displayed twice.  I think this is a minor tradeoff for the ability to override.

kdrumm: This call should occur at the first instance where LoggingConfig is imported (Right?).  Therefore it should be called whenever MCS is used since we import in the __init__.py.  It should also happen BEFORE performers or scripts can initialize logging so they can override the default, but we can override that with a file.  @deanwetherby please verify I am understanding python correctly.

deanwetherby: That is correct.

deanwetherby (approved):

brianpippin (approved):

ThomasSchellenbergNextCentury (approved):
Maybe this isn't the right place for this discussion, but it seems like MCS swallowing errors by default (not showing any error messages unless error logging is enabled) has caused a lot of headache, both while debugging eval submissions, and previous development I've done. I guess we can try to remember to enable error logging ourselves (though we probably can't rely on TA1 being as rigorous), but it seems to me like we should just return the errors by default, instead of swallowing them (if not to console, then to an error log file). Thoughts?

kdrumm:
> Maybe this isn't the right place for this discussion, but it seems like MCS swallowing errors by default (not showing any error messages unless error logging is enabled) has caused a lot of headache, both while debugging eval submissions, and previous development I've done. I guess we can try to remember to enable error logging ourselves (though we probably can't rely on TA1 being as rigorous), but it seems to me like we should just return the errors by default, instead of swallowing them (if not to console, then to an error log file). Thoughts?

I think you are right and I can fix the default here.  My original goal in this PR was to not change the behavior, but I think your idea is an improvement.

kdrumm:
Before I merge, does anyone have any comments about the change to put the "fail-safe" logging level to errors on console?

PR #437, opened 2021-12-01
Adding support for Python 3.10

deanwetherby:
Package dependencies all install properly. Integration tests pass. Unit tests pass albeit with numpy deprecation warnings with our new numpy-aware serialization.

(mcs-p310) MCS $ python -m unittest
............................................................................................................/home/dwetherby/workspace/MCS/tests/test_history_writer.py:242: DeprecationWarning: np.int is a deprecated alias for the builtin int. To silence this warning, use int by itself. Doing this will not modify any behavior and is safe. When replacing np.int, you may wish to use e.g. np.int64 or np.int32 to specify the precision. If you wish to review your current use, check the release note link for additional information.
Deprecated in NumPy 1.20; for more details and guidance: NumPy 1.20.0 Release Notes — NumPy v1.23.dev0 Manual

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,6 +32,7 @@ setuptools.setup(
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: 3.9',
+        'Programming Language :: Python :: 3.10',
         'License :: OSI Approved :: Apache Software License',
         'Natural Language :: English',
         'Operating System :: OS Independent',
@@ -46,7 +47,7 @@ setuptools.setup(
         'Typing :: Typed'
     ],
     license='Apache-2',
-    python_requires=">3.6, <3.10",
+    python_requires=">3.6",
     packages=setuptools.find_packages(),
     package_data={'': ['*.ini'], },
     install_requires=[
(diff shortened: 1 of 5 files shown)


brianpippin (approved):

setup.py
```diff
@@ -46,7 +47,7 @@ def get_version():
         'Typing :: Typed'
     ],
     license='Apache-2',
-    python_requires=">3.6, <3.10",
+    python_requires=">3.6",
```
alexcraneunity, line 50: Can you explain why this was removed if we are adding support for 3.10? Was it so that it can pass tests, like we ran into with 3.9?

deanwetherby: The previous python_requires attribute would have prevented installation of the MCS package in python 3.10. So we were saying anything above 3.6 and below 3.10 was supported. Once I finished testing with 3.10 and showed it working, that "upper" restriction is being removed. Is that what you were asking about?

alexcraneunity: Oh yes, that makes sense. Thanks!

deanwetherby:
> Should we add 3.10 to the development-merge action?

Excellent idea

kdrumm (approved):
Looks good to me to let GH actions test on merge.

PR #438, opened 2021-12-09
Mcs 908 serializer maintainability

deanwetherby:
Removed test.msgpack file resource as it was very difficult to work with when making changes to the underlying classes. Wrote more extensive msgpack serializer unit tests. Dropped support for the JSON serializer as I don't think it is being used. If needed, we can accomplish serialization and deserialization more formally using the marshmallow package which we already have as an MCS dependency.


JuanRMarquez-Ortiz (approved):
I don't have enough context to really assess the intended result is being achieved. The syntax looks fine, though.

alexcraneunity (approved):
Syntax and tests look good.

kdrumm (approved):

PR #439, opened 2021-12-09
MCS-707 Trailing commas

deanwetherby:
Draft PR to discuss the merits of adding flake8-commas to enforce trailing commas.

### DIFF
--- a/tests/test_mcs.py
+++ b/tests/test_mcs.py
@@ -198,13 +198,13 @@ class TestMCS(unittest.TestCase):
                     "position": {
                         "x": 0,
                         "y": 0.5,
-                        "z": 3
+                        "z": 3,
                     },
                     "scale": {
                         "x": 0.25,
                         "y": 0.25,
-                        "z": 0.25
-                    }
+                        "z": 0.25,
+                    },
                 }],
                 "forces": [{
                     "stepBegin": 1,
@@ -212,10 +212,10 @@ class TestMCS(unittest.TestCase):
                     "vector": {
                         "x": 50,
                         "y": 0,
-                        "z": 0
-                    }
-                }]
-            }]
+                        "z": 0,
+                    },
+                }],
+            }],
         }
         self.assertEqual(actual, expected)
         self.assertEqual(status, None)
@@ -227,7 +227,7 @@ class TestMCS(unittest.TestCase):
         self.assertEqual(
             status,
             "The given file 'tests/test_scene_invalid.json' does not " +
-            "contain valid JSON."
+            "contain valid JSON.",
         )
 
     def test_load_scene_file_json_is_missing(self):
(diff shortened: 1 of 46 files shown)


kdrumm:
I'll never seen this before a couple times in this project.  What is the reasoning for trailing commas?

deanwetherby:
I did not look at any auto-tools for this as I wanted to get a feel for what the linter was going to complain about. I'm not 100% convinced on this change right now.

ThomasSchellenbergNextCentury:
@kdrumm 

> I'll never seen this before a couple times in this project. What is the reasoning for trailing commas?

Here's an article: https://medium.com/@nikgraf/why-you-should-enforce-dangling-commas-for-multiline-statements-d034c98e36f8

@deanwetherby 

> I'm not 100% convinced on this change right now.

That's cool. I just brought it up in case the existence of an auto-tool affected our decision on whether or not to use the rule.

tests/test_mcs.py
```diff
@@ -198,24 +198,24 @@ def test_load_scene_file_json(self):
...
                         "x": 50,
                         "y": 0,
-                        "z": 0
-                    }
-                }]
-            }]
+                        "z": 0,
+                    },
+                }],
+            }],
```
deanwetherby, line 218: This is one example of the dangling commas checker being fairly annoying. I suspect rather than writing code with the extra commas, I'll end up writing one-liners that are more difficult to read/maintain. That is unless the addition of those commas are generated with a tool like the one mentioned earlier.

JuanRMarquez-Ortiz:
I've always felt that the comma implies a subsequent element of whatever series the comma is delimiting, so having a comma when there's no more elements just looks inelegant and like bad grammar, even . I get the point as mentioned in the article Thomas linked, but to me they feel like nitpicking. None of the concerns presented in the article add such a cognitive load that I would enforce it. *shrugs*

rartiss55:
So I could be convinced if we were going to switch to a linter like Black (its a very opinionated linter BUT also seems pretty easy to use). In general though, I'm not the biggest fan of trailing commas/makes me think I deleted the last item in the list or something (plus the weird edge case listed above is hard to make sense of at first glance).

deanwetherby:
Closing PR without merging. Even with an auto-tool, trailing commas makes the code more difficult to read as the change is more invasive than anticipated. One could argue the cognitive load on maintaining/reading trailing commas is greater than those of future PRs without it.

ThomasSchellenbergNextCentury:
FYI Saying something objective like "trailing commas makes the code more difficult to read" about something that's subjective like linter rules might rub some people the wrong way.

JuanRMarquez-Ortiz:
I think we can see "difficult" -though subjective- as an analogue for cognitive load, which can be more objective. 

If it's a new item to pay attention all the time (while programming) for a benefit of not having to pay attention to something considerably less often (in PRs), I think we can say that it's a net increase in the cognitive load on the developers, even if for some it is easier or harder to handle than for others.

PR #440, opened 2021-12-09
MCS-591 pep8 naming

deanwetherby:
Enforces python naming conventions.

### DIFF
--- a/machine_common_sense/config_manager.py
+++ b/machine_common_sense/config_manager.py
@@ -265,14 +265,14 @@ class GoalSchema(Schema):
     answer = fields.Dict()  # UI property
     category = fields.Str()
     description = fields.Str()
-    domainsInfo = fields.Dict()  # UI property
+    domains_info = fields.Dict(data_key='domainsInfo')  # UI property
     habituation_total = fields.Int()
     info_list = fields.List(fields.Str())
     last_preview_phase_step = fields.Int()
     last_step = fields.Int()
     metadata = fields.Dict()
-    objectsInfo = fields.Dict()  # UI property
-    sceneInfo = fields.Dict()  # UI property
+    objects_info = fields.Dict(data_key='objectsInfo')  # UI property
+    scene_info = fields.Dict(data_key='sceneInfo')  # UI property
     skip_preview_phase = fields.Bool()
     task_list = fields.List(fields.Str())
     type_list = fields.List(fields.Str())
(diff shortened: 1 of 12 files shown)


deanwetherby:
Be sure to review the config_manager module. There are marshmallow schema classes that I am mapping from the scene json files into pythonic variables using the `data_key` property. This way we can have mixed case `stepBegin` mapped to `step_begin` and the pep8-naming checks all pass.

machine_common_sense/config_manager.py
```diff
@@ -283,7 +283,7 @@ def make_goal(self, data, **kwargs):
 
 
 class ChangeMaterialConfigSchema(Schema):
-    stepBegin = fields.Int()
+    step_begin = fields.Int(data_key='stepBegin')
```
ThomasSchellenbergNextCentury, line 286: I'm assuming the `data_key` property means that they'll be sent to Unity as camelCase?

deanwetherby: `data_key` is for deserialization from scene json files. Is there a concern going the other way? I ran a run_pass on the playroom without issue.

machine_common_sense/config_manager.py
```diff
@@ -571,11 +591,11 @@ class OpenCloseConfig:
...
 class PhysicsConfig:
     enable: bool = False
-    angularDrag: float = None
+    angular_drag: float = None
     bounciness: float = None
     drag: float = None
-    dynamicFriction: float = None
-    staticFriction: float = None
+    dynamic_friction: float = None
+    static_friction: float = None
```
ThomasSchellenbergNextCentury, line 598: This will affect the ILE scene generator since one of its files (global_settings_component.py) imports and uses the PhysicsConfig, so we'll need to update that code too.

ThomasSchellenbergNextCentury (commented):
I like enforcing Python naming conventions, so as long as it works with Unity (which it sounds like it does) and the other code, it's fine with me.

machine_common_sense/config_manager.py
ThomasSchellenbergNextCentury: The action/step payloads we send to Unity need to have camelCase variables to match the variables in the corresponding C# classes, like this: https://github.com/NextCenturyCorporation/ai2thor/blob/development/unity/Assets/Scripts/MCSMain.cs#L1933-L1936

However I'm assuming if the playroom loaded correctly for you then everything's working fine! :smile:

deanwetherby: I thought I would need to specify the serialization field as well but it looks like marshmallow is handling that for me. Thanks marshmallow.

https://marshmallow.readthedocs.io/en/stable/quickstart.html#specifying-serialization-deserialization-keys

JuanRMarquez-Ortiz (approved):

PR #441, opened 2021-12-14
MCS-342 validate types

deanwetherby:


machine_common_sense/action.py
```diff
@@ -646,16 +649,17 @@ def __repr__(self):
...
         return self._key
 
-    @ property
+    @property
     def desc(self):
         return self._desc
 
     @staticmethod
-    def input_to_action_and_params(input_str):
+    @typeguard.typechecked
```
deanwetherby, line 661: typeguard decorator has to go inside the static method decorator

machine_common_sense/__init__.py
```diff
@@ -3,6 +3,9 @@
 import logging.config
 import signal
 from contextlib import contextmanager
+from typing import Dict, Union
+
+import typeguard
```
deanwetherby, line 8: We can do either way below although I prefer the namespaced option better.

```
import typeguard

@typeguard.typechecked
def func():
  ...
```

or

```
from typeguard import typechecked

@typechecked
def func():
  ...
```

machine_common_sense/controller.py
```diff
@@ -583,8 +589,8 @@ def validate_and_convert_params(self, action, **kwargs):
             receptacleObjectImageCoords=receptacle_vector
         )
 
-    # Override
-    def step(self, action: str, **kwargs) -> StepMetadata:
+    @typeguard.typechecked
+    def step(self, action: str, **kwargs) -> Union[StepMetadata, None]:
```
deanwetherby, line 593: With type checking, I had to add the Union to also include None.

machine_common_sense/controller.py
```diff
@@ -706,19 +712,21 @@ def mcs_action_to_ai2thor_action(self, action):
     def retrieve_action_list_at_step(self, goal, step_number):
         return goal.retrieve_action_list_at_step(step_number)
 
-    def retrieve_object_states(self, object_id):
+    @typeguard.typechecked
+    def retrieve_object_states(self, object_id: str) -> List:
         """Return the state list at the current step for the object with the
         given ID from the scene configuration data, if any."""
-        self._scene_config.retrieve_object_states(
+        return self._scene_config.retrieve_object_states(
```
deanwetherby, line 719: There was no return on this function from previous release(s) even though it was listed in the MCS API.

machine_common_sense/reward.py
```diff
@@ -210,12 +211,13 @@ def _calculate_default_reward(
         return GOAL_NOT_ACHIEVED
 
     @staticmethod
+    @typeguard.typechecked
     def calculate_reward(
-            goal: GoalMetadata,
-            objects: Dict,
+            goal: Union[GoalMetadata, None],
+            objects: List[Dict],
```
deanwetherby, line 217: Type checking exposed that objects are always a List even though we type hinted they were a dictionary.

machine_common_sense/reward.py
```diff
@@ -210,12 +211,13 @@ def _calculate_default_reward(
...
             agent: Dict,
             number_steps: int,
-            reach: float) -> float:
+            reach: Union[float, None]) -> float:
```
deanwetherby, line 220: Again with the explicit None

machine_common_sense/reward.py
```diff
@@ -210,12 +211,13 @@ def _calculate_default_reward(
         return GOAL_NOT_ACHIEVED
 
     @staticmethod
+    @typeguard.typechecked
     def calculate_reward(
-            goal: GoalMetadata,
-            objects: Dict,
+            goal: Union[GoalMetadata, None],
```
deanwetherby, line 216: Again with the explicit None

machine_common_sense/validation.py
```diff
@@ -61,7 +65,7 @@ def is_number(value, label=None):
         try:
             float(value)
             return True
-        except ValueError:
+        except (ValueError, TypeError):
```
deanwetherby, line 66: Tested None as value which through an unhandled TypeError so I added it.

requirements.txt
```diff
@@ -21,4 +21,5 @@ Shapely==1.8.0; python_version>="3.10"
 Sphinx==3.2.1
 sphinxcontrib-websupport==1.2.4
 sphinx-markdown-builder==0.5.4
+typeguard==2.13.2
```
deanwetherby, line 24: Adding latest version to requirements.txt

setup.py
```diff
@@ -60,7 +60,8 @@ def get_version():
         'ai2thor==2.5.0',
         'scikit-image>=0.17.1',
         'dataclasses==0.8; python_version<"3.7"',
-        'marshmallow>=3.5.2'
+        'marshmallow>=3.5.2',
+        'typeguard>=2.11.1'
```
deanwetherby, line 64: Added lowest package version that supported latest python versions of 3.9 and 3.10. (Trailing commas diff here isn't a big deal)

tests/test_reward.py
```diff
@@ -11,21 +11,21 @@ class TestReward(unittest.TestCase):
     def test_default_reward(self):
         goal = mcs.GoalMetadata()
         reward = mcs.Reward.calculate_reward(
-            goal, objects={}, agent={}, number_steps=1, reach=1.0)
+            goal, objects=[{}], agent={}, number_steps=1, reach=1.0)
```
deanwetherby, line 14: Had to update reward tests because the type checking was correctly identifying an error with the Dict not being a list.

tests/test_reward.py
```diff
@@ -213,9 +209,7 @@ def test_traversal_reward_large_object_inside(self):
         obj['objectBounds']['objectBoundsCorners'].append(
             {'x': 0.0, 'y': 1.0, 'z': 10.0})
         obj['position'] = {'x': 0.0, 'z': 0.0}
-        obj_list = []
-        obj_list.append(obj)
-
+        obj_list = [obj]
```
deanwetherby, line 212: small cleanup

tests/test_validation.py
```diff
@@ -39,12 +39,15 @@ def test_is_in_range(self):
...
-        self.assertEqual(mcs.Validation.is_number('asdf'), False)
+        self.assertTrue(mcs.Validation.is_number(3))
+        self.assertTrue(mcs.Validation.is_number(-2.70))
+        self.assertTrue(mcs.Validation.is_number('0'))
+        self.assertTrue(mcs.Validation.is_number('1'))
+        self.assertTrue(mcs.Validation.is_number('12.34'))
+        self.assertTrue(mcs.Validation.is_number('01'))
+        self.assertFalse(mcs.Validation.is_number(''))
+        self.assertFalse(mcs.Validation.is_number('asdf'))
+        self.assertFalse(mcs.Validation.is_number(None))
```
deanwetherby, line 50: Updated tests to use correct assert and added a few more.

machine_common_sense/controller_output_handler.py
```diff
@@ -148,7 +148,8 @@ def return_status(self):
                 ].name
         except KeyError:
             logger.error(
-                "Return status {self._raw_output.metadata['lastActionStatus']}"
+                f"Return status "
```
deanwetherby, line 151: Supposed to be an f-string. And then the 'f' pushed the line to 80 characters.

tests/test_controller_output_handler.py
```diff
@@ -10,6 +10,10 @@
 from machine_common_sense.controller_output_handler import (
     ControllerOutputHandler, SceneEvent)
 from machine_common_sense.goal_metadata import GoalMetadata
+from machine_common_sense.logging_config import LoggingConfig
+
+# ignore printing of errors for unit testing
+LoggingConfig.init_logging(LoggingConfig.get_no_logging_config())
```
deanwetherby, line 16: KeyError occurred in one of the tests but we don't need that printed out to the console so turning off logging (i.e. correctly swallow errors)

JuanRMarquez-Ortiz (approved):

machine_common_sense/controller.py
ThomasSchellenbergNextCentury: Couldn't this be `Optional[StepMetadata]`?

machine_common_sense/controller.py
```diff
@@ -731,7 +739,7 @@ def _remove_none(self, d):
                 del d[key]
         return d
 
-    def wrap_step(self, **kwargs):
+    def wrap_step(self, **kwargs) -> Dict:
```
ThomasSchellenbergNextCentury, line 742: Why no typeguard here? Is it because of the kwargs?

ThomasSchellenbergNextCentury:
This is really awesome, but the ticket (MCS-342) was originally written to implement this behavior in the scene generator. Should we make a separate ticket to add typeguard to the scene generator?

deanwetherby:
> This is really awesome, but the ticket (MCS-342) was originally written to implement this behavior in the scene generator. Should we make a separate ticket to add typeguard to the scene generator?

Once we understand an approach, it'll take very little work to integrate into the scene generator. I was more concerned at the time with prototyping as well as thinking about how to protect our Python API.

machine_common_sense/controller.py
deanwetherby: I only type checked the functions exposed in the public API for now. If we think internal type checking will help, we are free to sprinkle it in as necessary.

machine_common_sense/controller.py
deanwetherby: Looks like typeguard handles that case. Change incoming

brianpippin (approved):
This looks super useful for the API and I think will be useful to spread to other places.

machine_common_sense/controller.py
kdrumm: Should we add a unit test for this function to prevent this?  Should we make sure we have a couple unit tests for all API functions?  (we probably have a bunch)

kdrumm (approved):
Looks good.  I did make a comment about additional unit tests, but approving anyway because I'm not that concerned if we don't add them.

ThomasSchellenbergNextCentury (approved):

PR #443, opened 2021-12-14
MCS-951 Reduced function complexity of validate_and_convert_params

deanwetherby:
Refactored this function pretty much as-is although I do not agree with the validation approach here. Why resort to a default if we are given an invalid value? Raising an exception seems more appropriate to me in that case. Providing a default if no value is given does seem like a good idea especially if we are relying on having a value for that attribute.

### DIFF
--- a/machine_common_sense/controller.py
+++ b/machine_common_sense/controller.py
@@ -5,7 +5,7 @@ import json
 import logging
 import os
 import random
-from typing import Dict, Union
+from typing import Any, Dict, Union
 
 import ai2thor.controller
 import ai2thor.server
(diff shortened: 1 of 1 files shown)


deanwetherby:
MCS-677 is for raising exceptions. That'll be next.

ThomasSchellenbergNextCentury (approved):
Regarding using default values if given invalid input: I have no problem if we want to change the behavior here to raise an error if given invalid inputs. (I don't remember why we chose to do it this way originally, but our initial rationale is probably OBE by now anyway.)

kdrumm (approved):

PR #444, opened 2021-12-16
MCS-677 replace defaults with exceptions

deanwetherby:


machine_common_sense/__init__.py
```diff
@@ -25,7 +25,6 @@
 from .stringifier import Stringifier
 from .subscriber import add_subscribers
 from .unity_executable_provider import UnityExecutableProvider
-from .validation import Validation
```
deanwetherby, line 28: Validation module no longer needed

machine_common_sense/action.py
```diff
@@ -694,10 +692,13 @@ def input_to_action_and_params(input_str: str) -> Tuple:
         try:
             for param in input_split[1:]:
                 param_key, param_value = param.split('=')
-                if Validation.is_number(param_value.strip()):
+                params[param_key.strip()] = param_value.strip()
```
deanwetherby, line 695: For Action, just assume a string for the value.

machine_common_sense/action.py
```diff
@@ -694,10 +692,13 @@ def input_to_action_and_params(input_str: str) -> Tuple:
...
+
+                try:
                     params[param_key.strip()] = float(param_value.strip())
-                else:
-                    params[param_key.strip()] = param_value.strip()
+                except ValueError:
```
deanwetherby, line 699: Replace the action string with a float if it is float-able. Otherwise ignore the conversion ValueError exception.

machine_common_sense/controller.py
```diff
@@ -98,17 +97,17 @@ class Controller():
...
+    DEFAULT_OBJECT_MOVE_AMOUNT = 1.0
 
-    MAX_FORCE = 1
-    MIN_FORCE = 0
-    MAX_AMOUNT = 1
-    MIN_AMOUNT = 0
+    MAX_FORCE = 1.0
+    MIN_FORCE = 0.0
+    MAX_AMOUNT = 1.0
+    MIN_AMOUNT = 0.0
```
deanwetherby, line 110: Explicit about floats

tests/test_controller.py
```diff
@@ -643,31 +646,23 @@ def test_step_validate_parameters_open_close(self):
...
-                action='MCSOpenObject',
-                moveMagnitude=mcs.Controller.DEFAULT_AMOUNT,
+        self.assertRaises(
+            ValueError,
+            lambda: self.controller.step(
+                'OpenObject',
+                amount=1.5,
                 objectId='test_id_1',
-                receptacleObjectId='test_id_2'))
+                receptacleObjectID='test_id_2')
```
deanwetherby, line 655: Learned about using lambda in unittest assertRaises so it looks exactly like the function being called.

deanwetherby:
I do feel like the parameter conversion needs to happen in a module outside the controller but that'll be for another day.

brianpippin (approved):
Code looked good, all unit tests passed.

JuanRMarquez-Ortiz (approved):

ThomasSchellenbergNextCentury (approved):
Looks good. I was concerned about whether old agent scenes would still run correctly, but they work fine. I like the new unit test.

PR #446, opened 2021-12-19
MCS-1069 Drop support for python3.6

deanwetherby:


tests/test_history_writer.py
```diff
@@ -239,16 +239,16 @@ def test_write_history_file_with_numpy(self):
         writer = mcs.HistoryWriter(self.prefix_config_data)
 
         history_item = mcs.SceneHistory(
-            step=np.int(1),
+            step=np.int32(1),
```
deanwetherby, line 242: Deprecation warning with numpy for np.int. Either use python int or specify int32/int64.

kdrumm (approved):

PR #447, opened 2021-12-20
MCS-1064 verify end scene type exceptions

deanwetherby:


.github/workflows/ci.yaml
```diff
@@ -27,7 +27,9 @@ jobs:
       - name: Run unit tests
         run: python -m unittest
       - name: Run isort
-        uses: jamescurtin/isort-action@master
+        uses: isort/isort-action@master
+        with:
+            requirementsFiles: "requirements.txt requirements-test.txt"
```
deanwetherby, line 32: isort was failing in GH Actions only with this branch's latest push.

docs/source/containerize.rst
```diff
@@ -4,7 +4,7 @@ Containerization
 GPU Image
 ---------
 
-Please note that the GPU image requires an Nvidia GPU and Nvidia Docker to be installed.
+GPU images requires an Nvidia GPU and Nvidia Docker to be installed.
```
deanwetherby, line 7: "Please note that" doesn't add any value to the sentence. Although polite, it's not needed.

machine_common_sense/controller.py
```diff
@@ -244,8 +244,8 @@ def _set_config(self, config):
     @typeguard.typechecked
     def end_scene(
         self,
-        rating: Union[float, int, str] = None,
-        score: float = 1.0,
+        rating: Optional[float] = None,
+        score: Optional[float] = None,
```
deanwetherby, line 248: All of these arguments are "optionally required" depending on the task. Has an API smell that could use improving.

machine_common_sense/controller.py
```diff
@@ -255,20 +255,20 @@ def end_scene(
 
         Parameters
         ----------
-        rating : float or int or string, required
+        rating : float, optional
```
deanwetherby, line 258: Removed confusing string type.

machine_common_sense/controller.py
```diff
@@ -331,7 +331,7 @@ def end_scene(
         """
         if(not self._end_scene_called):
             payload = self._create_event_payload_kwargs()
-            payload['rating'] = str(rating)
+            payload['rating'] = rating
```
deanwetherby, line 334: why were we converting to a string?

tests/test_controller.py
```diff
@@ -5,7 +5,7 @@
 from types import SimpleNamespace
 from unittest.mock import ANY, MagicMock
 
-import numpy
+import numpy as np
```
deanwetherby, line 8: Numpy is nearly always imported as 'np'.

kdrumm (approved):

machine_common_sense/controller.py
brianpippin: It would almost seem like we would need an end scene that filtered to other end scenes that were test type specific that would then check for required returned parameters and return exceptions if they weren't provided?  Was that what you were thinking?

machine_common_sense/controller.py
brianpippin: I assume that is left over where we were expecting it to be Plausible or Implausible but that definitely made Ingest more tricky when doing Evaluation 4 scoring.

brianpippin (approved):

PR #448, opened 2021-12-21
MCS-936 action pruning

deanwetherby:
Removing lying, crawling, standing and throw object actions from the code base. There are no plans to test performers on anything that would require these actions. As a result, there's no need for the pose module either.

Now there are 4 integration tests that rely on these actions or pose states. What do we do with these integration tests?


kdrumm:
> Removing lying, crawling, standing and throw object actions from the code base. There are no plans to test performers on anything that would require these actions. As a result, there's no need for the pose module either.
> 
> Now there are 4 integration tests that rely on these actions or pose states. What do we do with these integration tests?

Which tests are they?  if they test something else important, we should probably keep and modify them.

deanwetherby:
> Which tests are they? if they test something else important, we should probably keep and modify them.

071.pickup_container
076.ramps_obstructed_60deg
077.ramps_obstructed_from_side
079.ramps_pose_change

kdrumm:
I think

> > Which tests are they? if they test something else important, we should probably keep and modify them.
> 
> 071.pickup_container 076.ramps_obstructed_60deg 077.ramps_obstructed_from_side 079.ramps_pose_change

I don't see why the ramp ones would need a pose.  I believe i used crawl in the container one because dropping the object added some randomness due to physics and crawling made it fall lower.  I can try to fix these.

kdrumm (approved):

PR #450, opened 2021-12-30
MCS-668 controller refactor

deanwetherby:
Continuous improvements to the controller by refactoring components out of that module where possible. Improved test coverage on previously untested but key functions. Found a few issues in that process but those have been addressed as part of this PR.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -164,13 +164,6 @@ save_debug_json
 
 Save AI2-THOR/Unity input, AI2-THOR/Unity output, and MCS StepMetadata output to JSON file on each step. Default: False
 
-seed
-^^^^
-
-(int, optional)
-
-A seed for the Python random number generator (defaults to None).
-
 size
 ^^^^
 
--- a/integration_tests/additional_integration_tests.py
+++ b/integration_tests/additional_integration_tests.py
@@ -32,7 +32,6 @@ def run_depth_and_segmentation_test(controller, metadata_tier):
 
     # Initialize the test scene.
     step_metadata_0 = controller.start_scene(scene_data)
-    print(step_metadata_0.object_list)
 
     # Verify the depth map and object mask lists from step 0.
     if len(step_metadata_0.depth_map_list) != 1:
@@ -67,9 +66,6 @@ def run_depth_and_segmentation_test(controller, metadata_tier):
                 f'{len(step_metadata_1.object_mask_list)} should be length 1'
             )
     # Verify the consistent segment_color of each object across both steps.
-    print(step_metadata_0.object_list)
-    print(step_metadata_1.object_list)
-    # DW: these are both empty lists for oracle when they should not be
     if metadata_tier == 'oracle':
         if (
             step_metadata_0.object_list[0].uuid !=
@@ -136,7 +132,6 @@ def run_habituation_trial_counts_test(controller, metadata_tier):
             f'Step {step_metadata.step_number} failed: habituation_trial '
             f'{step_metadata.habituation_trial} != 1'
         )
-    print(step_metadata.object_list)
 
     # Try a couple of pass actions.
     for i in range(0, 2):
--- a/integration_tests/additional_integration_tests.py
+++ b/integration_tests/additional_integration_tests.py
@@ -32,6 +32,7 @@ def run_depth_and_segmentation_test(controller, metadata_tier):
 
     # Initialize the test scene.
     step_metadata_0 = controller.start_scene(scene_data)
+    print(step_metadata_0.object_list)
 
     # Verify the depth map and object mask lists from step 0.
     if len(step_metadata_0.depth_map_list) != 1:
--- a/integration_tests/additional_integration_tests.py
+++ b/integration_tests/additional_integration_tests.py
@@ -52,20 +52,23 @@ def run_depth_and_segmentation_test(controller, metadata_tier):
     step_metadata_1 = controller.step('Pass')
 
     # Verify the depth map list and object mask list from step 1.
-    if len(step_metadata_0.depth_map_list) != 1:
+    if len(step_metadata_1.depth_map_list) != 1:
         return (
             False,
             f'Step 1 {metadata_tier} failed: depth_map_list with length '
-            f'{len(step_metadata_0.depth_map_list)} should be length 1'
+            f'{len(step_metadata_1.depth_map_list)} should be length 1'
         )
     if metadata_tier == 'level2' or metadata_tier == 'oracle':
-        if len(step_metadata_0.object_mask_list) != 1:
+        if len(step_metadata_1.object_mask_list) != 1:
             return (
                 False,
                 f'Step 1 {metadata_tier} failed: object_mask_list with length '
-                f'{len(step_metadata_0.object_mask_list)} should be length 1'
+                f'{len(step_metadata_1.object_mask_list)} should be length 1'
             )
     # Verify the consistent segment_color of each object across both steps.
+    print(step_metadata_0.object_list)
+    print(step_metadata_1.object_list)
+    # DW: these are both empty lists for oracle when they should not be
     if metadata_tier == 'oracle':
         if (
             step_metadata_0.object_list[0].uuid !=
@@ -132,6 +135,7 @@ def run_habituation_trial_counts_test(controller, metadata_tier):
             f'Step {step_metadata.step_number} failed: habituation_trial '
             f'{step_metadata.habituation_trial} != 1'
         )
+    print(step_metadata.object_list)
 
     # Try a couple of pass actions.
     for i in range(0, 2):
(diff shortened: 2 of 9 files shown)


docs/source/install.rst
```diff
@@ -164,13 +164,6 @@ save_debug_json
 
 Save AI2-THOR/Unity input, AI2-THOR/Unity output, and MCS StepMetadata output to JSON file on each step. Default: False
 
-seed
```
deanwetherby, line 167: Seed is unused and therefore removed.

integration_tests/additional_integration_tests.py
```diff
@@ -52,18 +52,18 @@ def run_depth_and_segmentation_test(controller, metadata_tier):
     step_metadata_1 = controller.step('Pass')
 
     # Verify the depth map list and object mask list from step 1.
-    if len(step_metadata_0.depth_map_list) != 1:
+    if len(step_metadata_1.depth_map_list) != 1:
```
deanwetherby, line 55: Copy/paste error in test did not check against the correct variable.

integration_tests/run_handmade_tests.py
```diff
@@ -321,9 +321,12 @@ def start_handmade_tests(
                 successful_test_list.append((test_name, metadata_tier))
             else:
                 failed_test_list.append((test_name, metadata_tier, status))
+
         # Run each additional test at this metadata tier.
         for runner_function in (
-                [] if only_test_name else FUNCTION_LIST):
+                [func for func in FUNCTION_LIST
+                 if only_test_name is None or
+                 only_test_name in str(func)]):
```
deanwetherby, line 329: Allows for testing individual non-numbered integration tests.

machine_common_sense/config_manager.py
```diff
@@ -112,6 +112,13 @@ def _validate_screen_size(self):
                 str(self.SCREEN_WIDTH_DEFAULT)
             )
 
+    def is_file_writing_enabled(self):
+        return (
+            self.is_save_debug_images() or
+            self.is_save_debug_json() or
+            self.is_video_enabled()
+        )
```
deanwetherby, line 120: Made this part of the config mgr instead of the controller doing all that logic.

machine_common_sense/config_manager.py
```diff
@@ -135,13 +142,6 @@ def set_metadata_tier(self, mode):
             mode
         )
 
-    def get_seed(self):
-        return self._config.getint(
-            self.CONFIG_DEFAULT_SECTION,
-            self.CONFIG_SEED,
-            fallback=None
-        )
```
deanwetherby, line 143: Removed seed

machine_common_sense/config_manager.py
```diff
@@ -521,10 +521,25 @@ class SceneConfigurationSchema(Schema):
...
     training = fields.Bool()
 
-    @ post_load
+    @post_load
     def make_scene_configuration(self, data, **kwargs):
         return SceneConfiguration(**data)
 
+    # TODO DW: Needs testing/confirmation
+    @post_dump
+    def remove_none(self, d, **kwargs) -> Dict:
```
deanwetherby, line 530: Added post-processing `remove_none` function to the class via post_dump decorator. Seems cleaner.

machine_common_sense/controller.py
```diff
@@ -85,71 +82,24 @@ class Controller():
 
     Parameters
     ----------
-    unity_app_file: str
+    unity_app_file_path: str
```
deanwetherby, line 85: Updated docstring to match other documentation.

machine_common_sense/controller.py
```diff
@@ -85,71 +82,24 @@ class Controller():
...
-    TELEPORT_Z_POS = 'zPosition'
-    TELEPORT_Y_ROT = 'yRotation'
-
-    # Hard coding actions that effect MoveMagnitude so the appropriate
-    # value is set based off of the action
-    # TODO: Move this to an enum or some place, so that you can determine
-    # special move interactions that way
-    FORCE_ACTIONS = ["PushObject", "PullObject"]
-    OBJECT_MOVE_ACTIONS = ["CloseObject", "OpenObject"]
-    MOVE_ACTIONS = ["MoveAhead", "MoveLeft", "MoveRight", "MoveBack"]
```
deanwetherby, line 133: Moved all of these constants and all associated code to the parameter module.

machine_common_sense/parameter.py
```diff
@@ -0,0 +1,277 @@
...
+    def _convert_y_image_coord_for_unity(self, y_coord: int) -> int:
+        '''Pixel coordinates are expected to start at the top left,
+        but in Unity, (0,0) is the bottom left.
+        '''
+        if not isinstance(y_coord, int):
+            raise TypeError(f"{y_coord} y_coord is not an integer")
+        screen_height = self.config.get_screen_height()
+        if(0 <= y_coord < screen_height):
+            return screen_height - y_coord - 1
+        raise ValueError(f"{y_coord} is not in range 0-{screen_height-1}")
```
deanwetherby, line 93: This function was not correct when copied over from the controller. Updated and tested.

machine_common_sense/parameter.py
```diff
@@ -0,0 +1,277 @@
...
+        object_image_coords_x = int(self._get_number_with_default(
+            self.OBJECT_IMAGE_COORDS_X_KEY, self.DEFAULT_IMG_COORD, **kwargs))
+        object_image_coords_y = int(self._get_number_with_default(
+                                    self.OBJECT_IMAGE_COORDS_Y_KEY,
+                                    self.DEFAULT_IMG_COORD, **kwargs))
+        receptable_image_coords_x = int(self._get_number_with_default(
+            self.RECEPTACLE_IMAGE_COORDS_X_KEY,
+            self.DEFAULT_IMG_COORD,
+            **kwargs))
+        receptacle_image_coords_y = int(self._get_number_with_default(
```
deanwetherby, line 209: This variable was actually `receptacle_image_coords_x` when I copied it over from the controller. I'm not sure how this was working.

machine_common_sense/parameter.py
```diff
@@ -0,0 +1,277 @@
...
+        receptacle_vector = {
+            'x': receptable_image_coords_x,
+            'y': self._convert_y_image_coord_for_unity(
+                receptacle_image_coords_y)
+        }
+        move_magnitude = self._get_move_magnitude(action, force, amount)
+        (teleport_rotation, teleport_position) = self._get_teleport(**kwargs)
+
+        # TODO is this a feature we need?
+        if self.config.is_noise_enabled():
```
deanwetherby, line 234: I'd really like to remove noise.

machine_common_sense/parameter.py
```diff
@@ -0,0 +1,277 @@
...
+            horizon=horizon,
+            teleportRotation=teleport_rotation,
+            teleportPosition=teleport_position,
+            moveMagnitude=move_magnitude,
+            objectImageCoords=object_vector,
+            receptacleObjectImageCoords=receptacle_vector
+        )
+
+    def _mcs_action_to_ai2thor_action(self, action: str) -> str:
+        if action == Action.CLOSE_OBJECT.value:
```
deanwetherby, line 252: Why are we comparing strings to enums and not have them as enums all the way through the code base? There seems to be this pattern in a lot of places.

kdrumm (approved):
You bring up good questions that we should resolve before merging or add to a ticket for later.  However, I'm not sure the answers and the rest is good improvements!  All Tests passed.

machine_common_sense/parameter.py
rartiss55: This code looked okay in the last release/probably a naming mistake in recent refactoring, so I wouldn't worry about it too much. Least its fixed now!

docs/source/install.rst
ThomasSchellenbergNextCentury: CORA/Vikash had specifically requested this feature, but that was a while ago, and I'm not sure if they actually use it, but it might be worth checking with them first before removing it? Not sure.

ThomasSchellenbergNextCentury: Actually, I was confusing this with `noise`, which is what Vikash had specifically requested. I'm not sure if `seed` came from CORA/TA1 or from us. Probably not a big deal to remove it.

machine_common_sense/parameter.py
ThomasSchellenbergNextCentury: I think(?) there was a concern about strings being easier for TA1 to use? But enums seem pretty darn simple as well, so I wouldn't have any qualms about using them instead.

ThomasSchellenbergNextCentury (approved):
Looks good. All unit tests pass.

machine_common_sense/parameter.py
rartiss55: Been a while but even if we use or start with enums in the python API wrapper, does it also have to be converted back to a string anyway before forwarding it along to ai2thor?

rartiss55 (approved):

PR #456, opened 2022-01-12
Two ways of importing getch module

deanwetherby:
Adding getch to the MCS package init places that module within the auto-documentation path which I am not in favor of. The best way to handle this in my opinion is with try/except ImportError.


brianpippin (approved):
I assume it is okay to crash out if neither of them work?  (which shouldn't happen I would think)

deanwetherby:
> I assume it is okay to crash out if neither of them work? (which shouldn't happen I would think)

I agree otherwise we'll have all kinds of import try/except logic. Fail fast except for this special case where we try one more thing before failing.

kdrumm (approved):

PR #457, opened 2022-01-12
Seesaw works on pass,rotation, and other actions, integrations tests

jaudick:
Unity --- https://github.com/NextCenturyCorporation/ai2thor/pull/181

I updated 69 because matching the agent height to the structure below now occurs on pass


jaudick:
69 changed because matching the agent height to the structure below now occurs on pass

brianpippin (approved):

PR #458, opened 2022-01-12
MCS-1074 Plotter image comparison unit tests

deanwetherby:


tests/test_plotter.py
```diff
@@ -359,6 +362,42 @@ def test_draw_holes(self):
...
+        )
+
+        plotter = TopDownPlotter(
+            team="test",
+            scene_config=scene_config)
+        img = plotter.base_room_img.copy()
+        img = plotter._draw_holes(img, scene_config.holes)
+        holes_img = plotter._export_plot(img)
+        # save image to resources folder in the event of plotter changes
+        # pil_img.save(os.path.join(resources_path, 'plotter_holes.png'))
```
ThomasSchellenbergNextCentury, line 379: Good idea, and thanks for including the comment.

tests/test_plotter.py
```diff
@@ -359,6 +362,42 @@ def test_draw_holes(self):
...
+        truth_img = PIL.Image.open(
+            os.path.join(
+                resources_path,
+                'plotter_holes.png'))
+        # calculate image difference
+        diff = PIL.ImageChops.difference(holes_img, truth_img)
+        stat = ImageStat.Stat(diff)
+        self.assertListEqual(stat.sum, [0.0, 0.0, 0.0])
+
+    def test_draw_lava(self):
```
ThomasSchellenbergNextCentury, line 391: I must have missed the previous PR for this, but I'm a little concerned that lava looks like other red objects, when they are visible. Would it be unreasonable to ask for a minor tweak to the lava's appearance? Maybe add a black X in the middle of each lava section? If you like the idea, but not the timing, we can make a separate ticket to implement it.

deanwetherby: Completely agree. It wasn't clear how to distinguish lava floor texture so red squares were used until we figured out a better way to display. A black X wouldn't be difficult but I would like to make another ticket for that and keep this one to the new image comparison unit/regression tests.

ThomasSchellenbergNextCentury:
I really like this approach. I'm wondering if there are other tests we should add, like:
- Image containing a performer agent (they have noses now, right?)
- Image containing multiple objects, some of which are visible (filled)
- Image containing a target object (it uses a special icon, right?)
- Image containing different colored walls
- Image containing ramps and platforms

deanwetherby:
> I really like this approach. I'm wondering if there are other tests we should add, ...

Sure. I can add some or all of those.

tests/test_plotter.py
ThomasSchellenbergNextCentury: Sounds good. I'll make a ticket. I also vaguely remember punting on special icons for ramps and platforms, since they're not specifically labeled in the scenes or metadata. Does that ring a bell? Sound we revisit them at some point?

deanwetherby: I do recall that conversation. There's no difference between a ball, ramp, platform or a wall as far as the plotter is concerned.

ThomasSchellenbergNextCentury: https://nextcentury.atlassian.net/browse/MCS-1097

ThomasSchellenbergNextCentury (approved):

PR #460, opened 2022-01-13
MCS-1071 - bug fix for files being deleted incorrectly with --saveVideo flag enabled

rartiss55:
Hopefully my explanation makes sense: There was a weird edge case where we were deleting/attempting to delete all files within the directory from which a run is attempted. 

This ended up being an issue with not being able to determine a scene name -- generally, if we're saving debug files, we delete all the old local versions of debug files if we're running a scene multiple times so we can save the new versions without issue, and we look for those files in a directory like `./(scene_name)`. In this case, the scene file had no scene name and the code couldn't parse it properly from the path given, so it was blank, and therefore a delete was attempted in the current directory (in this case, machine_common_sense/scripts). I added a `name` for the problematic scenes in the docs and decided to throw an exception if a name does not exist for a scene.

### DIFF
--- a/docs/source/scenes/container_open_close.json
+++ b/docs/source/scenes/container_open_close.json
@@ -1,4 +1,5 @@
 {
+    "name": "container_open_close",
     "version": 2,
     "ceilingMaterial": "AI2-THOR/Materials/Walls/Drywall",
     "floorMaterial": "AI2-THOR/Materials/Fabrics/CarpetWhite 3",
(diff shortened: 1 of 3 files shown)


ThomasSchellenbergNextCentury (approved):
Thanks for tracking this down Rachel. I agree with raising an exception if the scene name is blank (I believe we have that property listed as required in our schema doc anyway).

brianpippin (approved):
I know we are trying to raise more exceptions instead of handling weird cases for TA1 I think this is the way to go as the developers on TA1 will know what must be done now when interacting with our API.

PR #461, opened 2022-01-13
added configurable logging

kdrumm:
Do we like this better?

This is somewhat of a prototype for something to reduce the amount of copied logging config between the ILE and MCS spurned on by a comment by @ThomasSchellenbergNextCentury.  We can discuss whether we like it, like it the old way, or have a better idea.  

Either way, I believe the ILE won't be able to use this until the next MCS release, right?  I don't think this warrants releasing early either.


kdrumm:
I thought about adding the new function to the docs.  In the end I slightly leaned towards not wanting to advertise it to TA1 and leave it as a somewhat hidden feature to be used between projects.  Thoughts?

ThomasSchellenbergNextCentury (commented):
1. Looks great! I particularly like how you were able to use the new function for the "errors only console config" as well.
2. I don't have any preference about publicizing the function in the docs versus keeping it hidden.
3. Yes, we'll need to keep the duplicate code in the ILE for now, until the next MCS release (or change the requirements file to use git rather than pypi, but that likely introduces more problems than it solves).
4. It would be nice to add a few unit tests for the new function.

deanwetherby (approved):

deanwetherby:
There is this section on logging:
https://nextcenturycorporation.github.io/MCS/install.html#logging

ThomasSchellenbergNextCentury (approved):
I get a deprecation warning in the unit tests for assertEquals, but otherwise looks good

PR #466, opened 2022-01-17
Fixed holding object placement so dropping work consistently

jaudick:
[Unity](https://github.com/NextCenturyCorporation/ai2thor/pull/186)

These updates only change where the objects are held in the agents hand when currently holding an object. It has nothing to do with where objects get placed. They are very small float amounts but all actions and results are the same.

### DIFF
--- a/integration_tests/data/005.ball_close_pickup.oracle.outputs.json
+++ b/integration_tests/data/005.ball_close_pickup.oracle.outputs.json
@@ -35,7 +35,7 @@
         "objects": [
             {
                 "direction_x": 0.0,
-                "direction_y": -0.559,
+                "direction_y": -0.558,
                 "direction_z": 0.829,
                 "distance": 0.591,
                 "held": true,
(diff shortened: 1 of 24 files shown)


rartiss55:
These updated tests work, but curious if anyone else is seeing this or something like it?:
FAILED INTEGRATION TESTS:
    050.drop_tall_object (oracle): Step 10 failed:
        position_z: -0.003974522929638624 != 0.0

rartiss55 (approved):

kdrumm (approved):

kdrumm:
> These updated tests work, but curious if anyone else is seeing this or something like it?: FAILED INTEGRATION TESTS: 050.drop_tall_object (oracle): Step 10 failed: position_z: -0.003974522929638624 != 0.0

I didn't see this when I ran the tests yesterday, however this after seeing your comment I reran test 050 and got some similar errors about every other time I run it.  Mine started at step 10, 11 or 15 depending on run.  One of my 6 runs also had position_x failures.  I think an argument could be made that these values are close enough and variation is ok when throwing, but I'm not sure where I stand.  Either way, I do think we want to make sure these tests pass consistently.

PR #472, opened 2022-01-18
Haptic feedback step output

jaudick:
[Unity](https://github.com/NextCenturyCorporation/ai2thor/pull/187)

Haptic feedback is currently an array of strings derived from an enum inside unity. The array has potential to be like ["LAVA", "UNDER_WATER"....] if we add more feedbacks. We could change the LAVA output to IN_LAVA or something like that too. Integration tests could be made too once we decide on how this should work.

### DIFF
--- a/integration_tests/data/066.end_habituation.scene.json
+++ b/integration_tests/data/066.end_habituation.scene.json
@@ -34,5 +34,8 @@
                 "z": 0.5
             }
         }]
-    }]
+    }],
+    "goal": {
+        "action_list": [[], [], [], ["EndHabituation"], [], [], [], ["EndHabituation,yRotation=30"], [], [], [], [], [], ["EndHabituation,xPosition=3"], [], [], ["EndHabituation,xPosition=0,zPosition=0"], [], [], [], [], [], [], ["EndHabituation,xPosition=0,zPosition=0,yRotation=0"]]
+    }
 }
--- a/integration_tests/habituation_trial_counts.scene.json
+++ b/integration_tests/habituation_trial_counts.scene.json
@@ -36,6 +36,7 @@
         }]
     }],
     "goal": {
+        "action_list": [[], [], ["EndHabituation"], [], [], [], [], [], [], [], [], [], [], ["EndHabituation"], ["EndHabituation"]],
         "habituation_total": 3
     }
 }
--- a/integration_tests/run_handmade_tests.py
+++ b/integration_tests/run_handmade_tests.py
@@ -211,6 +211,8 @@ def run_single_scene(controller, scene_filename, metadata_tier, dev, autofix):
 
     # Run the specific actions for the test scene.
     for index, action_data in enumerate(action_list + [None]):
+        if not step_metadata:
+            return False, f'Step {index} failed: output step_metadata is None'
         # Validate the test scene's output metadata at each action step.
         failed_validation_list = validate_single_output(
             expected_output_data_list[index],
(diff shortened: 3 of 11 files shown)


ThomasSchellenbergNextCentury:
Looking good so far. I started some discussions in the corresponding Unity PR. It looks like some of the StepMetadata unit tests are failing. Additionally, you might want to update the MCS Controller unit tests for `start_scene` and `step` to verify the haptic feedback variable in the function output.

deanwetherby:
What do you think about using a dictionary for haptic_feedback with boolean value for each feedback mechanism? 

```
haptic_feedback = {"in_lava": False, "on_ramp": True}
```

then checks by TA1 teams would be more readable as in

```
if metadata.haptic_feedback.get("in_lava"):
   # handle lava condition
```

rather than checking the existence of a string in a list

```
if "LAVA" in metadata.haptic_feedback:
  # handle lava condition
```

kdrumm:
I am getting integration test failures at test 66 related 'AttributeError: 'NoneType' object has no attribute 'action_list''  Anyone else?

jaudick:
> I am getting integration test failures at test 66 related 'AttributeError: 'NoneType' object has no attribute 'action_list'' Anyone else?

Yes I get that too

ThomasSchellenbergNextCentury:
@deanwetherby 

> What do you think about using a dictionary for haptic_feedback with boolean value for each feedback mechanism?

I like this idea, though I don't feel strongly either way (dictionary vs list).

I think there's a second question, which is: where would we want to put the dictionary? Right now, Unity handles making the variable, and Python just passes what it gets from Unity. Would we want to update Unity to use a dictionary, or convert the array in Python? I don't personally have a preference, and I'm not sure if one would be "better" than the other.

ThomasSchellenbergNextCentury:
@kdrumm 

> I am getting integration test failures at test 66 related 'AttributeError: 'NoneType' object has no attribute 'action_list'' Anyone else?

This _might_ be my fault. Let me investigate.

jaudick:
@deanwetherby @ThomasSchellenbergNextCentury 
Regarding dictionary vs list, I think the list implementation works just as good as a dictionary for understanding what haptic feedback is happening. If we did the dictionary I would probably do it on the unity side.

ThomasSchellenbergNextCentury:
@kdrumm @jaudick I believe this PR fixes the integration test error: https://github.com/NextCenturyCorporation/MCS/pull/473

jaudick:
> @kdrumm @jaudick I believe this PR fixes the integration test error: #473

Yes it fixed it

brianpippin:
> What do you think about using a dictionary for haptic_feedback with boolean value for each feedback mechanism?
> 
> ```
> haptic_feedback = {"in_lava": False, "on_ramp": True}
> ```
> 
> then checks by TA1 teams would be more readable as in
> 
> ```
> if metadata.haptic_feedback.get("in_lava"):
>    # handle lava condition
> ```
> 
> rather than checking the existence of a string in a list
> 
> ```
> if "LAVA" in metadata.haptic_feedback:
>   # handle lava condition
> ```

I think this implementation even if implemented in Unity would be preferable for interacting with our API.  I think it would make our API more readable and easier to interact with, which should be the primary function of the API.  The list leave a lot of onus on the API user to potentially write more code and have to handle each thing we add to this list.

jaudick:
@brianpippin Making it more readable/easier for the users makes sense, I will change it to a dictionary to alleviate the onus

jaudick:
@brianpippin @deanwetherby @ThomasSchellenbergNextCentury  I changed it to a dictionary. The changes are implemented on the unity branch

kdrumm (approved):
Once https://github.com/NextCenturyCorporation/MCS/pull/473 is merged in everything clears up and it all looks good.

brianpippin (approved):
Agree with Kyle's comment above.

ThomasSchellenbergNextCentury (approved):

deanwetherby (approved):

PR #473, opened 2022-01-19
Fixing MCS Integration Tests

ThomasSchellenbergNextCentury:
- Fixed the integration tests broken by PR https://github.com/NextCenturyCorporation/MCS/pull/462
- Fixed a bug in which we weren't handling numbers correctly when comparing the user's input parameters to the action restrictions (comparing numeric strings, ints, and floats).

### DIFF
--- a/machine_common_sense/controller.py
+++ b/machine_common_sense/controller.py
@@ -4,7 +4,7 @@ import glob
 import json
 import logging
 import os
-from typing import Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Union
 
 import ai2thor.controller
 import ai2thor.server
@@ -238,6 +238,18 @@ class Controller():
 
         return output
 
+    def _compare_param_values(self, value_1: Any, value_2: Any) -> bool:
+        """Compares two parameter values and returns if they are equal,
+        making sure that string numbers are converted to floats, and integer
+        floats are converted to ints."""
+        data = {'1': value_1, '2': value_2}
+        for key in data:
+            if isinstance(data[key], str) and data[key].isnumeric():
+                data[key] = float(data[key])
+            if isinstance(data[key], float) and data[key].is_integer():
+                data[key] = int(data[key])
+        return data['1'] == data['2']
+
     def _convert_scene_config(self, config_data) -> SceneConfiguration:
         if isinstance(config_data, SceneConfiguration):
             return config_data
@@ -270,7 +282,7 @@ class Controller():
         """
         if (self._goal.last_step is not None and
                 self._goal.last_step == self.__step_number):
-            logger.warning(
+            logger.error(
                 "You have passed the last step for this scene. "
                 "Ignoring your action. Please call controller.end_scene() "
                 "now.")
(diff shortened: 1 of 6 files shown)


jaudick (approved):

machine_common_sense/controller.py
```diff
@@ -283,25 +295,22 @@ def step(self, action: str, **kwargs) -> Optional[StepMetadata]:
...
-            logger.warning(
+            logger.error(
                 f"The given action '{action}' with parameters "
                 f"'{kwargs}' isn't in the action_list. Ignoring your action. "
                 f"Please call controller.step() with an action in the "
                 f"action_list. Possible actions at step {self.__step_number}:"
             )
             for action_data in action_list:
-                logger.warning(f'    {action_data}')
+                logger.error(f'    {action_data}')
```
kdrumm, line 313: Note to others, this change will cause an additional log when running the unit tests, specifically 'run_restricted_action_list_test'

kdrumm (approved):
I got the following log
``

machine_common_sense/controller.py
ThomasSchellenbergNextCentury: Ah, yeah, I changed that when I was debugging. I can change it back, though maybe it's better this way? This condition makes `step()` return `None`, which then might cause the calling script to fail for confusing reasons, which is a reason to keep the change... but I'm not sure.

machine_common_sense/controller.py
```diff
@@ -238,6 +238,18 @@ def start_scene(self, config_data: Union[SceneConfiguration, Dict]) -> \
 
         return output
 
+    def _compare_param_values(self, value_1: Any, value_2: Any) -> bool:
```
deanwetherby, line 241: Is it possible to move this function to the parameter module so as not to clutter the controller?

machine_common_sense/controller.py
kdrumm: It seems more correct to be an error log, but I prefer not seeing it in integration tests.  I'm fine with either.

brianpippin: I agree with Kyle's last statement.  Seems more correct as an error would be good to not appear in the tests.

machine_common_sense/controller.py
brianpippin: It would be good to move it to the parameter module now so we don't need to refactor it later imo

ThomasSchellenbergNextCentury: Great idea! Not only did I move the function over, but I also added unit tests (now that it's easier to test), which exposed a bug, that I then fixed.

machine_common_sense/controller.py
ThomasSchellenbergNextCentury: I ended up changing these back to warning. I figure most people will run using the default logging (that's set to DEBUG), so they'll see the warning message either way.

deanwetherby (approved):
Thanks for moving the function to the parameters module and writing unit tests. For the purposes of fixing the integration tests, I'm going to approve this PR. However, there is some fundamental/root issue here as this change seems to go counter to our recent push for raising exceptions. I'm not sure what to do about it quite yet though.

PR #474, opened 2022-01-19
Track number of steps on lava and add a default negative reward for each step

jaudick:
[UNITY](https://github.com/NextCenturyCorporation/ai2thor/pull/188)

### DIFF
--- a/machine_common_sense/controller_logger.py
+++ b/machine_common_sense/controller_logger.py
@@ -53,6 +53,7 @@ class ControllerLogger(AbstractControllerSubscriber):
         logger.debug("  POSITION: " + str(step_output.position))
         logger.debug("  ROTATION: " + str(step_output.rotation))
         logger.debug("  HAPTIC FEEDBACK: " + str(step_output.haptic_feedback))
+        logger.debug("  STEPS ON LAVA: " + str(step_output.steps_on_lava))
         logger.debug("OBJECTS: " +
                      str(len(step_output.object_list)) +
                      " TOTAL")
--- a/machine_common_sense/controller_logger.py
+++ b/machine_common_sense/controller_logger.py
@@ -53,7 +53,6 @@ class ControllerLogger(AbstractControllerSubscriber):
         logger.debug("  POSITION: " + str(step_output.position))
         logger.debug("  ROTATION: " + str(step_output.rotation))
         logger.debug("  HAPTIC FEEDBACK: " + str(step_output.haptic_feedback))
-        logger.debug("  STEPS ON LAVA: " + str(step_output.steps_on_lava))
         logger.debug("OBJECTS: " +
                      str(len(step_output.object_list)) +
                      " TOTAL")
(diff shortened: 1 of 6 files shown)


kdrumm (commented):
I am getting unit test failures.  I think they are due to the mock_controller needing to be updated.

machine_common_sense/reward.py
```diff
@@ -9,6 +9,7 @@
 GOAL_ACHIEVED = 1
 GOAL_NOT_ACHIEVED = 0
 STEP_PENALTY = 0.001
+LAVA_PENTALTY = 1000
```
ThomasSchellenbergNextCentury, line 12: Starting a discussion on this number:

1. I'm trying to remember how the reward code actually works. I believe you start at `1`, and then lose `0.001` each step, so after 1000 steps, your reward is negative.
2. If that's the case, then the LAVA_PENALTY should be between 0 and 1.
3. From last week's engineering collab meeting, we decided that a single step in the lava should be allowed. I took that concept to mean that the reward would still be positive after a single step in the lava. If we're in agreement, I propose that the LAVA_PENALTY be `0.5`. This means that a single step in the lava doesn't make the reward negative, but two steps does.

Thoughts?

machine_common_sense/reward.py
```diff
@@ -247,4 +251,5 @@ def calculate_reward(
                                                                      objects,
                                                                      agent,
                                                                      reach)
-        return Reward._adjust_score_penalty(current_score, number_steps)
+        return Reward._adjust_score_penalty(
+            current_score, number_steps, steps_on_lava)
```
ThomasSchellenbergNextCentury, line 255: We should add some unit tests for the new behavior.

machine_common_sense/controller_logger.py
```diff
@@ -53,6 +53,7 @@ def _write_debug_output(self, payload):
         logger.debug("  POSITION: " + str(step_output.position))
         logger.debug("  ROTATION: " + str(step_output.rotation))
         logger.debug("  HAPTIC FEEDBACK: " + str(step_output.haptic_feedback))
+        logger.debug("  STEPS ON LAVA: " + str(step_output.steps_on_lava))
```
ThomasSchellenbergNextCentury, line 56: I don't think this property needs to be in the debug output.

machine_common_sense/reward.py
deanwetherby: Seconded. 0.5 is a large enough penalty compared to 0.001 and we will provide an override in the configuration anyway.

machine_common_sense/reward.py
jaudick: I think unit tests should be a separate ticket, what do you think?

deanwetherby: Unit tests go hand-in-hand with the ticket and code modifications. You may, like I frequently do, discover an issue while testing.

jaudick:
@deanwetherby tests look good and pass for me

machine_common_sense/controller_logger.py
ThomasSchellenbergNextCentury: Bump

jaudick: Fixed

rartiss55 (approved):

ThomasSchellenbergNextCentury (approved):

PR #477, opened 2022-01-24
MCS-1097 plotter depictions

deanwetherby:
* Xs drawn in red lava cells to distinguish them from other red scene objects
* Ramps now illustrate directionality with the arrow peak pointing at the highest side (convention)

I have not address occluders, platforms or walls that were part of the ticket. I'm not sure what the intention was there.

plotter_lava
plotter_ramp
plotter_ramp_hidden


machine_common_sense/plotter.py
```diff
@@ -19,15 +19,19 @@
...
+    x: float
+    z: float
 
     def __sub__(self, coord: SceneCoord):
         return SceneCoord(self.x - coord.x, self.z - coord.z)
 
     def __add__(self, coord: SceneCoord):
         return SceneCoord(self.x + coord.x, self.z + coord.z)
 
+    def __truediv__(self, coord: SceneCoord):
```
deanwetherby, line 31: Perhaps too cute?

machine_common_sense/plotter.py
```diff
@@ -52,15 +56,47 @@ class Robot():
...
+        peak_mid = self._peak_midpoint()
+        floor_pts = self._floor_points()
+        return Arrow(floor=floor_pts, peak=peak_mid)
+
+    def _peak_midpoint(self) -> SceneCoord:
+        '''The midpoint of the highest ramp side'''
+        sorted_bounds = sorted(self.bounds, key=lambda p: p['y'])
+        peak_pts = sorted_bounds[-2:]
+        peak_scene_pts = [SceneCoord(x=pt['x'], z=pt['z']) for pt in peak_pts]
+        return peak_scene_pts[0] / peak_scene_pts[1]
```
deanwetherby, line 87: Mid point calculation from SceneCoord dunder truediv.

machine_common_sense/plotter.py
```diff
@@ -52,15 +56,47 @@ class Robot():
...
+
+    def _floor_points(self) -> List[SceneCoord]:
+        '''The points of the ramp side nearest the floor'''
+        ys = [p['y'] for p in self.bounds]
+        # get all of the indices for the value that is nearest 0 ie the floor
+        indices = np.where(np.isclose(ys, ys[np.argmin(np.abs(ys))]))
```
deanwetherby, line 93: A little complex in this one-liner but essentially gets multiple indices in the list of y's for the value nearest 0. We're are expecting two points nearest the floor at the same y-height.

machine_common_sense/plotter.py
```diff
@@ -309,45 +355,52 @@ def _draw_holes(self, img: np.ndarray, holes: List) -> np.ndarray:
...
             start=(upper_left.y, upper_left.x),
             end=(lower_right.y - 1, lower_right.x - 1),
             shape=img.shape[:2],
             clip=True)
-        img[rr, cc] = self.DEFAULT_COLOR
+        img[rr, cc] = color
         return img
 
-    def _draw_hole_x(
+    def _draw_x(
```
deanwetherby, line 385: Genericized draw_x to be useful for holes and lava and whatever else comes up.

machine_common_sense/plotter.py
```diff
@@ -52,15 +56,47 @@ class Robot():
...
 
 @dataclass
-class Object():
+class Arrow():
+    floor: List[SceneCoord]
+    peak: SceneCoord
+
+
+@dataclass
+class SceneAsset():
```
deanwetherby, line 65: Renamed Object to SceneAsset. It's been bugging me.

machine_common_sense/plotter.py
```diff
@@ -440,35 +493,33 @@ def _draw_objects(self, img: np.ndarray, objects: Dict,
...
-        return Object(
-            held=held,
-            visible=visible,
-            uuid=uuid,
+
+        obj_bounds = object_metadata.get('objectBounds')
+        corners = None if obj_bounds is None else obj_bounds.get(
+            'objectBoundsCorners')
+        bounds = None if corners is None else [SceneCoord(
+            x=corner['x'], z=corner['z']) for corner in corners]
```
deanwetherby, line 515: None checking here seems awkward. Started to think about using python-box here.

machine_common_sense/plotter.py
ThomasSchellenbergNextCentury: Maybe? I'm only skeptical because you're calculating the midpoint, rather than doing a straight division. I might instead expect this overload to return something like `SceneCoord(self.x / coord.x, self.z / coord.z)`

machine_common_sense/plotter.py
ThomasSchellenbergNextCentury: I like that name. My only concern is that Asset already has a meaning in Unity, but since we're using it outside of Unity, I don't think it's a problem.

machine_common_sense/plotter.py
```diff
@@ -52,15 +56,47 @@ class Robot():
...
+        '''The points of the ramp side nearest the floor'''
+        ys = [p['y'] for p in self.bounds]
+        # get all of the indices for the value that is nearest 0 ie the floor
+        indices = np.where(np.isclose(ys, ys[np.argmin(np.abs(ys))]))
+        return [SceneCoord(
+            self.bounds[int(indx)]['x'], self.bounds[int(indx)]['z'])
+            for indx in indices[0]]
+
+
+@ dataclass
```
ThomasSchellenbergNextCentury, line 99: I've noticed Kyle's editor does this as well, but you have extra whitespace here. I think the decorator still works either way, but I didn't know if you were concerned about being consistent throughout the file.

machine_common_sense/plotter.py
```diff
@@ -507,12 +564,42 @@ def _draw_object(self, img: np.ndarray,
         clr = colour.COLOR_NAME_TO_RGB.get(
             obj.color.lower(), self.DEFAULT_COLOR)
 
+        if(obj.uuid.startswith('ramp')):
+            self._draw_ramp_arrow(img, obj)
```
ThomasSchellenbergNextCentury, line 568: We might want to add a comment here about the naming assumption. As we discussed before, it's a fine assumption to make for now, but we may need to change it for some unforeseeable reason in the future.

machine_common_sense/plotter.py
deanwetherby: I don't know why that happens. It's not PEP-8 compliant AFAIK. I'll fix and try to figure out how to avoid that in the future.

machine_common_sense/plotter.py
deanwetherby: What about using the `__or__` instead? 

```
midpoint = SceneCoord(x=1, z=1) | SceneCoord(x=2, z=3)
# midpoint = SceneCoord(x=1.5, z=2.0)
```

ThomasSchellenbergNextCentury: That makes more sense to me. :+1:

ThomasSchellenbergNextCentury (approved):

PR #479, opened 2022-01-25
MCS-1131 Fixed action push parameter so it works as described in the API

ThomasSchellenbergNextCentury:
For our "force" actions (pull, push, and throw), the "force" input argument is intended to be a percentage from 0 to 1. We then multiply this force by the actual maximum force (in Newtons, I think) that we want to allow the robot to be able to apply in the Unity world. Since TA1 users would likely have no concept of what force values are useful in Unity, asking them to provide a simple percentage seemed more user-friendly than asking them to provide a value in Newtons (or whatever). I'm not tied to this approach, so I'm fine if we want to change it, but we'd need to update the documentation as well as the input parameter validation code.

Please note that while the previous max force was 50, that value was chosen somewhat arbitrarily, so I increased it to 250 in order to make tool pushing work properly.

### DIFF
--- a/machine_common_sense/parameter.py
+++ b/machine_common_sense/parameter.py
@@ -36,7 +36,7 @@ class Parameter:
     DEFAULT_IMG_COORD = 0
     DEFAULT_OBJECT_MOVE_AMOUNT = 1.0
 
-    FORCE_PERCENTAGE_MULTIPLIER = 250.0
+    UNITY_FORCE = 250.0
 
     MAX_AMOUNT = 1.0
     MIN_AMOUNT = 0.0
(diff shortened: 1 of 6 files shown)


machine_common_sense/parameter.py
```diff
@@ -37,8 +37,7 @@ class Parameter:
     DEFAULT_IMG_COORD = 0
     DEFAULT_OBJECT_MOVE_AMOUNT = 1.0
 
-    MAX_FORCE = 1.0
-    MIN_FORCE = 0.0
+    MAX_FORCE = 250.0
```
deanwetherby, line 40: This part has always been a little confusing to me and may have resulted in unexpected behavior after my parameter module refactor. If I'm understanding this now, we have a default Unity force (250.0) and a percentage force value between 0.0 and 1.0 that TA1 gives us. The resulting applied force is a percentage of the default Unity force. If true (and it may not be), I think renaming variables to reflect that notion may help with future readability. `force * MAX_FORCE` doesn't indicate that concept well.

ThomasSchellenbergNextCentury: I'm not opposed. Do you have any suggestions? Maybe something like FORCE_PERCENTAGE_MULTIPLIER?

deanwetherby (approved):
I was originally thinking the Unity force would be 250 and the 0.0-1.0 would be the percentage multiplier but this works too.

ThomasSchellenbergNextCentury:
> I was originally thinking the Unity force would be 250

Good idea

PR #480, opened 2022-01-26
Configurable reward and penalty scores

jaudick:
Right now the goal reward can be a float but before this change it was an int. How do we feel about it being a float if configured that way?

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -121,6 +121,14 @@ To use a specific configuration, you can either pass in a file path or dictionar
 Config File Properties
 **********************
 
+goal_reward
+^^^^^^^^^^^^^^^
+
+(float, optional)
+
+Changes the postive reward recieved for achieving a goal.
+
+
 history_enabled
 ^^^^^^^^^^^^^^^
 
@@ -142,6 +150,13 @@ The `metadata` property describes what metadata will be returned by the MCS Pyth
 
 If no metadata level is set:
 - `default`: Fallback if no metadata level is specified. Only meant for use during development (evaluations will never be run this way). Includes metadata for visible and held objects in the scene, as well as camera info and properties corresponding to the player. Does not include depth maps or object masks.
+
+lava_penalty
+^^^^^^^^^^^^^^^
+
+(float, optional)
+
+Changes the negative penalty recieved for every step on lava.
 
 noise_enabled
 ^^^^^^^^^^^^^^^
@@ -170,6 +185,13 @@ size
 (int, optional)
 
 Desired screen width. If value given, it must be more than `450`. If none given, screen width will default to `600`.
+
+step_penalty
+^^^^^^^^^^^^^^^
+
+(float, optional)
+
+Changes the negative penalty recieved for every step.
 
 video_enabled
 ^^^^^^^^^^^^^
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -126,7 +126,7 @@ goal_reward
 
 (float, optional)
 
-Changes the postive reward recieved for achieving a goal.
+Changes the postive reward recieved for achieving a goal. Default: 1
 
 
 history_enabled
@@ -156,7 +156,7 @@ lava_penalty
 
 (float, optional)
 
-Changes the negative penalty recieved for every step on lava.
+Changes the negative penalty recieved for every step on lava.  Default: 0.5
 
 noise_enabled
 ^^^^^^^^^^^^^^^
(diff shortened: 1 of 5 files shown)


docs/source/install.rst
```diff
@@ -121,6 +121,14 @@ To use a specific configuration, you can either pass in a file path or dictionar
 Config File Properties
 **********************
 
+goal_reward
+^^^^^^^^^^^^^^^
+
+(float, optional)
+
+Changes the postive reward recieved for achieving a goal.
```
ThomasSchellenbergNextCentury, line 129: Would be good for these three options to list their respective defaults.

tests/test_reward.py
```diff
@@ -565,6 +599,46 @@ def test_transferral_reward_on_top_of_with_pickedup_object(self):
...
+            # create upper plane (y = 1) + i
+            obj['objectBounds']['objectBoundsCorners'].append(
+                {'x': 0.0, 'y': 1.0 + i, 'z': 0.0})
+            obj['objectBounds']['objectBoundsCorners'].append(
+                {'x': 1.0, 'y': 1.0 + i, 'z': 0.0})
+            obj['objectBounds']['objectBoundsCorners'].append(
+                {'x': 1.0, 'y': 1.0 + i, 'z': 1.0})
+            obj['objectBounds']['objectBoundsCorners'].append(
+                {'x': 0.0, 'y': 1.0 + i, 'z': 1.0})
+            obj['position'] = {'x': 0.5, 'y': 0.0 + i, 'z': 0.5}
```
ThomasSchellenbergNextCentury, line 632: FYI In the future you may find it easier to do something like this:

```python
obj = {
    "objectId": str(i),
    "objectBounds": {
        "objectBoundsCorners": [
            {'x': 0.0, 'y': 0.0 + i, 'z': 0.0},
            {'x': 0.0, 'y': 0.0 + i, 'z': 0.0},
            {'x': 1.0, 'y': 0.0 + i, 'z': 0.0},
            {'x': 1.0, 'y': 0.0 + i, 'z': 1.0},
            {'x': 0.0, 'y': 0.0 + i, 'z': 1.0},
            {'x': 0.0, 'y': 1.0 + i, 'z': 0.0},
            {'x': 1.0, 'y': 1.0 + i, 'z': 0.0},
            {'x': 1.0, 'y': 1.0 + i, 'z': 1.0},
            {'x': 0.0, 'y': 1.0 + i, 'z': 1.0}
        ]
    },
    "position": {'x': 0.5, 'y': 0.0 + i, 'z': 0.5}
}
```

ThomasSchellenbergNextCentury (changes requested):
I'm getting a bunch of unit test errors. They're all `TypeError: float() argument must be a string or a number, not 'NoneType'`. Is it just me?

jaudick:
> I'm getting a bunch of unit test errors. They're all `TypeError: float() argument must be a string or a number, not 'NoneType'`. Is it just me?

Not just you, looks like I missed that configs were being set in this test file. I had only tested the reward unit tests. I will get on this

jaudick:
@ThomasSchellenbergNextCentury the error is fixed

deanwetherby (commented):
Some unit tests specifically for the score adjustment function would be good to add to this PR. There are numerous conditions to test for including penalty values of None and if the task has already been completed successfully.

jaudick:
> There are numerous conditions to test for including penalty values of None
@deanwetherby This is confusing me a little. Are you saying that None would be passed in intentionally to have penalty values of zero?

> and if the task has already been completed successfully.
Also confused here. Are you asking to test multiple values of goal_reward configs like: test 500, test 1.5, test 0.1? or test if penalties happen after the goal is achieved?

deanwetherby:
> @deanwetherby This is confusing me a little. Are you saying that None would be passed in intentionally to have penalty values of zero?
> 
> Are you asking to test multiple values of goal_reward configs like: test 500, test 1.5, test 0.1? or test if penalties happen after the goal is achieved?

You'll want to check each branching condition in the function. What do you expect to happen if step_penalty or lava_penalty is None? What if the current_score is 1 vs not 1? etc. The adjusted score output is going to vary on those conditions.

Unit tests are also a form of documentation so another developer can see how the function gets used and what to expect for an output.

```
        step_penalty = STEP_PENALTY if step_penalty is None else step_penalty
        lava_penalty = LAVA_PENALTY if lava_penalty is None else lava_penalty
        if current_score == 1:
            return current_score - ((number_steps - 1) * float(step_penalty))

        if steps_on_lava is None:
            steps_on_lava = 0
```

jaudick:
@deanwetherby I added three additional unit tests based on your suggestions

deanwetherby (approved):

ThomasSchellenbergNextCentury (approved):

brianpippin (approved):

PR #489, opened 2022-02-03
RotateObject action

jaudick:
[UNITY](https://github.com/NextCenturyCorporation/ai2thor/pull/196)

I'm not sure if I handled an input bool parameter for an action properly since this appears to be the first action with a bool parameter. This was how I got it to work:  take in a string of: "True/False" or "true/false" and convert to a bool whenever its not a bool. 

I also added controller tests for TorqueObject. I never made any when I merged that branch and felt it was necessary since this branch originally branched from the TorqueObject branch. The PushObject controller test also did not have a test with a default force parameter so I added one test in its test_controller method since I was already touching the force code of TorqueObject here


tests/test_parameter.py
```diff
@@ -236,6 +236,50 @@ def test_get_force(self):
...
+        self.assertIsInstance(clockwise, bool)
+        self.assertEqual(clockwise, False)
+
+        clockwise = self.parameter_converter._get_clockwise(clockwise=True)
+        self.assertIsInstance(clockwise, bool)
+        self.assertEqual(clockwise, True)
+
+        clockwise = self.parameter_converter._get_clockwise()
+        self.assertIsInstance(clockwise, bool)
+        self.assertEqual(clockwise, True)
```
ThomasSchellenbergNextCentury, line 266: I know you were questioning your input handling, but the fact that you have these good unit tests makes me pretty confident you're handling it well :)

machine_common_sense/parameter.py
```diff
@@ -251,6 +271,8 @@ def _validate_and_convert_params(self, **kwargs) -> Tuple:
         }
         move_magnitude = self._get_move_magnitude(action, force, amount)
         (teleport_rotation, teleport_position) = self._get_teleport(**kwargs)
+        clockwise = self._get_clockwise(**kwargs)
+        (teleport_rotation, teleport_position) = self._get_teleport(**kwargs)
```
ThomasSchellenbergNextCentury, line 275: I think you accidentally copied this one line (compare 273 and 275)

machine_common_sense/action.py
```diff
@@ -464,6 +464,53 @@ class Action(Enum):
...
 
+    ROTATE_OBJECT = (
+        "RotateObject",
+        "9",
+        "Apply rotation to a nearby object. (objectId=string, " +
+        "clockwise=bool(default:True), objectImageCoordsX=float, " +
+        "objectImageCoordsY=float)"
+    )
+    """
+    Apply rotation to a nearby object.
```
ThomasSchellenbergNextCentury, line 475: Please change this to: `Apply a rotation of 5 degrees to a nearby object.` (so it includes the rotation amount)

machine_common_sense/parameter.py
```diff
@@ -66,7 +68,7 @@ class Parameter:
     # # TODO: Move this to an enum or some place, so that you can determine
     # # special move interactions that way
     FORCE_ACTIONS = ["PushObject", "PullObject", "TorqueObject"]
-    OBJECT_MOVE_ACTIONS = ["CloseObject", "OpenObject"]
+    OBJECT_MOVE_ACTIONS = ["CloseObject", "OpenObject", "RotateObject"]
```
ThomasSchellenbergNextCentury, line 71: I don't think this needs to be here, does it? The rotation amount is always the same so it doesn't need those input checks.

jaudick: That makes sense, I made the requested changes

kdrumm:
Please make sure the unit tests pass

deanwetherby (approved):
I like all the unit tests.

kdrumm (approved):

PR #490, opened 2022-02-04
fix for not setting reward for retrieval tasks

kdrumm:
Integration tests kept crashing in PRs.  Once it happened in 2, I checked in the dev branch where it still occurred.  I decided to attempt to fix it.  Here is my fix.  I still get integration test failures, but this fixes the crash.  We should investigate why the failures occur (it could just be we need to adjust expected values).


deanwetherby:
I'm trying to confirm right now.  What is the condition where this applies? Is `goal_reward` explicitly set to None so the defaults don't apply? Should there be a corresponding unit test? 

I also think that the reward module is out of date as it was written for Eval 2 with transferral and traversal rewards. All of that code should be removed and the reward module simplified as a result.

kdrumm:
> I'm trying to confirm right now. What is the condition where this applies? Is `goal_reward` explicitly set to None so the defaults don't apply? Should there be a corresponding unit test?
> 
> I also think that the reward module is out of date as it was written for Eval 2 with transferral and traversal rewards. All of that code should be removed and the reward module simplified as a result.

The goal reward is not set in the config so it is set to None.  Them that None is passed into a function with a default, but since it is explicitly set, that default is not used.  Although I just realized my fix won't work if goal_reward is explicitly 0.

Either way, I'm not sure how this snuck into development, but is now holding up other PRs.  I'm trying to determine where.

machine_common_sense/reward.py
```diff
@@ -280,7 +280,7 @@ def calculate_reward(
             objects,
             agent,
             reach,
-            goal_reward
+            goal_reward or GOAL_ACHIEVED
```
ThomasSchellenbergNextCentury, line 283: If you change this line to `GOAL_ACHIEVED if goal_reward is None else goal_reward` then it will work with the `0` condition

ThomasSchellenbergNextCentury:
Shouldn't Jenkins have caught that the integration tests are failing on development right now? Maybe @brianpippin would know?

kdrumm:
> Shouldn't Jenkins have caught that the integration tests are failing on development right now? Maybe @brianpippin would know?

Just checked jenkins.  The logs are indicating that the process stopped, but it isn't recording as a failure.  This was a good idea to help identify the issue.  In build 288, integration tests failed to run properly at all, due to lack of instance capacity.  In build 289, they crashed at test 73 like I was experiencing.

My hypothesis:
PR #480 had an issue that caused the crash
PR #479 changed some values that caused some failures to some tests.  I'm far from convinced on this since the tests failed.
Build for PR 479 failed as a fluke.  (instance capacity)

kdrumm:
Here are all the integration test failures I'm getting now.  Does anyone have an opinion on autofixing vs investigating further?

I think autofixing 39, 96 and 97 seems viable as they all seem very close.  I'm more confused on what happened with 45.  Before we do this, we should verify that someone else is getting consistent values.

`
RUNNING SCENE: 039.stacking_blocks_many.scene.json
Step 8 failed:
        objects testBlock1 position_z: 0.9018111228942871 != 0.9007366
RUNNING SCENE: 045.pickup_and_put_with_image_coord.scene.json
Step 5 failed:
        objects testBall direction_y: -0.011101920157670975 != -0.003
        objects testBall distance: 0.9417247176170349 != 0.72
        objects testBall position_z: 1.1416666507720947 != 0.92
RUNNING SCENE: 096.pull_object.scene.json
Step 4 failed:
        objects testBall direction_z: 0.5036666393280029 != 0.505
        objects testBall position_z: 0.38595670461654663 != 0.388
RUNNING SCENE: 097.push_object.scene.json
Step 3 failed:
        objects testBall direction_y: -0.5729975700378418 != -0.576
        objects testBall direction_z: 0.8195571303367615 != 0.818
        objects testBall distance: 1.1553276777267456 != 1.15
        objects testBall position_z: 0.9468570351600647 != 0.94
Step 4 failed:
        objects testBall direction_y: -0.55560302734375 != -0.562
        objects testBall direction_z: 0.8314477205276489 != 0.827
        objects testBall distance: 1.191498041152954 != 1.179
        objects testBall position_z: 0.9906682968139648 != 0.975`

kdrumm:
At first, i changed the precision of floating point checks improperly.  After consulting @ThomasSchellenbergNextCentury, we corrected that and decided to autofix the one leftover problem in test 97.  I've run the full integration tests twice and saw no errors.  I believe this is ready to go.  Test 50 still may occasionally fail, but we have a separate ticket to worry about that and it doesn't fail consistently.

ThomasSchellenbergNextCentury (approved):

jaudick (approved):

deanwetherby (approved):

PR #494, opened 2022-02-04
Changed default lava penalty to 100.

ThomasSchellenbergNextCentury:
I've changed the default lava penalty to a big value. We've been talking around this for a while. I figured I'd go ahead and make the change. We can discuss any concerns as part of this PR.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -156,7 +156,7 @@ lava_penalty
 
 (float, optional)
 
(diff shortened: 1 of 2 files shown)


brianpippin (approved):
So long as the value can be overridden from the config file I don't think it matters much what the default value is.

kdrumm (approved):

jaudick (approved):

PR #502, opened 2022-02-15
fixing plotter rectangles

kdrumm:
Fixed plotter holes and lava drawing.  The draw function draws a rectangle around the points you give it so we just needed to move our points in by 1 pixel.


ThomasSchellenbergNextCentury (approved):
Looks fine. Tests pass.

PR #503, opened 2022-02-15
MCS-1122 Plotter sorting objects by height

deanwetherby:


machine_common_sense/plotter.py
```diff
@@ -499,7 +499,13 @@ def _find_plottable_objects(
             obj.get('objectId', '').startswith('floor')
         ]
         objects = scene_event.metadata.get('objects', [])
-        return filtered_structural_objects + objects
+        combined_objects = filtered_structural_objects + objects
+        # the combined list needs to be sorted by y height
+        # so that the lowest objects are drawn first
+        combined_objects = sorted(
+            combined_objects,
+            key=lambda obj: obj['objectBounds']['objectBoundsCorners'][0]['y'])
```
kdrumm, line 507: It looks like index 0 is usually the lowest y (its possible its not deterministic).  Shouldn't we use the max of all or if it is deterministic, it seems that index 7 is the high y.  I'm thinking how i'd expect to show a table over something under the table or a chest over something inside the chest.

deanwetherby: I was using the lowest for the case of an open container. The container would get drawn first because it's on the floor. The thing inside the container would go next.

kdrumm: OK, that makes sense.  Seems there is no perfect answer, but your case is probably more likely useful.

kdrumm (approved):

ThomasSchellenbergNextCentury (approved):
Looks fine. Tests pass.

PR #506, opened 2022-02-15
MCS-1179 Removed status from json loading

deanwetherby:
Relying on exceptions per previous conversations rather than relying on status strings in the event that something unexpected happens.

Former
`scene_data, status = mcs.load_json_scene_file('playroom.json')`

New
`scene_data = mcs.load_json_scene_file('playroom.json')`

New approach will properly raise a ValueError when the json scene file can't be properly parsed and a FileNotFoundError if the file itself can't be found. This is more descriptive than string parsing including additional possible exceptions raised by the json library and open function call.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ controller = mcs.create_controller(config_file_or_dict='./some-path/sample_confi
 
 # Either load the scene data dict from an MCS scene config JSON file or create your own.
 # We will give you the training scene config JSON files and the format to make your own.
-scene_data, status = mcs.load_scene_json_file(scene_json_file_path)
+scene_data = mcs.load_scene_json_file(scene_json_file_path)
 
 output = controller.start_scene(scene_data)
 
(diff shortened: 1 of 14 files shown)


rartiss55 (approved):

ThomasSchellenbergNextCentury (commented):
Seems like this change will break every TA1 run script. Do we want to support backwards compatibility by always returning a None status, while also throwing an error if one is raised?

deanwetherby:
> Seems like this change will break every TA1 run script. Do we want to support backwards compatibility by always returning a None status, while also throwing an error if one is raised?

It will and they'll have to change their code. There have been no guarantees made by me about keeping things static for the life of the program. We can't improve the API if we're always tied to past decisions. Returning a tuple with None doesn't seem useful to me. We'll just document this breaking change with release notes, slack messages, and support.

ThomasSchellenbergNextCentury (approved):
> It will and they'll have to change their code. There have been no guarantees made by me about keeping things static for the life of the program. We can't improve the API if we're always tied to past decisions. Returning a tuple with None doesn't seem useful to me. We'll just document this breaking change with release notes, slack messages, and support.'

Sounds good. Makes sense. Just pointing it out in case it was previously overlooked.

PR #507, opened 2022-02-16
MCS-1176 Docs and Integration tests for gaps in lips

kdrumm:
[Unity](https://github.com/NextCenturyCorporation/ai2thor/pull/204)
[MCS](https://github.com/NextCenturyCorporation/MCS/pull/507)
[ILE](https://github.com/NextCenturyCorporation/mcs-private/pull/142)

### DIFF
--- a/docs/source/schema.rst
+++ b/docs/source/schema.rst
@@ -187,6 +187,24 @@ Each **grid config** has the following properties:
 - `x` (integer)
 - `z` (integer)
 
+Lip Gaps Config
+**************
+
+Each **lip gaps config** has the following properties:
+
+- `front` (:ref:`lip gaps span config list <Lip Gaps Span Config>`):: Gaps on the positive Z axis
+- `back` (:ref:`lip gaps span config list <Lip Gaps Span Config>`):: Gaps on the negative Z axis
+- `left` (:ref:`lip gaps span config list <Lip Gaps Span Config>`):: Gaps on the negative X axis
+- `right` (:ref:`lip gaps span config list <Lip Gaps Span Config>`):: Gaps on the positive X axis
+
+Lip Gaps Span Config
+**************
+
+Each **lip gaps span config** has the following properties:
+
+- `low` (float): low end of the gap in a range 0 to 1 along the edge of the platform
+- `high` (float): high end of the gap in a range 0 to 1 along the edge of the platform
+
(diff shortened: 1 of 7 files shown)


integration_tests/data/101.gaps_in_lips.scene.json
```diff
@@ -0,0 +1,175 @@
...
+      "z": 15
+    },
+    "performerStart": {
+      "position": {
+        "x": 1.3,
+        "y": 0.5,
+        "z": 2.15
+      },
+      "rotation": {
+        "x": 0,
```
jaudick, line 160: I would change this to 60 just for when we visually watch this test to see whats going on

jaudick (approved):

integration_tests/data/101.gaps_in_lips.scene.json
kdrumm: If I rotate the ramp, I am going to need to rotate the performer as well.  I did sacrifice visibility to try to limit the number of steps required in the test, but I think that is desired.

docs/source/schema.rst
```diff
@@ -187,6 +187,24 @@ Each **grid config** has the following properties:
...
+- `left` (:ref:`lip gaps span config list <Lip Gaps Span Config>`):: Gaps on the negative X axis
+- `right` (:ref:`lip gaps span config list <Lip Gaps Span Config>`):: Gaps on the positive X axis
+
+Lip Gaps Span Config
+**************
+
+Each **lip gaps span config** has the following properties:
+
+- `low` (float): low end of the gap in a range 0 to 1 along the edge of the platform
+- `high` (float): high end of the gap in a range 0 to 1 along the edge of the platform
```
ThomasSchellenbergNextCentury, line 206: So do these values get multiplied by the platform's scale, since the lips are represented as child objects? We might want to add a note here.

kdrumm: They are 0 means one end of the edge and 1 is the other end.  We use those to determine scale and position of any lips on either side of the gap.  Its not as simple as just they are multiplied by the scale.  Sounds like we should update the docs though.  I changed it, let me know if that makes sense or if you have any suggestions.

ThomasSchellenbergNextCentury: The updated description is perfect. Thanks!

PR #509, opened 2022-02-16
MCS-1169 secret action params

deanwetherby:
Updated EndHabituation action to return no parameters with step metadata but retrieve those parameters from the goal metadata.


machine_common_sense/controller.py
```diff
@@ -354,6 +342,22 @@ def step(self, action: str, **kwargs) -> Optional[StepMetadata]:
 
         return output
 
+    def _build_endhabituation_params(self, action: str) -> str:
```
deanwetherby, line 345: Kept the rebuilt of end habituation parameters isolated but I have an open question of where this really belongs. Controller doesn't seem right. Open to suggestions or I'll decide what to do tomorrow.

machine_common_sense/controller.py
```diff
@@ -303,12 +297,11 @@ def step(self, action: str, **kwargs) -> Optional[StepMetadata]:
...
                 f"The given action '{action}' with parameters "
                 f"'{kwargs}' isn't in the action_list. Ignoring your action. "
-                f"Please call controller.step() with an action in the "
-                f"action_list. Possible actions at step {self.__step_number}:"
+                f"Possible actions at step {self.__step_number}:"
             )
             for action_data in action_list:
                 logger.error(f'    {action_data}')
-            return None
+            raise ValueError(f"{action}-{kwargs} not in {action_list}")
```
deanwetherby, line 304: Raising an exception if the action/param isn't in the list. Logging and returning None doesn't help as some performers were in an infinite loop during evals.

machine_common_sense/controller.py
```diff
@@ -170,9 +171,9 @@ def start_scene(self, config_data: Union[SceneConfiguration, Dict]) -> \
                             'file cannot be empty.')
 
         if (self._config.is_file_writing_enabled()):
-            os.makedirs('./' + scene_config.name, exist_ok=True)
-            self.__output_folder = './' + scene_config.name + '/'
-            file_list = glob.glob(self.__output_folder + '*')
+            os.makedirs(f'./{scene_config.name}', exist_ok=True)
+            self.__output_folder = f'./{scene_config.name}/'
+            file_list = glob.glob(f'{self.__output_folder}*')
```
deanwetherby, line 176: cleaning up string concatenation with f-strings

tests/test_goal_metadata.py
```diff
@@ -48,6 +48,68 @@ def test_metadata(self):
         self.assertFalse(self.goal_metadata.metadata)
         self.assertIsInstance(self.goal_metadata.metadata, dict)
 
+    def test_retrieve_action_list_with_preview_phase(self):
```
deanwetherby, line 51: Added a few more unit tests to illustrate/document the various conditions.

kdrumm (approved):

machine_common_sense/controller.py
```diff
@@ -235,18 +236,6 @@ def start_scene(self, config_data: Union[SceneConfiguration, Dict]) -> \
...
-        making sure that string numbers are converted to floats, and integer
-        floats are converted to ints."""
-        data = {'1': value_1, '2': value_2}
-        for key in data:
-            if isinstance(data[key], str) and data[key].isnumeric():
-                data[key] = float(data[key])
-            if isinstance(data[key], float) and data[key].is_integer():
-                data[key] = int(data[key])
-        return data['1'] == data['2']
-
```
ThomasSchellenbergNextCentury, line 249: Was this function deleted because it isn't actually being used? (It looks like I previously copied this function into `parameter.py`, but forgot to remove it here.)

deanwetherby: Correct. I use VS Code to find where it was used ("Find All References") and it wasn't used anywhere.

machine_common_sense/controller.py
ThomasSchellenbergNextCentury: Maybe in `parameter.py`, or even `action.py`?

machine_common_sense/controller.py
```diff
@@ -285,11 +274,16 @@ def step(self, action: str, **kwargs) -> Optional[StepMetadata]:
                 "now.")
             return None
 
+        # reformulate hidden EndHabituation parameters
+        if action == Action.END_HABITUATION.value:
+            action = self._build_endhabituation_params(action)
```
ThomasSchellenbergNextCentury, line 279: Should we add a sentence explaining this behavior to the documentation for `action_list` (in StepMetadata's docstring)?

tests/test_goal_metadata.py
```diff
@@ -90,6 +152,18 @@ def test_retrieve_action_list_at_step(self):
...
+        goal_metadata = mcs.GoalMetadata(action_list=[
+            ['Pass'],
+            ['LookDown', 'LookUp', 'RotateLeft', 'RotateRight', 'Pass'],
+            [],
+            ['EndHabituation,xPosition=0,zPosition=0,yRotation=90'],
+            [('PickupObject', {'objectId': 'target'})]
+        ])
+        self.assertEqual(goal_metadata.retrieve_action_list_at_step(3), [
+            ('EndHabituation', {})
+        ])
```
ThomasSchellenbergNextCentury, line 165: This seems a little redundant with the existing `test_retrieve_action_list_at_step_with_custom_action_list`. Maybe have a simpler test case containing just the EndHabituation action?

machine_common_sense/controller.py
deanwetherby: I should do that since the EndHabituation action could cause robot displacement.

machine_common_sense/controller.py
ThomasSchellenbergNextCentury: Thanks for confirming that. :+1:

machine_common_sense/controller.py
deanwetherby: I should also update in action.py

PR #513, opened 2022-02-18
MCS-1181 enum usage

deanwetherby:
Updating MCS Python API to leverage existing Action enum rather than relying on magic strings continually. One could argue that the input to our system should also be an Action enum rather than a string to avoid potential parsing issues.


kdrumm (approved):

machine_common_sense/parameter.py
```diff
@@ -309,21 +299,22 @@ def _validate_and_convert_params(self, **kwargs) -> Tuple:
             clockwise=clockwise
         )
 
-    def _mcs_action_to_ai2thor_action(self, action: str) -> str:
-        if action == Action.CLOSE_OBJECT.value:
+    def _mcs_action_to_ai2thor_action(self, action: Action) -> str:
+        # TODO MCS-1181 this does feel like it belongs in Action
+        if action == Action.CLOSE_OBJECT:
```
ThomasSchellenbergNextCentury, line 304: Is this TODO still relevant?

machine_common_sense/scene_history.py
```diff
@@ -7,7 +7,7 @@ class SceneHistory(object):
     def __init__(
         self,
         step=-1,
-        action=None,
+        action=None,  # MCS-1181 Enum vs string
```
ThomasSchellenbergNextCentury, line 10: Is this comment still relevant?

ThomasSchellenbergNextCentury (approved):
There isn't a good way to enforce this, is there? Our team just needs to be more aware of this pattern, and more diligent about using it?

deanwetherby:
> There isn't a good way to enforce this, is there? Our team just needs to be more aware of this pattern, and more diligent about using it?

It's not so much that this is a pattern itself but the ticket/PR were created to address the "anti-pattern" of magic strings. There is no way to enforce other than to recognize the misuse of magic strings instead of using the Action enum.

PR #514, opened 2022-02-21
MCS-1173 Added holes and lava to step metadata

kdrumm:


machine_common_sense/material.py
```diff
@@ -72,3 +72,6 @@ def verify_material_enum_string(enum_string):
             return True
         except KeyError:
             return False
+
+
+LAVA_TEXTURES = ["Stylized Lava Texture/Materials/Stylize_Lava_diffuse"]
```
kdrumm, line 77: I don't like this, but I'm not sure what else to do.  We have the lava texture as a constant in 3 repos now.  (here, in one of the registries in AI2Thor, in the scene generator)

ThomasSchellenbergNextCentury: If it makes more sense now, we can split out the "lava" from the `floorTextures` as separate properties in the JSON scene files. (It seemed fine early on to combine them, but I'm always happy to reevaluate design choices, especially if making a change now leads to easier understandability and maintainability long term.) Yes, we'd need to make the changes across three repos, but I don't think it would all take longer than a day or two (and now's the best time to do it, if we choose to).

kdrumm: Yeah, I think that would be a good idea.  I'll add a ticket.

machine_common_sense/step_metadata.py
```diff
@@ -171,7 +177,9 @@ def __init__(
             {} if haptic_feedback is None else haptic_feedback
         )
         self.head_tilt = head_tilt
+        self.holes = holes
         self.image_list = [] if image_list is None else image_list
+        self.lava = lava
```
ThomasSchellenbergNextCentury, line 182: We should update the unit tests to make sure that holes and lava are returned at oracle metadata level and not others. I think they're the `test_wrap_output_*` tests in `test_controller_output_handler.py`

machine_common_sense/controller_output_handler.py
```diff
@@ -283,7 +283,9 @@ def _get_step_metadata(self, goal, habituation_trial,
             ),
             haptic_feedback=self._scene_event.haptic_feedback,
             head_tilt=self._scene_event.head_tilt,
+            holes=None if restrict_non_oracle else self._scene_config.holes,
```
ThomasSchellenbergNextCentury, line 286: I think this is pass-by-reference and, if so, then any changes the user makes to `StepMetadata.holes` will also affect `self._scene_config.holes`, which we obviously don't want... So I think we need to deep copy it? Unless I'm mistaken / confused?

machine_common_sense/step_metadata.py
kdrumm: done

ThomasSchellenbergNextCentury (approved):

rartiss55 (approved):

PR #515, opened 2022-02-21
MCS-1164: End Scene if too many steps in Lava

brianpippin:
Add a config for the maximum number of steps allowed on lava.  The default is 0.  If the user exceeds the maximum allowable steps, the only option is to end the scene.

### DIFF
--- a/machine_common_sense/config_manager.py
+++ b/machine_common_sense/config_manager.py
@@ -63,6 +63,7 @@ class ConfigManager(object):
     CONFIG_TEAM = 'team'
     CONFIG_VIDEO_ENABLED = 'video_enabled'
     CONFIG_LAVA_PENALTY = 'lava_penalty'
+    CONFIG_STEPS_ALLOWED_IN_LAVA = 'steps_allowed_in_lava'
     CONFIG_STEP_PENALTY = 'step_penalty'
     CONFIG_GOAL_REWARD = 'goal_reward'
 
@@ -70,6 +71,9 @@ class ConfigManager(object):
     # on this assumption.
     SCREEN_WIDTH_DEFAULT = 600
     SCREEN_WIDTH_MIN = 450
+
+    # Default steps allowed in lava before calling end scene
+    STEPS_ALLOWED_IN_LAVA_DEFAULT = 0
 
     def __init__(self, config_file_or_dict=None):
         '''
@@ -242,6 +246,13 @@ class ConfigManager(object):
             self.CONFIG_DEFAULT_SECTION,
             self.CONFIG_GOAL_REWARD,
             fallback=None
+        )
+
+    def get_steps_allowed_in_lava(self):
+        return self._config.getint(
+            self.CONFIG_DEFAULT_SECTION,
+            self.CONFIG_STEPS_ALLOWED_IN_LAVA,
+            fallback=self.STEPS_ALLOWED_IN_LAVA_DEFAULT
         )
 
 
@@ -824,7 +835,7 @@ class SceneConfiguration:
                 return [str(state) for state in state_list]
         return []
 
-    def retrieve_goal(self):
+    def retrieve_goal(self, config):
         if not self.goal:
             return self.update_goal_target_image(GoalMetadata())
 
@@ -852,6 +863,7 @@ class SceneConfiguration:
                 last_preview_phase_step=(goal.last_preview_phase_step or 0),
                 last_step=goal.last_step or None,
                 metadata=goal.metadata or {},
+                steps_allowed_in_lava=config.get_steps_allowed_in_lava()
             )
         )
 
--- a/machine_common_sense/config_manager.py
+++ b/machine_common_sense/config_manager.py
@@ -835,9 +835,11 @@ class SceneConfiguration:
                 return [str(state) for state in state_list]
         return []
 
-    def retrieve_goal(self, config):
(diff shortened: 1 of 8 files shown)


machine_common_sense/config_manager.py
```diff
@@ -824,7 +835,7 @@ def retrieve_object_states(self,
                 return [str(state) for state in state_list]
         return []
 
-    def retrieve_goal(self):
+    def retrieve_goal(self, config):
```
deanwetherby, line 838: Maybe not pass in the whole config but instead just the number of steps in lava allowed?

machine_common_sense/controller.py
```diff
@@ -329,7 +338,7 @@ def step(self, action: str, **kwargs) -> Optional[StepMetadata]:
             action=action, **kwargs)
         step_output = self._controller.step(ai2thor_step)
 
-        (pre_restrict_output, output) = self._output_handler.handle_output(
+        (pre_restrict_output, output, self.__steps_in_lava) = self._output_handler.handle_output(
```
deanwetherby, line 341: Are we returning the steps in lava as part of the step metadata? Would that make the steps_in_lava return redundant?

brianpippin: We are, but I wasn't sure if it made sense to add it as it's own, or pull out of one of the outputs, and if it was abundantly clear which output to pull it from

machine_common_sense/config_manager.py
brianpippin: Yes this makes sense for sure.

machine_common_sense/controller.py
```diff
@@ -274,6 +275,14 @@ def step(self, action: str, **kwargs) -> Optional[StepMetadata]:
                 "now.")
             return None
 
+        # if they call end scene action they should have
+        #   called end_scene instead of step
+        if action == Action.END_SCENE.value:
+            logger.error(
+                "You have called EndScene action. "
+                "Please call controller.end_scene() now.")
+            return None
```
ThomasSchellenbergNextCentury, line 284: Should we just automatically call `end_scene` here? We couldn't include any of the optional function parameters in the call, but those parameters are just used for the passive scenes anyway, so they shouldn't be relevant.

brianpippin: We can't call end scene for them, calling end scene in our code doesn't do anything with their controller, I tried this path and it wasn't very fruitful.  It just leads to multiple end scene calls when they try to shut things down.

ThomasSchellenbergNextCentury: Oh, that's unfortunate. So, to clarify, the workflow would be:
1. Agent steps on lava
2. Agent receives EndScene as only possible action
3. Agent would need to call `end_scene()`, rather than `step("EndScene")`

(I don't have a problem with this -- just wanted to make sure I understood)

brianpippin: Yes if they called a different action like MoveAhead it would crash, because it wasn't in the action list.  If they called "EndScene" it would remind them to call "end_scene".  They would need to know they need to call end_scene.

deanwetherby: Should `return None` be replaced with raising an exception?

brianpippin: That raises an interesting thing I was thinking about this morning.  If we give them an action to call and it's EndScene, and they call EndScene and we say no call end_scene that what is really is the point of the EndScene action?  Maybe it would be better to return an empty action list so they know they can't do any actions and that means they need to call end_scene in that case any action would raise an exception.  Or we could also have EndScene raise an exception, but it almost seems counter productive to give them an action that just leads to an exception, or it could lead to an infinite loop of them trying to call EndScene.   Just some free form thoughts.

deanwetherby: An empty list currently means that all available actions can be used. No matter what we do here, it still seems quirky.

deanwetherby: https://nextcenturycorporation.github.io/MCS/api.html?highlight=action#machine_common_sense.StepMetadata

brianpippin: We don't know for sure that they are checking for empty arrays, so they either change their code to check for [] or they change it to check for a list with one item containing endScene, I think either way is good, not changing the empty array behavior probably provides the least amount of collision.  We can always revisit.

rartiss55:
Just to make sure this is the correct behavior since its what I'm seeing, but there were a lot of possibilities discussed: if I am told "EndScene" is my only action available in the action list, but I call another action instead, I get an exception and it quits.

brianpippin:
> Just to make sure this is the correct behavior since its what I'm seeing, but there were a lot of possibilities discussed: if I am told "EndScene" is my only action available in the action list, but I call another action instead, I get an exception and it quits.

Not currently, but I am changing it to that.

deanwetherby:
I had another thought on this since the EndScene action is so unlike our other Actions. Can we `os.kill()` the process? I know calling `end_scene()` internally does nothing but stopping the process might be ok.

brianpippin:
> I had another thought on this since the EndScene action is so unlike our other Actions. Can we `os.kill()` the process? I know calling `end_scene()` internally does nothing but stopping the process might be ok.

I made these changes and this actually functions better and acts just like calling end_scene

kdrumm (approved):

PR #517, opened 2022-02-22
Grid system for holes and lava

jaudick:
[Scene-Gen](https://github.com/NextCenturyCorporation/mcs-private/pull/146)
[MCS](https://github.com/NextCenturyCorporation/MCS/pull/517)
[Unity](https://github.com/NextCenturyCorporation/ai2thor/pull/208)

[Teams thread about a potential other solution](https://gov.teams.microsoft.us/l/message/19:gcch:959d7221399c4fe08b3ac2def85e1f0d@thread.skype/1645466647969?tenantId=74cf14f4-38e0-460b-9d96-c0a51cb4a55c&groupId=bd0afc79-5b79-4a03-a04a-00fc785501de&parentMessageId=1645466647969&teamName=PIAGET%20(MCS)&channelName=MCS%20Technical&createdTime=1645466647969)


brianpippin:
This seems like a pretty big change for holes and lava.  I feel like we should only have one grid system, it seems like this would be extremely confusing for TA1.  Perhaps we should have a design meeting to discuss this at some point.

machine_common_sense/plotter.py
```diff
@@ -309,7 +314,9 @@ def _draw_floor_textures(
             texture_color = colour.COLOR_NAME_TO_RGB['red']
 
             for position in floor_texture.positions:
-                texture_pos = SceneCoord(x=position.x, y=0, z=position.z)
+                texture_pos = SceneCoord(
+                    x=-((room_dim_x - 1) / 2) + position.x,
+                    y=0, z=-((room_dim_z - 1) / 2) + position.z)
```
ThomasSchellenbergNextCentury, line 319: You might want to extract this conversion code into a separate one-line function -- something like `_convert_floor_grid_coordinates` ?

ThomasSchellenbergNextCentury (commented):
I like this change. I don't think there's a perfect/straightforward solution, but this approach seems like the simplest one. Plus, since holes and lava haven't been around for that long, I'm not super concerned about backwards compatibility -- like Dean said (paraphrasing), sometimes we have to break things in order to make them better.

ThomasSchellenbergNextCentury (approved):

brianpippin:
I think this solution is  going to cause us big headaches.  Sometimes the simplest solutions aren't always the best long term solutions and I think a bigger discussion needs to occur.  We need one grid system throughout our repos I believe as we need to think primarily about the usability it will cause TA1 and not the ease for TA2 to code solutions.  We should strive to make things as ease as possible for TA1.

ThomasSchellenbergNextCentury:
> I think this solution is going to cause us big headaches. Sometimes the simplest solutions aren't always the best long term solutions and I think a bigger discussion needs to occur. We need one grid system throughout our repos I believe as we need to think primarily about the usability it will cause TA1 and not the ease for TA2 to code solutions. We should strive to make things as ease as possible for TA1.

If making the grid system for the holes/lava sync up with the other objects is so important, then maybe we should consider keeping the current behavior, and simply handling the holes/lava-underneath-room-walls case in the plotter code?

jaudick:
> > I think this solution is going to cause us big headaches. Sometimes the simplest solutions aren't always the best long term solutions and I think a bigger discussion needs to occur. We need one grid system throughout our repos I believe as we need to think primarily about the usability it will cause TA1 and not the ease for TA2 to code solutions. We should strive to make things as ease as possible for TA1.
> 
> If making the grid system for the holes/lava sync up with the other objects is so important, then maybe we should consider keeping the current behavior, and simply handling the holes/lava-underneath-room-walls case in the plotter code?

@brianpippin @ThomasSchellenbergNextCentury @deanwetherby Even though i did all this work I would completely support scrapping this implementation and doing the simpler: **"handling the holes/lava-underneath-room-walls case in the plotter code"** Perhaps we just detect when floor texture or hole it out of bounds and just cut ones off directly under walls to be half drawn while ones outside walls are not drawn at all.

PR #522, opened 2022-02-25
Convert to f-strings

deanwetherby:
Converted string concatenation and .format() to modern f-strings. I used flynt with the -tc option to accomplish this conversion automatically. Flynt requires python 3.8+ for that option so I just made a temporary virtual environment to do the conversion.

`(venv) $ flynt -tc .`

https://github.com/ikamensh/flynt#command-line-options


kdrumm (approved):

deanwetherby:
> Is this something we can/should enforce with flake8?

We should probably do that. I'll make a ticket to do the appropriate amount of research but enforcement will not be part of this PR.

PR #526, opened 2022-03-01
MCS-1184 Moved images & video to s3 and linked to docs

deanwetherby:
Created an `mcs-documentation` bucket open to the public as read-only. Changed local image and video references to new s3 urls. This should reduce our repo size by ~800 MB which will also make GH Actions faster.


deanwetherby:
I thought about moving the scenes folder as well but it's only a few hundred KB.

deanwetherby:
You can switch to this branch and run 'make html' from the docs folder. Then use your file browser to find and open the index.html. I would appreciate extra scrutiny.

deanwetherby:
> > I thought about moving the scenes folder as well but it's only a few hundred KB.
> 
> It's also useful to have some example scenes with the source code -- sometimes I use them for runs.

That's true. I do that sometimes too. Ok. leaving scenes in the repo.

ThomasSchellenbergNextCentury (approved):
LGTM

PR #529, opened 2022-03-02
Changing assertion tests to use context manager instead of lambda

deanwetherby:


ThomasSchellenbergNextCentury (approved):
Unit tests all pass for me.

brianpippin (approved):

PR #530, opened 2022-03-02
MCS-1115 Fixed integration test script. Updated integration test 095 and 050.

ThomasSchellenbergNextCentury:


integration_tests/run_handmade_tests.py
```diff
@@ -149,7 +149,7 @@ def validate_single_output(expected, actual):
                 actual_data,
                 rel_tol=0.01,
                 abs_tol=0.01
-            )] for expected_number in expected_list)
+            ) for expected_number in expected_list])
```
ThomasSchellenbergNextCentury, line 152: The typo I mentioned on Teams.

deanwetherby (approved):

jaudick:
@ThomasSchellenbergNextCentury test 50 should be fixed now. I ran all tests with oracle 3 times and then all tiers in order 3 times to make sure.

kdrumm (approved):

jaudick (approved):

integration_tests/run_handmade_tests.py
```diff
@@ -135,21 +135,18 @@ def validate_single_output(expected, actual):
...
+                failed = (not math.isclose(
+                    expected_data,
+                    actual_data,
+                    rel_tol=0.01,
+                    abs_tol=0.01
+                ))
+            elif isinstance(expected_data, list):
+                # Assume that the list is a numerical range.
+                min_max = list(sorted(expected_data))
+                failed = (min_max[1] < actual_data < min_max[0])
```
ThomasSchellenbergNextCentury, line 149: As luck would have it, test 050 failed on my first try :(

```
RUNNING SCENE: 050.drop_tall_object.scene.json
Step 11 failed:
        position_z: -0.010710992850363255 != 0.0
Step 12 failed:
        position_z: -0.010710992850363255 != 0.0
Step 13 failed:
        position_z: -0.010710992850363255 != 0.0
Step 14 failed:
        position_z: -0.010710992850363255 != 0.0
Step 15 failed:
        position_z: -0.0285969041287899 != [0.0, -0.015]
Step 16 failed:
        position_z: -0.03298699110746384 != [0.0, -0.019]
Step 17 failed:
        position_z: -0.03298699110746384 != [0.0, -0.019]
Step 18 failed:
        position_z: -0.03298699110746384 != [0.0, -0.019]
```

I came up with a new solution, though: Rather than the array being specific possible values, it's a range (min/max) of possible values. The test only fails if the actual value falls outside of the expected range. It should obviously be used sparingly, but seems like nice functionality to have available.

As for the root cause of this inconsistency, I'm now wondering if the large ball you drop on step 2 sometimes rolls too close to you and bumps you over a little? Maybe something to investigate at a future time.

Let me know what you think. Thanks.

kdrumm (approved):
I like it!

integration_tests/run_handmade_tests.py
brianpippin: Do we have a ticket for investigating the inconsistency?  It would be nice to know if is the structure of the test or some underlying code issue that causes this to behave ever so slightly differently each time it might be run.

brianpippin (approved):

PR #535, opened 2022-03-03
Object IDs from image coordinates output metadata

jaudick:
[Unity](https://github.com/NextCenturyCorporation/ai2thor/pull/218/files)

### DIFF
--- a/tests/test_step_metadata.py
+++ b/tests/test_step_metadata.py
@@ -36,6 +36,8 @@ class TestStepMetadata(unittest.TestCase):
         "performer_reach": 0.0,
         "physics_frames_per_second": 0,
         "position": {},
+        "resolved_object": "",
+        "resolved_receptacle": "",
         "return_status": "UNDEFINED",
         "reward": 0,
         "rotation": 0.0,
@@ -73,6 +75,8 @@ class TestStepMetadata(unittest.TestCase):
         "performer_reach": 0.0,
         "physics_frames_per_second": 0,
         "position": {},
+        "resolved_object": "",
+        "resolved_receptacle": "",
         "return_status": "UNDEFINED",
         "reward": 0,
         "rotation": 0.0,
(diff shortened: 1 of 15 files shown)


tests/test_step_metadata.py
```diff
@@ -159,6 +163,14 @@ def test_performer_reach(self):
     def test_position(self):
         self.assertIsInstance(self.step_metadata.position, dict)
 
+    def test_resolved_object(self):
+        self.assertEqual(self.step_metadata.resolved_object, '')
+        self.assertIsInstance(self.step_metadata.resolved_object, str)
+
+    def test_resolved_receptacle(self):
+        self.assertEqual(self.step_metadata.resolved_receptacle, '')
+        self.assertIsInstance(self.step_metadata.resolved_receptacle, str)
```
deanwetherby, line 172: I'd like to see unit tests explaining the other expectations returning a non-empty string. Unit tests also serve as documentation.

ThomasSchellenbergNextCentury (commented):
The integration tests pass, and I really appreciate your updates to them, but I agree with Dean's comment that we need to update a few more unit tests (in particular, the `test_wrap_output*` tests in `tests/test_controller_output_handler.py`).

tests/test_step_metadata.py
jaudick: @deanwetherby Could you give an example of what you would like to see for _'unit tests explaining the other expectations returning a non-empty string.'_

deanwetherby: @jaudick Besides an empty string, what can the values for resolved object and resolved receptable look like? Empty string is the default but there are no assurances in these tests that it can take on any other value.

jaudick:
@deanwetherby @ThomasSchellenbergNextCentury I added new unit test changes

deanwetherby (approved):

ThomasSchellenbergNextCentury (approved):

PR #537, opened 2022-03-09
MCS-960 save target position per step

rartiss55:
Saving the target position at each step to the history file, which is needed for the scorecard to account for tasks with moving targets where their original position may be incorrect. 

Associated ingest/scorecard PR here: https://github.com/NextCenturyCorporation/mcs-ingest/pull/75

### DIFF
--- a/machine_common_sense/history_writer.py
+++ b/machine_common_sense/history_writer.py
@@ -143,11 +143,12 @@ class HistoryWriter(object):
                         self.history_obj,
                         cls=NumpyEncoder))
 
-    def filter_history_output(
+    def update_history_output(
             self,
             history: SceneHistory) -> SceneHistory:
         """ filter out images from the step history data and
-            object lists and action list """
+            object lists and action list, incorporate
+            additional target info if needed """
         if history.output:
             history.output.action_list = None
             history.output.structural_object_list = None
@@ -199,7 +200,7 @@ class HistoryWriter(object):
                     step_obj)
             logger.debug("Adding history step")
             self.current_steps.append(
-                dict(self.filter_history_output(step_obj)))
+                dict(self.update_history_output(step_obj)))
 
     def is_target_visible(
             self,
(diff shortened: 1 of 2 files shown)


machine_common_sense/history_writer.py
```diff
@@ -143,23 +143,44 @@ def write_file(self):
...
+                    target_id = history.output.goal.metadata[target].get('id',
+                                                                         None)
+
+                    target_info = list(filter(
+                        lambda x: x.uuid == target_id,
+                        history.output.object_list))[0]
+
+                    history.output.goal.metadata[target][
+                        "position"
+                    ] = target_info.position
```
rartiss55, line 181: Originally was going to leave the target object as the only object in the object_list, but moving the "position" info to goal.metadata["target"] made more sense to me.

machine_common_sense/history_writer.py
```diff
@@ -143,23 +143,44 @@ def write_file(self):
...
+                ):
+                    # Save target position at each step for history file/
+                    # scorecard purposes (accounts for tasks where the
+                    # target position changes)
+                    target_id = history.output.goal.metadata[target].get('id',
+                                                                         None)
+
+                    target_info = list(filter(
+                        lambda x: x.uuid == target_id,
+                        history.output.object_list))[0]
```
deanwetherby, line 177: I prefer using next() here to avoid any potential index errors with the hard-coded "[0]". Next also allows for a default value that you can check against more safely. I used the same approach in the Reward class `__get_object_from_list` function.

```
target_info = next((obj for obj in history.output.object_list if obj.uuid == target_id), None)
```

kdrumm (approved):
Approved assuming Dean's comment is resolved.

machine_common_sense/history_writer.py
rartiss55: Thanks! Made that change/now check below that target_info isn't None.

deanwetherby (approved):

brianpippin (approved):

ThomasSchellenbergNextCentury (approved):

PR #538, opened 2022-03-10
Update test for passive scene to better catch issues in the future.

rartiss55:
Update related to the recent bug with passive scenes I mentioned this morning (belatedly made a ticket here: https://nextcentury.atlassian.net/browse/MCS-1239)

Thomas brought up the fair point on the Unity PR that the integration tests should've caught this. It didn't because our integration test passive scene was empty and only had one frame (this error in particular didn't result in Unity closing and you could still "Pass", but the frames just never updated properly after the first one). I added a moving occluder to the passive scene test (069) so that if something like this happens again, the test should fail. Note that this should pass with the recent ai2thor development build, but will fail (correctly) against 0.5.1.

### DIFF
--- a/integration_tests/data/069.intuitive_physics_scene_start.actions.txt
+++ b/integration_tests/data/069.intuitive_physics_scene_start.actions.txt
@@ -1,1 +1,11 @@
 ﻿Pass
+Pass
+Pass
+Pass
+Pass
+Pass
+Pass
+Pass
+Pass
+Pass
+Pass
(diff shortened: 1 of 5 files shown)


brianpippin (approved):

ThomasSchellenbergNextCentury (approved):
As expected, test 69 failed with 0.5.1, but passed with `--mcs_unity_version dev`. Thanks for figuring this out and fixing it!

deanwetherby (approved):

PR #539, opened 2022-03-11
MCS-1240 Fixed issue with ramps in top-down plotter.

ThomasSchellenbergNextCentury:
Now that we've changed ramps from being planks (long thin cubes) to wedges (triangles), I needed to make a few tweaks to how they're rendered by the top-down plotter. There was also a bug with the order in which the ramp box and its arrow were drawn (I've also updated the plotter's unit tests to detect this issue in the future).


machine_common_sense/plotter.py
```diff
@@ -86,23 +86,17 @@ def arrow(self) -> Arrow:
 
     def _peak_midpoint(self) -> SceneCoord:
         '''The midpoint of the highest ramp side'''
-        sorted_bounds = sorted(self.bounds, key=lambda p: p.y)
-        peak_pts = sorted_bounds[-2:]
+        # Safe to assume the bounds will always be in a consistent order.
+        peak_pts = self.bounds[-4:-2]
```
deanwetherby, line 90: I prefer a defensive coding posture over making assumptions that could be broken in the future.

machine_common_sense/plotter.py
```diff
@@ -86,23 +86,17 @@ def arrow(self) -> Arrow:
...
-        ys = [p.y for p in self.bounds]
-        # get all of the indices for the value that is nearest 0 ie the floor
-        indices = np.where(np.isclose(ys, ys[np.argmin(np.abs(ys))]))
-        return [SceneCoord(
-            self.bounds[int(indx)].x,
-            self.bounds[int(indx)].y,
-            self.bounds[int(indx)].z
-        ) for indx in indices[0]]
+        # Safe to assume the bounds will always be in a consistent order.
+        return [SceneCoord(pt.x, pt.y, pt.z) for pt in self.bounds[:4]]
```
deanwetherby, line 99: I suspect the change to wedge cause issues here. Like mentioned in another comment, I would prefer more defensive coding to handle this piece.

machine_common_sense/plotter.py
```diff
@@ -689,8 +685,11 @@ def _draw_arrow(self, img: np.ndarray, ramp: Ramp) -> np.ndarray:
                 c0=peak_pt.x,
                 r1=img_pt.y,
                 c1=img_pt.x)
-            img[rr, cc] = self.BACKGROUND_COLOR \
-                if ramp.visible else self.DEFAULT_COLOR
+            img[rr, cc] = (
+                self.BACKGROUND_COLOR
+                if (ramp_color != self.BACKGROUND_COLOR and ramp.visible) else
```
deanwetherby, line 690: We have BACKGROUND_COLOR ramps? Another interesting  use for the ramp color (since you're passing it in), is to use it for the arrow in the !visible case rather than the DEFAULT_COLOR.

deanwetherby (commented):
Something looks off on the upper and lower lines in the not visible ramp box.

machine_common_sense/plotter.py
ThomasSchellenbergNextCentury: A wedge's bounds contain 8 points; 4 of those points have an equal low Y value, and 4 of them have an equal high Y value. Therefore simply sorting on the Y value would still necessitate assuming that the values we need are the elements at indices 4 and 5. So, while I agree in the abstract, I'm not really sure how to address it here? Definitely open to suggestions.

machine_common_sense/plotter.py
ThomasSchellenbergNextCentury: It's theoretically possible to have black or white ramps (or even other objects), though I'm not sure how often it actually occurs. The scene I was using for testing had white ramps. If specific colors are a problem, we could add checks in the scene generator to avoid using them.

I like your idea of using the ramp color for the arrow color if the ramp isn't visible!

ThomasSchellenbergNextCentury: Updated. I also fixed the issue with the lines in the ramp test image.

machine_common_sense/plotter.py
```diff
@@ -86,23 +122,24 @@ def arrow(self) -> Arrow:
 
     def _peak_midpoint(self) -> SceneCoord:
         '''The midpoint of the highest ramp side'''
-        sorted_bounds = sorted(self.bounds, key=lambda p: p.y)
-        peak_pts = sorted_bounds[-2:]
+        # Assumes ramps are always wedges (triangles).
+        peak_pts = [
+            self.bounds.top_front_right_corner(),
+            self.bounds.top_front_left_corner()
+        ]
```
ThomasSchellenbergNextCentury, line 129: @deanwetherby I'm not sure if you think this is "better" or not -- open to feedback.

I created a SceneBounds class that takes the list of 8 bounding box points and exposes them via functions like "top_front_right_corner", etc. I'm hoping this will at least make the behavior clearer.

These assumptions become problematic when we start assigning X/Z rotation, which was the case previously for the plank ramps, but is not the case for wedge ramps. I added a TODO about this in the SceneBounds, but I didn't try to tackle the math (since it sounds like this top-down plotter will be getting phased out soon, spending the time on it didn't seem worthwhile).

deanwetherby: I like it more. I'd be hesitant to put much more effort into the plotter if Plan B works out. Good to know that someone is familiar enough with the plotter to make these changes.

deanwetherby (approved):

kdrumm (approved):

PR #542, opened 2022-03-14
MCS-1238 pydantic conversion

deanwetherby:
Converted most marshmallow classes/schema to pydantic.


machine_common_sense/controller.py
```diff
@@ -248,8 +238,7 @@ def start_scene(
     def _convert_scene_config(self, config_data) -> SceneConfiguration:
         if isinstance(config_data, SceneConfiguration):
             return config_data
-        schema = SceneConfigurationSchema()
-        return schema.load(config_data)
+        return SceneConfiguration(**config_data)
```
deanwetherby, line 241: Easier class construction unpacking the config_data dictionary.

machine_common_sense/plotter.py
```diff
@@ -309,7 +309,8 @@ def _out_of_bounds(
     def _even_room_dim_edge_size_reducer(
             self,
             pos: SceneCoord,
-            room_dim: Vector3d = Vector3d(10, 0, 10)) -> Tuple(SceneCoord):
+            room_dim: Vector3d = Vector3d(x=10, y=0, z=10)
```
deanwetherby, line 312: Pydantic will not allow positional constructor arguments so x, y and z must be specified.

tests/test_config_manager.py
```diff
@@ -419,135 +418,145 @@ def test_objects(self):
...
+        self.assertIsNone(object_1.resizes)
+        self.assertIsNone(object_1.rotates)
+        self.assertIsNone(object_1.salient_materials)
+        self.assertIsNone(object_1.shows)
+        self.assertIsNone(object_1.shrouds)
+        self.assertIsNone(object_1.states)
+        self.assertIsNone(object_1.structure)
+        self.assertIsNone(object_1.teleports)
+        self.assertIsNone(object_1.toggle_physics)
+        self.assertIsNone(object_1.torques)
```
deanwetherby, line 462: Updated a bunch of tests that didn't conform to unittest/MCS standards.

tests/test_config_manager.py
```diff
@@ -692,11 +701,12 @@ def test_update_goal_target_image_oracle_img_as_str(self):
...
 
 
-class TestSchemeConfigurationSchema(unittest.TestCase):
+class TestSchemeConfiguration(unittest.TestCase):
 
     def setUp(self):
-        self.scheme_config = SceneConfigurationSchema()
+        self.scheme_config = SceneConfiguration()
 
+    @unittest.skip
```
deanwetherby, line 709: Skipping the tests associated with removing attributes with a None value for now.

machine_common_sense/config_manager.py
```diff
@@ -2,28 +2,30 @@
...
-    # There is probably a class like this in python somewhere
-    # but i don't know where it is.
-    # TODO change later, potentially rename?
-    x: float = 0
-    y: float = 0
-    z: float = 0
+def to_camel_case(string: str) -> str:
+    words = string.split('_')
+    return ''.join(word.capitalize() if word !=
+                   words[0] else word for word in words)
```
deanwetherby, line 20: helper function converts python variables names `herp_derp` to camel case `herpDerp`.

machine_common_sense/config_manager.py
```diff
@@ -2,28 +2,30 @@
...
+
+
+class BaseModel(PydanticBaseModel):
```
deanwetherby, line 23: Create a custom BaseModel class in order to provide a module configuration.

machine_common_sense/config_manager.py
```diff
@@ -44,14 +46,218 @@ class MetadataTier(Enum):
...
+    habituation_total: Optional[int]
+    info_list: Optional[List[str]]
+    last_preview_phase_step: Optional[int]
+    last_step: Optional[int]
+    metadata: Optional[Dict]
+    objects_info: Optional[Dict]
+    scene_info: Optional[Dict]
+    skip_preview_phase: Optional[bool]
+    task_list: List[str] = None
+    type_list: List[str] = None
```
deanwetherby, line 85: Many Goal fields are wrapped in Optional typing because fields are missing arbitrarily in unit tests.

machine_common_sense/config_manager.py
```diff
@@ -44,14 +46,218 @@ class MetadataTier(Enum):
...
+
+
+# TODO duplicate class name (GoalMetadata)
+# TODO TargetMetadata maybe?
+# TODO Is this class being used anywhere?
+class TargetMetadata(BaseModel):
+    choose: GoalObject
+    target: GoalObject
+    target_1: GoalObject
+    target_2: GoalObject
```
deanwetherby, line 67: Like the comments say, this was originally named `GoalMetadata` and I can't find where/if it was being used.

deanwetherby:
All unit tests and integration tests now pass.

machine_common_sense/controller_events.py
```diff
@@ -1,17 +1,31 @@
...
+                   words[0] else word for word in words)
+
+
+class BaseModel(PydanticBaseModel):
+    # global configs are enabled by creating a custom BaseModel class
+    # contain the desired config
+    class Config:
+        alias_generator = to_camel_case
+        allow_population_by_field_name = True
+        arbitrary_types_allowed = True
```
deanwetherby, line 26: This attribute allows for a class to be used as an attribute but not validated other than ensuring the type.

machine_common_sense/controller_events.py
```diff
@@ -22,52 +36,42 @@ class EventType(enum.Enum):
...
 
 
-@dataclass
 class BasePostActionEventPayload(BaseEventPayload):
-    output_folder: str
+    output_folder: Optional[str]
     timestamp: str
     wrapped_step: dict
-    step_metadata: Event  # ai2thor.server.event
+    step_metadata: Union[Event, MultiAgentEvent]
```
deanwetherby, line 49: Unsure why sometimes the output from ai2thor is an Event vs a MultiAgentEvent. Or is it always multi?

machine_common_sense/controller_events.py
```diff
@@ -22,52 +36,42 @@ class EventType(enum.Enum):
...
     step_output: StepMetadata
     restricted_step_output: StepMetadata
     goal: GoalMetadata
 
 
-@dataclass
 class StartScenePayload(BasePostActionEventPayload):
-    pass
+    ...
```
deanwetherby, line 56: For some reason I prefer ellipses here. Does the same thing.

integration_tests/additional_integration_tests.py
```diff
@@ -199,30 +193,6 @@ def run_habituation_trial_counts_test(controller, metadata_tier):
     return True, ''
 
 
-def run_numpy_array_data_test(controller, metadata_tier):
```
ThomasSchellenbergNextCentury, line 202: I'm assuming if someone tries to run a scene containing numpy arrays, it will immediately crash? I'm assuming no one will care, especially if they're using the ILE scene generator to make scenes now.

machine_common_sense/config_manager.py
ThomasSchellenbergNextCentury: Am I correct in understanding that we no longer need two separate classes for each object (schema and non-schema)? If so, that's great!

machine_common_sense/config_manager.py
```diff
@@ -44,14 +46,218 @@ class MetadataTier(Enum):
     DEFAULT = 'default'
 
 
-class ConfigManager(object):
+class Vector3d(BaseModel):
+    x: float = 0.0
+    y: float = 0.0
+    z: float = 0.0
```
ThomasSchellenbergNextCentury, line 52: Will this still have a constructor that accepts 3 floats? We use that everywhere in the scene generator.

machine_common_sense/config_manager.py
ThomasSchellenbergNextCentury: Seems fine to me. I don't see GoalMetadata used anywhere in the scene generator either.

integration_tests/additional_integration_tests.py
deanwetherby: Some kind of exception will be raised in this case but that is the direction we need to go. Follow the API or blow up.

machine_common_sense/config_manager.py
deanwetherby: This class accepts 3 floats but you do have to specify the arguments unfortunately. No more `Vector3d(0,0,0)`. It will have to be `Vector3d(x=0,y=0,z=0)`.

machine_common_sense/controller_events.py
```diff
@@ -1,17 +1,31 @@
...
 
 from .config_manager import ConfigManager, SceneConfiguration
 from .goal_metadata import GoalMetadata
 from .step_metadata import StepMetadata
 
 
+def to_camel_case(string: str) -> str:
+    words = string.split('_')
+    return ''.join(word.capitalize() if word !=
+                   words[0] else word for word in words)
```
ThomasSchellenbergNextCentury, line 17: Isn't this the same function that's defined in `config_manager`?

machine_common_sense/controller_events.py
```diff
@@ -1,17 +1,31 @@
...
+
+
+class BaseModel(PydanticBaseModel):
```
ThomasSchellenbergNextCentury, line 20: There's also a BaseModel class in `config_manager`. Should we name them distinctly, like `BaseConfigModel` and `BaseEventModel`?

machine_common_sense/controller_events.py
deanwetherby: It is. Need to find the right place to put it and import in both locations.

machine_common_sense/controller_events.py
ThomasSchellenbergNextCentury: I think an Event can contain one or more MultiAgentEvents...? But, yeah, I think a Union is fine.

machine_common_sense/controller_events.py
deanwetherby: It's a pydantic formalism to inherit from BaseModel. I'm only customizing the BaseModel in order to provide module configuration that won't need to be repeated in every class.

machine_common_sense/config_manager.py
```diff
@@ -470,8 +470,8 @@ class SceneConfiguration(BaseModel):
     floor_textures: List[FloorTexturesConfig] = None
     goal: Optional[Goal]
     holes: List[Vector2dInt] = []
-    intuitive_physics: Optional[bool]
-    isometric: Optional[bool]
+    intuitive_physics: bool = False
+    isometric: bool = False
```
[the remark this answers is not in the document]
ThomasSchellenbergNextCentury: Whatever pydantic is doing must be pretty cool. I was concerned about setting default values to empty lists and vectors, but pydantic is apparently taking care of that:

```
>>> from machine_common_sense.config_manager import SceneConfiguration, Vector2dInt
>>> a = SceneConfiguration()
>>> b = SceneConfiguration()
>>> a.lava.append(Vector2dInt(x=0,z=0))
>>> a.lava
[Vector2dInt(x=0, z=0)]
>>> b.lava
[]
```

Neat!

machine_common_sense/config_manager.py
ThomasSchellenbergNextCentury: This will unfortunately require updating hundreds of lines of code in the scene generator (or creating and using a separate vector class). Would it be possible to add a second, flexible constructor, at least just for the vectors?

deanwetherby: I'll think about options here.

machine_common_sense/controller_events.py
deanwetherby: Change to import from config manager. I wanted to put it in parameters but that caused a circular import situation.

machine_common_sense/config_manager.py
deanwetherby: Not having a lot of luck with your request. If I change Vector3d to be a python dataclass, the unit test for config_manager doesn't pass on SceneConfig. I changed Vector3d to be a pydantic dataclass which got around that issue but failed when running the integration tests because the pydantic dataclass is not json serializable.

kdrumm: Could we just write a simple (Ha!) sed script to fix all the vector3d's in the ILE?  We could do this now.

ThomasSchellenbergNextCentury: Sounds like you're volunteering? :)

deanwetherby: Multiple encoders didn't work out. It's a take it / leave it situation.

ThomasSchellenbergNextCentury (approved):
Overall I like this a lot. I tried running every Eval 4 validation scene for a few steps, and it never crashed, so I'm assuming it will work with all of the old data. Here's the command I used, if anyone's interested (and, yes, this works with passive scenes too):

```
python machine_common_sense/scripts/run_interactive_scenes_pass.py --last_step 2 --mcs_unity_version dev ~/eval_4_validation/
```

In general, I like enforcing the keyword arguments. For simple classes, like the vectors, I'd prefer to allow positional arguments as well, but, even though that's not possible, the Pros of this change still greatly outweigh the Cons.

machine_common_sense/config_manager.py
brianpippin: I know it's more work but I like the fact that it is Vector3d(x=0,y=0,z=0) the explicit nature of the call ensures we are doing it correctly.  Updating lots of code though does make it harder but in the end it is worth it i think.

machine_common_sense/config_manager.py
brianpippin: It seems like we need a ticket to determine which of these field are actually Optional and then update the unit tests accordingly.

brianpippin (approved):

PR #547, opened 2022-03-21
MCS-1197:  End Scene Integration Tests

brianpippin:
- Add ability to have configs per integration test, metadata levels
- Ability to call EndScene and not have it end the integration tests


integration_tests/data/119.lava_end_scene.level1.outputs.json
```diff
@@ -0,0 +1,100 @@
...
+        "lava": null,
+        "objects_count": 0,
+        "position_x": null,
+        "position_z": null,
+        "return_status": "SUCCESSFUL",
+        "reward": -100.005,
+        "rotation_y": null,
+        "step_number": 5,
+        "structural_objects_count": 0
+    },
```
ThomasSchellenbergNextCentury, line 85: If you want to, you can add an extra check here by adding an `"action_list"` property to verify that the returned `action_list` restricts the robot to only using the `"EndScene"` action:

```json
"action_list": [["EndScene", {}]]
```

Note that, in `run_handmade_tests.py` line 38, you'll need to convert the elements within `actual.action_list` to tuples, like we're doing with `actual.holes` and `actual.lava` a few lines below.

integration_tests/data/119.lava_end_scene.level1.outputs.json
```diff
@@ -0,0 +1,100 @@
...
+        "lava": null,
+        "objects_count": 0,
+        "position_x": null,
+        "position_z": null,
+        "return_status": "SUCCESSFUL",
+        "reward": -200.006,
+        "rotation_y": null,
+        "step_number": 6,
+        "structural_objects_count": 0
+    }
```
ThomasSchellenbergNextCentury, line 99: You can delete this last block from each of the `outputs.json` files for test 119 since it never reaches this action/step.

kdrumm (approved):
Should probably make thomas's changes, but otherwise looks great to me.

ThomasSchellenbergNextCentury (approved):

PR #548, opened 2022-03-23
New action for interacting with agent

jaudick:
[Unity](https://github.com/NextCenturyCorporation/ai2thor/pull/228)
[Private](https://github.com/NextCenturyCorporation/mcs-private/pull/176)


kdrumm:
Test 123 is failing for me.  I won't rule out addressables not working right, but I've rebuilt them a couple times.

jaudick:
@kdrumm  I forgot to update mcs-private with the dwarf people, I made a new branch https://github.com/NextCenturyCorporation/mcs-private/pull/176

kdrumm:
> @kdrumm I forgot to update mcs-private with the dwarf people, I made a new branch [NextCenturyCorporation/mcs-private#176](https://github.com/NextCenturyCorporation/mcs-private/pull/176)

That was the trick!  Thanks.

kdrumm (approved):

machine_common_sense/action.py
```diff
@@ -578,6 +578,50 @@ class Action(Enum):
...
+        "InteractWithAgent",
+        "T",
+        "Interact with an Agent to make them hold out their "
+        "currently held object. "
+        "(objectId=string, "
+        "objectImageCoordsX=float, "
+        "objectImageCoordsY=float)"
+    )
+    """
+    Interact with an Agent to make them hold out their currently held object.
```
ThomasSchellenbergNextCentury, line 591: Can we add a little more detail here? Maybe something like this: `Interact with an agent. If that agent has an object, it will hold out the object for you to pickup; otherwise, the agent will look sad.` (I'm open to feedback)

integration_tests/data/123.agents_interaction_action.scene.json
```diff
@@ -0,0 +1,123 @@
+﻿{
+    "name": "agent_interaction_action",
```
ThomasSchellenbergNextCentury, line 2: Can we please add a few more integration test cases (either in this test, or as additional tests):
- Trying to Interact with an agent that's too far away (should fail)
- Trying to Interact with an agent that's behind a wall (should fail)
- Trying to Interact with an agent while it's already trying to give you the soccer ball (I'm assuming it should fail? The animation probably shouldn't restart. Maybe we need a new action status to return in this case?)
- Trying to Interact with an agent while it's performing some other movement-intensive animation, like waving or jumping (I'm assuming it should succeed, but make sure the transition doesn't look too weird)
- Trying to Interact with an agent while it's performing an animation on a loop (I think if the agent permanently halts its animation loop, that's probably OK, because the robot got what it needed out of the agent? It's OK if the agent returns to its animation loop as well -- whatever's easiest)
- Trying to Interact with a non-agent object (should fail)

machine_common_sense/object_metadata.py
```diff
@@ -58,6 +58,8 @@ class ObjectMetadata(object):
         Whether the object can be opened
     locked: boolean
         Whether the object is locked
+    associate_with_agent: str
+        The agent holding this object.
```
ThomasSchellenbergNextCentury, line 62: If possible, please try to keep this property list in alphabetical order.

machine_common_sense/object_metadata.py
```diff
@@ -80,7 +82,10 @@ def __init__(
         visible=False,
         is_open=False,
         openable=False,
-        locked=False
+        locked=False,
+        associated_with_agent="",
+        simulation_agent_held_object="",
+        simulation_agent_is_holding_held_object=False
```
ThomasSchellenbergNextCentury, line 88: Please add these properties to this class's docstring, with types and descriptions.

machine_common_sense/action.py
jaudick: I think that sounds great

machine_common_sense/action.py
```diff
@@ -578,6 +578,52 @@ class Action(Enum):
...
+        interactable object.
+    "NOT_OBJECT"
+        If the object corresponding to the "objectId" (or object corresponding
+        to the "objectImageCoords" vector) is not an object.
+    "OBSTRUCTED"
+        If you cannot move the object because your path is obstructed.
+    "OUT_OF_REACH"
+        If you cannot move the object because you are out of reach.
+    "FAILED"
+        Unexpected error; please report immediately to development team.
```
ThomasSchellenbergNextCentury, line 624: Please add the new possible return states here.

integration_tests/data/125.agents_interaction_advanced_action.scene.json
```diff
@@ -0,0 +1,137 @@
+﻿{
+    "name": "agent_interaction_action",
```
ThomasSchellenbergNextCentury, line 2: Currently tests 123, 125, and 126 all have the same scene names. (I use scene names to provide more details on what the test is testing, since I don't want to make the filenames too long!)

docs/source/schema.rst
```diff
@@ -169,7 +170,9 @@ Action Config
 Each **action config** has the following properties:
 
 - `stepBegin` (integer, required): The step on which the action should occur.  Must be non-negative.  A value of `0` means the action will occur during scene initialization.
+- `stepEnd` (integer, optional): The step on which the action should end.  Must be non-negative.
```
ThomasSchellenbergNextCentury, line 173: The default for this property is that the animation never ends, right?

kdrumm: Some animations appear to end after a certain amount of steps.  TPE_clap is one of them.

ThomasSchellenbergNextCentury: We should add a note here; something like: "By default, some animations will play indefinitely, while others will end on their own."

ThomasSchellenbergNextCentury: Could we just say something to that effect in the docs here?

ThomasSchellenbergNextCentury (approved):

PR #551, opened 2022-03-31
changed cache addressables to use the auto-downloader when build path isn't provided

kdrumm:
Jed wanted a script to run to download the proper unity build and then run cache addressables.  I figured this should be easy.  Additionally, maybe this is what we should have been doing, but never really thought about since cache_addressables was created prior to the auto downloader.

This was easy enough to just do the work and debate whether it is warranted in the PR.  We can easily delete if we don't want to do this.

Reference to conversation in with Jed:
https://machinecommonsense.slack.com/archives/CTR31V12M/p1648766818248429?thread_ts=1648692973.586849&cid=CTR31V12M

### DIFF
--- a/machine_common_sense/scripts/cache_addressables.py
+++ b/machine_common_sense/scripts/cache_addressables.py
@@ -1,6 +1,9 @@
 import argparse
 import os
 
+from machine_common_sense.unity_executable_provider import \
+    UnityExecutableProvider
+
 '''
 Script will start Unity build with a flag to clear the addressables cache,
 cache all the data in catalog and then close the application.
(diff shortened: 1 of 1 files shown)


machine_common_sense/scripts/cache_addressables.py
```diff
@@ -20,8 +23,10 @@
 
 def parse_args():
     parser = argparse.ArgumentParser()
-    parser.add_argument('build_path',
-                        help="Unity build path")
+    parser.add_argument('--build_path',
```
kdrumm, line 26: My opinion is the only downside to this is that we slightly changed the arguments of this script by forcing the addition of the '--'.  Otherwise the parameter is forced to be required.

brianpippin (approved):

ThomasSchellenbergNextCentury (approved):

PR #558, opened 2022-04-07
Bump to new version

brianpippin:


brianpippin:
This for 0.5.4 in advance - as I forgot to do 0.5.3 after the last release.

jaudick (approved):

rartiss55 (approved):

kdrumm (commented):
Didn't we start bumping the version last release and it broke something so we held back?

brianpippin:
> Didn't we start bumping the version last release and it broke something so we held back?

I believe it was something with the auto downloader, because 0.5.4 isn't out yet.  I am okay with this ticket sitting until we do the release.  I just think it's good we create after we do the prior release so we don't forget about it.

PR #561, opened 2022-04-12
MCS-1295 Added partitionFloor option for MCS scene files. Changed lava property format. (Python)

ThomasSchellenbergNextCentury:
Unity PR: https://github.com/NextCenturyCorporation/ai2thor/pull/234

Summary (copied from Unity PR):

A few months ago, when I was playing around with the moving target scenes, I encountered an issue in which the ball might bounce up a little when rolling at a fast speed across a supposedly smooth floor due to the way we're segmenting the floor in scenes containing lava. For reference, here's the Slack thread and a PR we merged with fixes:
- https://machinecommonsense.slack.com/archives/G01JFPF15R9/p1643925036628799
- https://github.com/NextCenturyCorporation/ai2thor/pull/199

While this worked in my initial testing, unfortunately, when you increase the room size, you need to increase the speed as well, and this issue reappears. It even happens if you eliminate all the `bounciness` from the ball!

Rather than looking for more workarounds, I've implemented a new scene file option to "partition" the room into three sections: left lava, center not-lava, and right lava. The width of the left and right sides can be configured, and the center section uses the "floorMaterial" as normal. This change eliminates the bouncing issue, and I believe I've addressed all of the side-effects:
- [Unity] For setting the "on_lava" status in the output metadata, the check that we're doing (seeing if the material of the floor underneath is a "lava" material) already works for this case
- [Python] I've updated the `lava` property in the output metadata to calculate and return the left and right sides. (The `lava` property now returns bounding boxes rather than points.)
- [Python] I've updated the top-down plotter code to handle the new format of the `lava` property.
- [Python] I've added integration tests to ensure that stepping on a side with lava decreases the reward as expected.

I'd appreciate hearing your feedback on this approach: if there are any important ramifications I've missed, or you're concerned about the change in the `lava` format, or you have any other ideas or suggestions for solving the issue, etc. Thanks!

Here's an example scene: https://github.com/NextCenturyCorporation/MCS/blob/MCS-1295-partition-floor/integration_tests/data/143.partitioned_floor.scene.json


machine_common_sense/config_manager.py
```diff
@@ -606,6 +612,25 @@ def retrieve_goal(self, steps_allowed_in_lava=0):
             )
         )
 
+    def retrieve_lava(self) -> List[Tuple[float, float, float, float]]:
+        """Return the list of lava locations as (X1, Z1, X2, Z2) tuples, where
+        X1/Z1 is the top-left corner and X2/Z2 is the bottom-left corner."""
```
jaudick, line 617: Should this be bottom right corner?

jaudick (approved):

kdrumm (approved):

brianpippin (approved):

PR #565, opened 2022-04-25
MCS-1302 adding focal length to step metadata and adding some tests

kdrumm:
[Unity](https://github.com/NextCenturyCorporation/ai2thor/pull/239)

### DIFF
--- a/integration_tests/data/001.empty_room_movement.oracle.outputs.json
+++ b/integration_tests/data/001.empty_room_movement.oracle.outputs.json
@@ -1,6 +1,7 @@
 ﻿[
     {
         "camera_height": 0.762,
+        "focal_length": 50.0,
         "head_tilt": 0,
         "objects_count": 0,
         "position_x": 0.0,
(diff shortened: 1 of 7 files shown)


brianpippin (approved):

jaudick (approved):

integration_tests/data/001.empty_room_movement.oracle.outputs.json
```diff
@@ -1,6 +1,7 @@
 ﻿[
     {
         "camera_height": 0.762,
+        "focal_length": 50.0,
```
ThomasSchellenbergNextCentury, line 4: Do we know why this is returning `50`? When I open a scene in the Unity Editor, look at the Camera in the FirstPersonCharacter game object, enable the Physical Camera settings, and adjust the Field of View back to its current setting (`42.5`), the Focal Length is shown in the Inspector as `30.85795` ... Now I could certainly be testing this incorrectly, but I want to make sure that we're not accidentally modifying the current Field of View, simply by accessing the `focalLength` setting in the camera... (I'm also a little confused because the [Unity docs](https://docs.unity3d.com/ScriptReference/Camera-focalLength.html) say this property isn't accessible unless `UsePhysicalProperties` is enabled, which is isn't in our code?)

PR #566, opened 2022-04-26
MCS-1306:  Do not allow default image coordinates when none passed with no object ID

brianpippin:


machine_common_sense/parameter.py
```diff
@@ -321,36 +370,20 @@ def _validate_and_convert_params(self, **kwargs) -> Tuple[Action, Dict]:
...
-        receptacle_vector = {
-            'x': receptable_image_coords_x,
-            'y': self._convert_y_image_coord_for_unity(
-                receptacle_image_coords_y)
-        }
+        object_vector = self._get_object_vector(**kwargs) if (
+            action in OBJECT_IMAGE_ACTIONS) else self.DEFAULT_IMG_DICT
+
+        receptacle_vector = self._get_receptacle_vector(**kwargs) if (
+            action in RECEPTACLE_ACTIONS) else self.DEFAULT_IMG_DICT
```
ThomasSchellenbergNextCentury, line 385: We should add a comment here about why the default is necessary (to prevent Unity from raising errors). I remember you talking about it at standup this morning, but I might not remember a few months from now :)

ThomasSchellenbergNextCentury (approved):

PR #567, opened 2022-04-26
MCS-1313 Trying to make the robot and the goal object more identifiable in the top-down plots

ThomasSchellenbergNextCentury:
I felt inspired to tackle this ticket today. Changes include:
- Thickened the border around the goal object and changed its color from gold to the opposite color of the object. (Technically I'm drawing a smaller polygon on top of a correctly-sized polygon.) For white/grey/black objects, the border is lime green.
- Added a border of the opposite color (cyan) around the robot. (Technically I'm drawing a smaller circle on top of the normal-size circle.) Since its "nose" is drawn over top of the border, it looks a little weird, but I think it would look weirder if the nose was underneath the border instead... Maybe if we changed its colors? I'm open to suggestions.
- Added a few new unit tests for drawing angled objects.


brianpippin:
The code looks good to me, but the border almost looks too thick in a lot of the examples I saw.  Will see what everyone else thinks, but that was my first impression.

ThomasSchellenbergNextCentury:
> The code looks good to me, but the border almost looks too thick in a lot of the examples I saw. Will see what everyone else thinks, but that was my first impression.

@brianpippin That's valid. It's an easy thing to change, if we want to.

ThomasSchellenbergNextCentury:
I just pushed an update for the color and shape of the agent in the top-down view:
- Changed the agent's colors from red/cyan to blue/yellow
- Changed the agent's shape from a circle with a circular "nose" to a triangle

Still open to feedback. Please let me know if you have any concerns about the changes!

brianpippin:
The code looks good and the color looks good.  I still think the border would look a little better somewhere between the old and the new border, but if everyone else is fine with it, I could be convinced to stick with the thicker border.

ThomasSchellenbergNextCentury:
It would be nice (but not strictly necessary) to get this into today's release.

kdrumm (approved):

jaudick (approved):

PR #577, opened 2022-05-17
Mcs 1332 put object crash

kdrumm:
I want to get @cedorman 's feedback to make sure it fixes all of the issues before opening this up for general review.  

I was unable to replicate the missing history on crash, but this may prevent a bunch of the crashes.  I would like to reproduce that issue and make sure that is fixed independently as well if I can.


brianpippin:
The code looks good.  Will wait for @cedorman to chime in though.

cedorman (approved):
Looks good.  Ran and verified that it worked.  thanks.

ThomasSchellenbergNextCentury (approved):

PR #578, opened 2022-05-17
MCS-1330 - Update integration tests with floor/hole changes made for segmentation masks

rartiss55:
Unity PR here: https://github.com/NextCenturyCorporation/ai2thor/pull/247


integration_tests/data/143.partitioned_floor.oracle.outputs.json
```diff
@@ -27,10 +27,10 @@
         "step_number": 0,
         "structural_objects": [
             {
-                "direction_x": -0.5,
-                "direction_y": -0.86,
+                "direction_x": -0.94,
```
rartiss55, line 30: Was initially worried about this, but talked it over with Thomas and it sounds like it makes sense that these numbers would change, since the floor sections were flattened/resized and therefore have a new center.

ThomasSchellenbergNextCentury (changes requested):
So when I ran all the integration tests, everything passed except for test 107:

```
FAILED INTEGRATION TESTS:
    107.lava (level2): Step 5 failed:
        haptic_feedback: {'on_lava': False} != {'on_lava': True}
        reward: -0.005 != -100.005
    107.lava (oracle): Step 5 failed:
        haptic_feedback: {'on_lava': False} != {'on_lava': True}
        reward: -0.005 != -100.005
```

But when I ran with `--test 107`, Level 2 passed, and only Oracle failed. And when I ran again with `--test 107 --metadata oracle`, it passed! @rartiss55 @jaudick Is this one of those weird edge cases with the lava you had mentioned?

I see that test 107 starts the robot at Z=0, so at its 5th step moving forward, the robot's position (Z=0.5) is exactly on the line between the floor and the lava. Should we just change the test itself so it's less of a weird case (maybe start the robot at Z=0.01)?

rartiss55:
> So when I ran all the integration tests, everything passed except for test 107:
> 
> ```
> FAILED INTEGRATION TESTS:
>     107.lava (level2): Step 5 failed:
>         haptic_feedback: {'on_lava': False} != {'on_lava': True}
>         reward: -0.005 != -100.005
>     107.lava (oracle): Step 5 failed:
>         haptic_feedback: {'on_lava': False} != {'on_lava': True}
>         reward: -0.005 != -100.005
> ```
> 
> But when I ran with `--test 107`, Level 2 passed, and only Oracle failed. And when I ran again with `--test 107 --metadata oracle`, it passed! @rartiss55 @jaudick Is this one of those weird edge cases with the lava you had mentioned?
> 
> I see that test 107 starts the robot at Z=0, so at its 5th step moving forward, the robot's position (Z=0.5) is exactly on the line between the floor and the lava. Should we just change the test itself so it's less of a weird case (maybe start the robot at Z=0.01)?

Weird! Wondering if in this case this is actually a python side/rounding issue with movement/positioning, since I tried to get this to fail in the Unity editor and the lava check worked fine same scene. Either way, I did what you suggested and it seems like it works fine now.

ThomasSchellenbergNextCentury:
I'll rerun the tests as well. I don't think this issue is on the python side, since we're just utilizing the "stepsOnLava" integer sent from Unity. Unity's internals don't appear as deterministic as I assumed when I started on MCS, so I'm happy to blame this inconsistency on the inner workings of SphereCast. Regardless, I have no problem with this solution (of changing the performer start position).

rartiss55:
@ThomasSchellenbergNextCentury Sounds good. Tests finished and passed for me by the way, hopefully its all good on your end as well.

PR #579, opened 2022-05-17
MCS-1339 Added start data to scene history file

kdrumm:
Adds data from step 0 (when the action is 'initialize' and no real commands have been sent)

### DIFF
--- a/machine_common_sense/history_writer.py
+++ b/machine_common_sense/history_writer.py
@@ -46,6 +46,15 @@ class HistoryEventHandler(AbstractControllerSubscriber):
             self.__history_writer = HistoryWriter(payload.scene_config,
                                                   hist_info,
                                                   payload.timestamp)
+            init_history = SceneHistory(
+                step=payload.step_number,
+                action='Initialize',
+                args={},
+                # We don't have the equivalent of params at this point
+                params=None,
+                output=payload.step_output.copy_without_depth_or_images(),
+                delta_time_millis=0)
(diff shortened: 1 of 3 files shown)


machine_common_sense/history_writer.py
```diff
@@ -46,6 +46,15 @@ def on_start_scene(self, payload: StartScenePayload):
             self.__history_writer = HistoryWriter(payload.scene_config,
                                                   hist_info,
                                                   payload.timestamp)
+            init_history = SceneHistory(
+                step=payload.step_number,
+                action='Initialize',
+                args={},
+                # We don't have the equivalent of params at this point
+                params=None,
```
ThomasSchellenbergNextCentury, line 54: We'll want to make sure that these empty/null properties don't cause an issue during ingest as part of the next ticket: https://nextcentury.atlassian.net/browse/MCS-1340

ThomasSchellenbergNextCentury (approved):

machine_common_sense/history_writer.py
brianpippin: Shouldn't be an issue, we don't use params now, and see no need to include it if it is none:
new_step = {
        'stepNumber': step["step"],
        'action': step["action"],
        'args': step["args"],
        'classification': step["classification"],
        'confidence': step["confidence"],
        'internal_state': step["internal_state"],
    }
    
    build_new_step_obj line 420

brianpippin (approved):

PR #580, opened 2022-05-19
MCS-1342: Add 'targetIsVisibleAtStart' to first object in history

brianpippin:
All, I have some more integration testing to do with the Jacobs branch, but wanted to get this in front of everyone.


kdrumm (approved):
On Step 0, we write null to the history file.  I think removing that would be a nice to have.  There are comments that indicate it might be possible in SceneHistory.__iter__(), but I'm not sure how to do it.

ThomasSchellenbergNextCentury (approved):

PR #588, opened 2022-06-10
Mcs 1371 timeout improvement

cedorman:
Slightly more complicated way to do a timeout, but it allows the opportunity to check and see what is going on and continue if needed.  The check is MCS-1372.


ThomasSchellenbergNextCentury (commented):
This looks fine -- unit tests pass -- but I'm not sure what's the goal here? I'm assuming the `concurrent` library does something useful for us?

brianpippin:
I am not sure of what the point of this fix as well.  I think we should table this until after our performance discussions with TA1.  Ideally our addressables improvements address this timeout issue.

brianpippin:
@ThomasSchellenbergNextCentury @rartiss55 - Thoughts on this ticket? Do we close and move on, or do we think this is worth adding?

ThomasSchellenbergNextCentury:
I'm fine with merging this. I trust that Clark believed it would be useful.

PR #593, opened 2022-08-05
Rotating cog integration tests

jaudick:
[Unity](https://github.com/NextCenturyCorporation/ai2thor/pull/263)


rartiss55 (approved):

ThomasSchellenbergNextCentury (commented):
Looks good to me, though I'd suggest adding the following as well (either as separate integration tests, or in the existing tests):
- A bucket (use one of the "bowl_" or "cup_" shapes) that contains a soccer ball sitting on top of the rotating cog
- A Toon People agent standing on top of the rotation cog

Both of these are specific scenarios we're discussing for Eval 6 tasks.

brianpippin:
I would agree with adding the tests Thomas suggested, especially since we are going to be making mock ups for these tasks in the near future.

jaudick:
@brianpippin @ThomasSchellenbergNextCentury I noticed that the performer could not walk on the cog so I fixed that. Check out the new commit on the unity branch for that. I also decreased the cog receptacle trigger box size because when the cog was scaled large it was rotating things that were quite far away. The three new tests cover:
1. Walk up to agent on cog, spin around it,, interact and wait for the agent to rotate to you then pickup
2. Pickup soccer ball in bowl
3. Start off of the cog, walk on it and walk across it to open a chest and pick up the soccer ball, then walk off the cog

ThomasSchellenbergNextCentury (approved):
The robot being able to walk on the cog wasn't strictly necessary, but it's nice to have anyway! Everything here looks good.

PR #594, opened 2022-08-15
MCS-1433 Added resetCenterOfMassAtY object schema property.

ThomasSchellenbergNextCentury:
The `resetCenterOfMass` property was added to address visual behavior when the object hit the ground that the TA2 psychs deemed undesirable. Its intent was to reset the implausible center of mass (back to the object's normal center of mass) when the object first gained an upward velocity due to hitting and bouncing off the floor. The issue exposed during Eval 5 was that sometimes affected objects would bounce when initially placed on the visible support, causing their center of mass to be reset too soon. Therefore I've now added a `resetCenterOfMassAtY` property, replacing the previous behavior, which only resets the center of mass when the object falls past a specified Y position. As a fallback, in case this approach doesn't work for whatever reason, we can remove `resetCenterOfMass` entirely, though then we'd return to having the undesired visual behavior. I think this PR represents a good middle-ground.

Unity PR: https://github.com/NextCenturyCorporation/ai2thor/pull/264
Hypercube Scene Generator PR: https://github.com/NextCenturyCorporation/mcs-private/pull/278

### DIFF
--- a/machine_common_sense/config_manager.py
+++ b/machine_common_sense/config_manager.py
@@ -277,6 +277,7 @@ class SceneObject(BaseModel):
     pickupable: Optional[bool]
     receptacle: Optional[bool]
     reset_center_of_mass: Optional[bool]
+    reset_center_of_mass_at_y: Optional[float]
     resizes: List[SizeConfig] = None
     rotates: List[MoveConfig] = None
     salient_materials: List[str] = None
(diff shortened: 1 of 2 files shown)


brianpippin (approved):

jaudick:
Should we have one integration test for this?

rartiss55 (approved):

PR #596, opened 2022-08-19
MCS-1439 Allow only Pass actions in each passive scene in case it doesn't have an action_list

ThomasSchellenbergNextCentury:
Historically we've restricted passive scene actions (no moving, rotating, etc.) within the MCS python library using the action_list property in the JSON scenes. However, it's easy for someone to simply delete the action_list and run around in a passive scene if they really wanted.

Currently the ILE scene generator does not setup passive physics scenes exactly the same way as in the eval data. Specifically, it doesn't add the "intuitivePhysics": true setting to the JSON. This setting alters the room in certain ways, like removing the side walls and the ceiling. I've been hesitant to add support for this setting in the past because if someone messes with the action_list settings, they could start looking and walking around the room and see the unrealistic room configuration.

This PR adds an extra safeguard to prevent move and rotate actions being performed in passive scenes. This is not intended to replace the action_list setting, since it's used for more than just passive scenes anyway. This PR simply makes me feel more comfortable about implementing changes to ensure the training and evaluation passive physics scenes more closely align with one another, which I believe will be necessary moving forward given the feedback from OPICS. Happy to discuss further.

Old (closed) PR: https://github.com/NextCenturyCorporation/ai2thor/pull/266


machine_common_sense/controller.py
```diff
@@ -479,9 +483,6 @@ def get_metadata_level(self) -> str:
         """
         return self._config.get_metadata_tier().value
 
-    def retrieve_action_list_at_step(self, goal, step_number):
-        return goal.retrieve_action_list_at_step(step_number)
-
```
ThomasSchellenbergNextCentury, line 484: I didn't see this function being used anywhere (just the function of the same name in the GoalMetadata class), so I deleted it (and its corresponding test).

brianpippin (approved):

jaudick (approved):

PR #599, opened 2022-08-25
Default image coords for human input mode

jaudick:


machine_common_sense/scripts/run_human_input.py
```diff
@@ -134,20 +131,26 @@ def _is_parameters_valid(self, action_str, params):
...
                 'objectImageCoordsY' in params))
+            if not valid:
+                print(
+                    f"Action: '{action_str}' using default image coords "
+                    f"x=300 y=200 because no objectId "
+                    f"or objectImageCoords were given")
+                params['objectImageCoordsX'] = 300
+                params['objectImageCoordsY'] = 200
         elif action in RECEPTACLE_ACTIONS:
-            valid = 'objectId' in params
```
jaudick, line 138: This check seemed completely unnecessary here since we need to call PutObject with a receptacle image coord or receptacle id to put the held object in it.

machine_common_sense/scripts/run_human_input.py
```diff
@@ -134,20 +131,26 @@ def _is_parameters_valid(self, action_str, params):
             valid = ('objectId' in params or (
                 'objectImageCoordsX' in params and
                 'objectImageCoordsY' in params))
+            if not valid:
+                print(
+                    f"Action: '{action_str}' using default image coords "
+                    f"x=300 y=200 because no objectId "
+                    f"or objectImageCoords were given")
+                params['objectImageCoordsX'] = 300
+                params['objectImageCoordsY'] = 200
```
ThomasSchellenbergNextCentury, line 140: If they give the X but not the Y, should we use the one value they provide, and the default value for the other parameter? I'm indifferent.

rartiss55 (approved):

machine_common_sense/scripts/run_human_input.py
jaudick: @ThomasSchellenbergNextCentury I added the option to only have one specified image coord in my new commit

ThomasSchellenbergNextCentury (approved):

PR #601, opened 2022-09-07
MCS-1452: Add room dimensions to oracle step metadata 

rartiss55:
Unity PR here: https://github.com/NextCenturyCorporation/ai2thor/pull/270


machine_common_sense/controller_output_handler.py
```diff
@@ -324,6 +328,9 @@ def _get_step_metadata(self, goal, habituation_trial,
             resolved_receptacle=(
                 None if restrict_non_oracle
                 else self._scene_event.resolved_receptacle),
+            room_dimensions=(
+                None if restrict_non_oracle
+                else self._scene_event.room_dimensions),
```
ThomasSchellenbergNextCentury, line 333: The need to send the room dimensions back from Unity, instead of simply using them directly from the SceneConfiguration -- I'm assuming that's in case Unity adjusts the room dimensions (like in passive scenes), or uses a default?

rartiss55: Yes, I ran into the default case with the playroom scene since it doesn't have room dimensions specified. Your comment made me realize I should probably add a test for one of those cases (just added one for the "no room dimensions in scene config" case) + I fixed the numbers of those tests based on other open PRs so they shouldn't conflict now.

ThomasSchellenbergNextCentury (approved):

jaudick (approved):

PR #609, opened 2022-09-29
MCS-1045 Added timeout for long scenes

rartiss55:
Uses instances of threading.Timer to check if we've been at the same step for x seconds (default is two hours). If so, end scene is called + an exception is thrown on the main thread. Added timeout value to the mcs config in case we need to change the length of time later.

### DIFF
--- a/machine_common_sense/config_manager.py
+++ b/machine_common_sense/config_manager.py
@@ -513,6 +513,8 @@ class ConfigManager:
         )
 
     def get_timeout(self):
+        """ Time (in seconds) to allow a run to be idle
+        before attempting to end scene"""
         return self._config.getint(
             self.CONFIG_DEFAULT_SECTION,
             self.CONFIG_TIMEOUT,
--- a/machine_common_sense/config_manager.py
+++ b/machine_common_sense/config_manager.py
@@ -318,6 +318,7 @@ class ConfigManager:
     CONFIG_GOAL_REWARD = 'goal_reward'
     CONFIG_TOP_DOWN_PLOTTER = 'top_down_plotter'
     CONFIG_TOP_DOWN_CAMERA = 'top_down_camera'
+    CONFIG_TIMEOUT = 'timeout'
 
     # Please keep the aspect ratio as 3:2 because the IntPhys scenes are built
     # on this assumption.
@@ -326,6 +327,10 @@ class ConfigManager:
 
     # Default steps allowed in lava before calling end scene
     STEPS_ALLOWED_IN_LAVA_DEFAULT = 0
+
+    # Default time to allow on a single step before timing out
+    # is 2 hours (represented in seconds)
+    TIMEOUT_DEFAULT = 7200
 
     def __init__(self, config_file_or_dict=None):
         '''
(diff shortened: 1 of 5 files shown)


machine_common_sense/controller.py
```diff
@@ -137,6 +144,34 @@ def _set_config(self, config: ConfigManager):
...
+        start_time = time.time()
+        self._timer_in_progress = False
+        timeout_secs = self._config.get_timeout()
+
+        if(self._last_step_check < self.__step_number):
+            # skip step check for Initialize step
+            if(self.__step_number != 0):
+                self._last_step_check = self.__step_number
+
+            timer_secs = timeout_secs - \
```
rartiss55, line 160: Seen this approach on line 160 used to avoid drifting over time -- not sure if that is an issue for us but figured it couldn't hurt.

machine_common_sense/controller.py
```diff
@@ -137,6 +144,34 @@ def _set_config(self, config: ConfigManager):
...
+        self._timer_in_progress = False
+        timeout_seconds = self._config.get_timeout()
+
+        if(self._last_step_check < self.__step_number):
+            # skip step check for Initialize step
+            if(self.__step_number != 0):
+                self._last_step_check = self.__step_number
+
+            timer_seconds = timeout_seconds - \
+                ((time.time() - start_time) % timeout_seconds)
```
rartiss55, line 161: Seen this approach on line 160 used to avoid drifting over time -- not sure if that is an issue for us but figured it couldn't hurt.

machine_common_sense/controller.py
```diff
@@ -137,6 +144,34 @@ def _set_config(self, config: ConfigManager):
...
+            self._timer = threading.Timer(timer_seconds,
+                                          self._check_step_for_timeout)
+            self._timer.start()
+            self._timer_in_progress = True
+        else:
+            time_str = str(datetime.timedelta(seconds=timeout_seconds))
+            logger.debug(
+                f"Attempting to end scene due to inactivity (user not taking"
+                f" any steps) for {time_str} (hh:mm:ss)")
+            self.end_scene(rating=None, score=-1)
```
rartiss55, line 171: I used the same values here that we use to handle exceptions -- not sure if we want something different to differentiate those situations from timeouts?

ThomasSchellenbergNextCentury: No opinion.

machine_common_sense/config_manager.py
```diff
@@ -327,6 +328,10 @@ class ConfigManager:
     # Default steps allowed in lava before calling end scene
     STEPS_ALLOWED_IN_LAVA_DEFAULT = 0
 
+    # Default time to allow on a single step before timing out
+    # is 2 hours (represented in seconds)
+    TIMEOUT_DEFAULT = 7200
```
ThomasSchellenbergNextCentury, line 333: Maybe this is a stupid question, but does this seem excessively long? Could we just do 1 hour? Shouldn't that be plenty?

brianpippin: I honestly think 1 hour is long enough to wait, it also might make sense to make this a config option so we can tweak it on the fly if we wanted?

rartiss55: As mentioned during standup, updated to 1 hour but we can update it later per run/performer if needed.

requirements.txt
```diff
@@ -5,7 +5,7 @@ commonmark==0.9.1
 flake8==3.9.1
 flake8-implicit-str-concat==0.2.0
 flake8-use-fstring==1.3
-importlib-metadata==4.13.0 python_version<"3.8"
+importlib-metadata==4.13.0; python_version<"3.8"
```
rartiss55, line 8: Heads up -- this may need to be python_version<="3.8", but the github actions for branches only run on 3.7 (won't know for sure until the merge).

ThomasSchellenbergNextCentury (approved):

jaudick (approved):

PR #611, opened 2022-10-05
MCS-895 Initial version of handling scenes with multiple targets.

ThomasSchellenbergNextCentury:
My initial version of the implementation. Happy to have a team discussion and adjust this approach if we decide it's not ideal.

1. There's now a goal `category` called "multi retrieval", which is distinct from normal "retrieval". This seemed better to me than adding a new/separate boolean property.
2. The `metadata` in the "multi retrieval" goals will have a `targets` property, that's a list of target dicts, rather than a single `target` dict. Like the single `target` dict, each dict will have an `id` corresponding to the target's "id". This seemed better than trying to reuse `target`, but making it sometimes a dict, and sometimes a list.
3. Added code to make sure the `targets` are handled properly, including clearing them when not running at oracle metadata level.
4. As discussed this morning, no reward is given until ALL targets are picked up.

EDIT -- ILE Scene Generator PR: https://github.com/NextCenturyCorporation/mcs-private/pull/299

### DIFF
--- a/machine_common_sense/controller_output_handler.py
+++ b/machine_common_sense/controller_output_handler.py
@@ -387,15 +387,24 @@ class ControllerOutputHandler():
         step_output.lava = None
         step_output.room_dimensions = None
 
-        target_name_list = ['target', 'target_1', 'target_2']
-        for target_name in target_name_list:
-            if (target_name in step_output.goal.metadata):
-                step_output.goal = copy.deepcopy(
-                    step_output.goal)
-                if 'image' in step_output.goal.metadata[target_name]:
-                    step_output.goal.metadata[target_name]['image'] = None
-                if 'id' in step_output.goal.metadata[target_name]:
-                    step_output.goal.metadata[target_name]['id'] = None
-                if 'image_name' in step_output.goal.metadata[target_name]:
-                    step_output.goal.metadata[
-                        target_name]['image_name'] = None
+        # Copy the goal object to avoid deleting data from the original object
+        step_output.goal = copy.deepcopy(step_output.goal)
+
+        metadata = step_output.goal.metadata or {}
+        # Different goal categories may use different property names
+        target_names = ['target', 'targets', 'target_1', 'target_2']
+        for target_name in target_names:
+            target = metadata.get(target_name)
+            # Some properties may be dicts, and some may be lists of dicts
+            if isinstance(target, list):
+                # Clear the whole array, to avoid revealing the total count
+                metadata[target_name] = []
+            elif target:
+                # Pretty sure the target always has an id, but just in case...
+                if 'id' in target:
+                    del target['id']
+                # Backwards compatibility: target used to have image data
+                if 'image' in target:
+                    del target['image']
+                if 'image_name' in target:
+                    del target['image_name']
--- a/machine_common_sense/goal_metadata.py
+++ b/machine_common_sense/goal_metadata.py
@@ -207,7 +207,8 @@ class GoalCategory(Enum):
     behaviors, and their interactions with objects in the environment.
 
     This goal category is only used for the **passive/VoE agent tasks**. All
-    interactive agent tasks will use the `retrieval` goal category.
+    interactive agent tasks will use either the `retrieval` or
+    `multi retrieval` goal category.
 
     Notes
     -----
@@ -239,7 +240,10 @@ class GoalCategory(Enum):
     RETRIEVAL = "retrieval"
     """
     In a trial that has a retrieval goal, you must find and pickup a target
-    object. This may involve exploring the scene, avoiding obstacles,
+    object. In MCS Evaluation 4 and onward, the target object will always be a
+    soccer ball (football).
+
+    This may involve exploring the scene, avoiding obstacles,
     interacting with objects (like closed containers) or agents, and tracking
     moving objects. These trials will demand a "common sense" understanding of
     self navigation (how to move and rotate yourself within a scene and around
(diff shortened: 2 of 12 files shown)


machine_common_sense/goal_metadata.py
```diff
@@ -239,22 +240,51 @@ class GoalCategory(Enum):
...
+    -----
+    At `oracle` metadata level, the `metadata` dict property of this
+    GoalMetadata object will contain a `targets` property, which is a list of
+    dicts that each contain the following parameters:
 
+    Parameters
+    ----------
+    id : string
+        The unique objectId of one of the target objects to retrieve.
     """
```
ThomasSchellenbergNextCentury, line 290: Information about the new "multi retrieval" goal category, which will be published to our API page:
https://nextcenturycorporation.github.io/MCS/api.html#machine_common_sense.GoalCategory

The 2nd paragraph I just copied from "retrieval".

machine_common_sense/controller_output_handler.py
```diff
@@ -387,15 +387,24 @@ def filter_step_output(self, step_output):
...
+                metadata[target_name] = []
+            elif target:
+                # Pretty sure the target always has an id, but just in case...
+                if 'id' in target:
+                    del target['id']
+                # Backwards compatibility: target used to have image data
+                if 'image' in target:
+                    del target['image']
+                if 'image_name' in target:
+                    del target['image_name']
```
ThomasSchellenbergNextCentury, line 410: The original code was setting these to None, but I thought deleting them would be better. Let me know if you have any concerns.

machine_common_sense/controller_output_handler.py
```diff
@@ -387,15 +387,24 @@ def filter_step_output(self, step_output):
...
+
+        metadata = step_output.goal.metadata or {}
+        # Different goal categories may use different property names
+        target_names = ['target', 'targets', 'target_1', 'target_2']
+        for target_name in target_names:
+            target = metadata.get(target_name)
+            # Some properties may be dicts, and some may be lists of dicts
+            if isinstance(target, list):
+                # Clear the whole array, to avoid revealing the total count
+                metadata[target_name] = []
```
ThomasSchellenbergNextCentury, line 401: Pointing out that I'm clearing this array at non-oracle metadata levels to avoid revealing the total count of targets.

machine_common_sense/history_writer.py
```diff
@@ -238,14 +232,22 @@ def add_step(self, step_obj: SceneHistory):
     def is_target_visible(
             self,
             history: SceneHistory) -> bool:
-        """Determine the visibility of the target object, if any"""
+        """Determine the current visibility of the target object, if any. In
+        scenes with multiple targets, returns true if any target is visible."""
```
ThomasSchellenbergNextCentury, line 236: This is used in the scorecard. I believe the intent of the metric was: "if they can see the target, they should move straight toward it and pick it up." Therefore it made sense (to me) that this should return `true` if ANY target was visible.

machine_common_sense/history_writer.py
```diff
@@ -238,14 +232,22 @@ def add_step(self, step_obj: SceneHistory):
...
+                targets = metadata.get(target_name) or []
+                targets = targets if isinstance(targets, list) else [targets]
+                for target in targets:
+                    for hist_obj in history.output.object_list:
+                        uuid = hist_obj.uuid
+                        if uuid == target.get('id') and hist_obj.visible:
+                            return True
             return False
         except Exception:
             return False
```
ThomasSchellenbergNextCentury, line 253: Not sure why this try/catch was here, but I didn't want to simply delete it, without knowing more first.

tests/test_controller_output_handler.py
```diff
@@ -638,13 +642,21 @@ def test_wrap_output_with_config_metadata_level2(self):
...
 
         self.assertEqual(actual.action_list, GoalMetadata.DEFAULT_ACTIONS)
         self.assertEqual(actual.camera_aspect_ratio, (600, 400))
         self.assertEqual(actual.camera_clipping_planes, (0, 150))
         self.assertEqual(actual.camera_field_of_view, 42.5)
         self.assertEqual(actual.camera_height, 0.1234)
+        self.assertEqual(actual.goal.metadata, {
+            'target': {},
+            'targets': [],
+        })
```
ThomasSchellenbergNextCentury, line 659: These tests weren't actually checking whether the target IDs were getting deleted at non-oracle metadata levels, so I added some in.

tests/test_plotter.py
```diff
@@ -488,15 +488,10 @@ def test_draw_holes_even_room_dimensions(self):
...
-        goal = {'metadata': {
-            'target': {'image': [0]},
-            'target_1': {'image': [1]},
-            'target_2': {'image': [2]}
-        }}
         scene_config = SceneConfiguration(
             name="testscene",
             version=1,
-            goal=goal,
+            goal={'metadata': {}},
```
ThomasSchellenbergNextCentury, line 494: These tests don't really need a goal object, so I simplified it.

machine_common_sense/reward.py
```diff
@@ -71,10 +71,23 @@ def _calc_retrieval_reward(
...
+            targets = targets if isinstance(targets, list) else [targets]
+            for target in targets:
+                goal_id = target.get('id')
+                goal_object = Reward.__get_object_from_list(objects, goal_id)
+                if goal_object:
+                    goal_objects.append(goal_object)
+
+        # Only attain the reward if all targets are picked up
+        if all([obj.get('isPickedUp', False) for obj in goal_objects]):
             reward = goal_reward
```
ThomasSchellenbergNextCentury, line 91: Reward function will be easy to adjust later if we get feedback about it from TA1.

rartiss55: Didn't know how to comment directly on it but I think line 174 also needs updating (the switch only looks for "retrieval" and not "multi retrieval")

machine_common_sense/history_writer.py
rartiss55: Hmm, not sure, but my guess would be to catch any "non existent key" errors and return false?

machine_common_sense/history_writer.py
rartiss55: So for the scorecard, this metric checks if one is heading toward the target, and if they consistently are moving away from the target, they get dinged for it after a certain point. It would need tweaking but I think this can still work for multiple, it would just have to be something like "moving consistently towards any target" and figure out which one they are moving towards. Wondering though if we should track which target(s) are visible here for the purpose of that calculation?

rartiss55: To elaborate, moving away from one target likely means we're moving away from another, so I wouldn't want to run into any issues if we calculated against the wrong target or something.

machine_common_sense/reward.py
ThomasSchellenbergNextCentury: Good catch!

machine_common_sense/history_writer.py
ThomasSchellenbergNextCentury: That was my thought as well, but using `.get()` should eliminate the need for it.

machine_common_sense/history_writer.py
ThomasSchellenbergNextCentury: That makes sense. Definitely something worth addressing now. I like the idea of tracking which targets are visible. Maybe `target_visible` could be a list of visible object ID strings, rather than a bool? Or we could keep it a bool for "retrieval" scenes, but make it a list for "multi retrieval" scenes?

rartiss55: I think either or is fine, although slightly leaning towards the list of visible strings for consistency.

Also might be nitpicky, but should it be `targets_visible` now?

ThomasSchellenbergNextCentury: I'm actually thinking now 1. keeping it `target_visible` and 2. keeping it a bool for "retrieval" scenes would be important for backwards compatibility.

brianpippin (approved):
The code looks good to me, and I think the scorecard for multi will need a different approach, one that will be different then a singular retrieval.

PR #616, opened 2022-10-14
Trigger placers by opening containers

jaudick:


integration_tests/data/172.triggered_by_incorrect_sequence.oracle.outputs.json
```diff
@@ -0,0 +1,6219 @@
...
+        "objects_count": 5,
+        "position_x": -0.4,
+        "position_z": 0.1,
+        "resolved_object": "chest_2",
+        "resolved_receptacle": "",
+        "return_status": "SUCCESSFUL",
+        "rotation_y": 90,
+        "step_number": 45,
+        "structural_objects_count": 6,
+        "triggered_by_sequence_incorrect": true
```
jaudick, line 6217: Note this, which would end the scene being set to true after the yellow chest is opened. Now the scene must end

machine_common_sense/goal_metadata.py
```diff
@@ -130,12 +130,14 @@ def retrieve_action_list_at_step(
         self,
         step_number: int,
         steps_in_lava: Optional[int] = 0,
+        triggered_by_sequence_incorrect: Optional[bool] = False,
```
ThomasSchellenbergNextCentury, line 133: We should add this property to the docs at the top of this file.

machine_common_sense/step_metadata.py
```diff
@@ -184,7 +186,8 @@ def __init__(
         segmentation_colors=None,
         step_number=0,
         steps_on_lava=0,
-        structural_object_list=None
+        structural_object_list=None,
+        triggered_by_sequence_incorrect=False
```
ThomasSchellenbergNextCentury, line 190: Since it's already in GoalMetadata, I don't think we need it in here as well?

machine_common_sense/goal_metadata.py
jaudick: I don't think this should go to the top of this file because its not something being set like  `steps_allowed_in_lava : integer The number of steps allowed in lava before the scene ends` which defines some condition in goal metadata.  Currently triggered by is attached to individual objects not the goal metadata.

machine_common_sense/step_metadata.py
jaudick: I think we need it here since its like steps_on_lava which is checked each step if the performer messed up the sequence

ThomasSchellenbergNextCentury: Then can we remove it from GoalMetadata?

machine_common_sense/goal_metadata.py
ThomasSchellenbergNextCentury: Marking this conversation as resolved since it's outdated.

machine_common_sense/step_metadata.py
ThomasSchellenbergNextCentury: Marking this conversation as resolved since it's outdated.

ThomasSchellenbergNextCentury (changes requested):
In addition to issues with the integration tests that I described above, I don't see us removing `triggered_by_target_sequence` from the GoalMetadata at non-oracle metadata levels. We definitely don't want to give TA1 the answers during the eval! The `filter_step_output` function in `controller_output_handler.py` would be a good place for that to go. Furthermore, adding some checks in the `test_wrap_step_*` functions in `test_controller_output_handler.py` might be a good idea, to ensure the correct properties are being returned (or removed) at each metadata level.

jaudick:
@brianpippin @ThomasSchellenbergNextCentury This is ready for review again

ThomasSchellenbergNextCentury (approved):

brianpippin (approved):

PR #617, opened 2022-10-17
MCS-1542: Fix case where an empty goal_objects would give the goal reward achieved

brianpippin:
If the goal_objects was an empty array and there was a goal id in the metadata object, it would say the reward was achieved, it appears that the only test this covered was 094.


ThomasSchellenbergNextCentury (approved):
Tests work. Glad it was a simple fix. Thanks Brian!

PR #628, opened 2022-11-04
MCS-1558 Updated integration tests due to corresponding bug fix.

ThomasSchellenbergNextCentury:
Unity PR: https://github.com/NextCenturyCorporation/ai2thor/pull/291

The area considered "on top" of the cog is now a bit wider, so I had to make a few minor changes:
- Updated integration test 152 to move the agent off of the cog
- Updated integration test 167 to Pass two more times at the start, since now it takes two fewer Move actions to get up on top of the cog


brianpippin (approved):

rartiss55:
I'm getting the following:

FAILED INTEGRATION TESTS:
    167.walk_on_rotating_cog_pickup_target_in_chest_then_walk_off (oracle): Step 49 failed:
        objects chest visible: False != True
    170.attachable_lid (oracle): Step 29 failed:
        objects container_1 visible: False != True

If no one else is getting this/its a build issue on my end, then disregard.

ThomasSchellenbergNextCentury:
@rartiss55 Thanks for finding this. I believe this was another case of the Unity random number generator running differently in isolation (with `--test <whatever>`) compared to the whole test suite at once.

The problematic checks I found concerned visibility in edge cases (e.g. does the sliver of the chest in the lower-right corner of this image mean that the chest is visible?), which I've had trouble with before. Since these particular tests aren't really concerned with checking these edge cases, I've simply removed the visibility checks on the appropriate steps. Hopefully if/when we figure out [MCS-1393](https://nextcentury.atlassian.net/browse/MCS-1393) then handling it this delicately will no longer be necessary.

frame_image_49

Please feel free to pull the latest and try again.

jaudick (approved):

PR #639, opened 2022-12-24
Mcs 1528 toggle metadata mcs config

phile-caci:
Added two config settings on the MCS config to toggle disable_object_list and disable_position.  They can only be disabled at oracle metadata level.  At level1,2  and none, they are always disabled.  (Hopefully I have that correct).


brianpippin:
The code looks good so far but I feel like we need additional configs.  I think is_depth_maps_enabled and is_object_masks_enabled needs an over ride config potential as well, so they could toggle one of them on or off, right now they both are returned and some teams might only want one.   Looking to see if there are other configs potentially missed.

rartiss55:
Left some comments, overall I think this makes sense + tested things out with some of the depth/object mask fields at different levels.

rartiss55 (changes requested):

tests/test_parameter.py
```diff
@@ -97,6 +97,8 @@ def test_wrap_step_action(self):
         expected = {
             "action": "TestAction",
```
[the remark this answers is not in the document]
phile-caci: Hi @rartiss55 ,

I think they're there already:

105            "renderDepthImage": False,
106            "renderObjectImage": False,

Let me know if that's not what you meant.
Thanks!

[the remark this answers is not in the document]
rartiss55: Whoops, they were hidden in the diff, I see them now, thanks!

machine_common_sense/parameter.py
```diff
@@ -100,15 +100,17 @@ def wrap_step(self, output_folder, **kwargs) -> Dict:
         consistent_colors = (metadata_tier == MetadataTier.ORACLE)
         # Create the step data dict for the AI2-THOR step function.
         return dict(
+            consistentColors=consistent_colors,
```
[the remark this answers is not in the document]
phile-caci: Should be here too (hidden in the diff like you mentioned already):

111            renderDepthImage=self.config.is_depth_maps_enabled(),
112            renderObjectImage=self.config.is_object_masks_enabled(),

docs/source/install.rst
```diff
@@ -121,6 +121,34 @@ To use a specific configuration, you can either pass in a file path or dictionar
...
+(boolean, optional)
+
+If `true`, does not generate position information output in the metadata. Metadata Tier will override `disable_object_list`. Cannot be false for Metadata tier [none, level1, level2]. Default: false
+
+enable_depth_maps
+^^^^^^^^^^^^^^^
+
+(boolean, optional)
+
+If `true`, will generate depth maps. Metadata Tier will override `enable_depth_maps`. Will only generate dept maps for Metadata tier [level1, level2, oracle]. Default: true
```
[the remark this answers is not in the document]
phile-caci: Corrected, thanks!

sample_config.ini
```diff
@@ -1,2 +1,11 @@
...
\ No newline at end of file
+metadata = oracle
+
+#Metadata includes object list. Cannot be false for Metadata tier [none, level1, level2]. Default: false
+disable_object_list = false
+#Metadata includes position information. Cannot be false for Metadata tier [none, level1, level2]. Default: false
+disable_position = false
+#Creates depth maps in output. Will only generate dept maps for Metadata tier [level1, level2, oracle]. Default: true
+enable_depth_maps = true
+#Creates object masks in output. Will only generate object masks for Metadata tier [level2, oracle]. Default: true1
```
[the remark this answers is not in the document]
phile-caci: Corrected, thanks!

ThomasSchellenbergNextCentury (changes requested):
@phile-caci @brianpippin When `disable_object_list` is `true` (including at non-oracle metadata levels, since it's `true` by default there), the performer can never achieve a positive reward, since the reward calculation requires the object list (in `controller_output_handler.py`, see `_get_step_metadata`, where it calls `calculate_reward`). This doesn't seem like it would work, since you'd never be able to succeed in any retrieval scenes.

ThomasSchellenbergNextCentury:
> > @phile-caci @brianpippin When `disable_object_list` is `true` (including at non-oracle metadata levels, since it's `true` by default there), the performer can never achieve a positive reward, since the reward calculation requires the object list (in `controller_output_handler.py`, see `_get_step_metadata`, where it calls `calculate_reward`). This doesn't seem like it would work, since you'd never be able to succeed in any retrieval scenes.
> 
> If that is the case perhaps this should be named to only_return_goal_object and that is the only object returned if we think this will improve performance at training time.

That seems reasonable to me. Even at non-oracle metadata levels, we still use the target object's output metadata within the MCS code for the reward calculation, even though we don't pass it back to the TA1 systems via the StepMetadata. I _believe_ this is the only place where we would need it. Unfortunately, this means that the MCS python code will need to pass information about all the target objects (probably their object IDs) to the Unity code (probably in `wrap_step`), so Unity knows which objects to process and which objects to skip. Shouldn't be too hard -- just another thing to do.

machine_common_sense/config_manager.py
```diff
@@ -312,6 +312,10 @@ class ConfigManager:
     DEFAULT_ROOM_DIMENSIONS = Vector3d(x=10, y=3, z=10)
     CONFIG_FILE_ENV_VAR = 'MCS_CONFIG_FILE_PATH'
     CONFIG_DEFAULT_SECTION = 'MCS'
+    CONFIG_ENABLE_DEPTH_MAPS = 'enable_depth_maps'
+    CONFIG_ENABLE_OBJECT_MASKS = 'enable_object_masks'
+    CONFIG_DISABLE_OBJECT_LIST = 'disable_object_list'
+    CONFIG_DISABLE_POSITION = 'disable_position'
```
ThomasSchellenbergNextCentury, line 318: In addition to your unit tests, it might be a good idea to add a few new integration tests for the `disable_position` and `disable_object_list` (or its replacement, as Brian and I have discussed) properties. You can write specific `config.ini` files for specific integration tests in order to test these properties -- see integration test 119 for examples. Given the hiccup with the rewards, we should have integration tests for both a normal "retrieval" scene (like test 073, though we use a soccer ball instead of a trophy these days) and a "multi retrieval" scene (like test 175).

rartiss55 (approved):

rartiss55:
I know Thomas had some additional concerns, but  just wanted to confirm my previous comments were addressed :)

phile-caci:
@ThomasSchellenbergNextCentury,  I didn't get too deep on the integration testing.  Just making sure we're good with the approach first.  

Replaced `disable_object_list` with `only_return_goal_object`. At oracle level will only return the goal objects.
Thanks!

integration_tests/config_oracle.ini
```diff
@@ -2,3 +2,4 @@
 history_enabled: false
 metadata: oracle
 steps_allowed_in_lava: 1000
+only_return_goal_object = true
```
ThomasSchellenbergNextCentury, line 5: We don't want to keep this here, right? Since it would affect ALL the integration tests?

integration_tests/data/182.multiple_targets_config_only_goal_objs.level2_only_goal_objects.outputs.json
```diff
@@ -0,0 +1,8 @@
+﻿[
+  {
+    "objects": []
+  },
+  {
+    "objects": []
+  }
```
ThomasSchellenbergNextCentury, line 7: I know you're not done with your integration tests yet, but I wanted to note that you'll need `"objects_count": X` here (where X is the number of target objects) in your `outputs.json` file for testing `only_return_goal_object = true`. This will ensure there are no extraneous objects being returned in the step metadata.

For some more detail: The integration tests don't do a 1-for-1 comparison of the `"objects"` in the `outputs` file and the objects in the step metadata. Instead, they iterate over the `"objects"` in the `outputs` file, find the corresponding object in the step metadata, and run the listed checks for that object. This is so you can have a scene with a bunch of objects in it, but maybe you only care about checking a subset of them. The same is true for the `structural_objects` (and `structural_objects_count`).

docs/source/install.rst
```diff
@@ -121,6 +121,34 @@ To use a specific configuration, you can either pass in a file path or dictionar
...
+(boolean, optional)
+
+If `true`, will generate depth maps. Metadata Tier will override `enable_depth_maps`. Will only generate depth maps for Metadata tier [level1, level2, oracle]. Default: true
+
+enable_object_masks
+^^^^^^^^^^^^^^^
+
+(boolean, optional)
+
+If `true`, will generate object masks. Metadata Tier will override `enable_depth_maps`. Will only generate object masks for Metadata tier [level2, oracle]. Default: true
```
ThomasSchellenbergNextCentury, line 150: It's not super important, but would these options be better as "disable_depth_maps" and "disable_object_masks"? Maybe it's just me, but I tend to think of config options as "off" by default, and you set them to turn them "on", but these are currently the reverse.

ThomasSchellenbergNextCentury:
@phile-caci This looks much better, though I'd appreciate having a few good integration tests before signing off (I know you said you're planning to work on them). Let me know if you'd like any help with them.

integration_tests/config_oracle.ini
phile-caci: I did rerun all of the integration test at oracle and they looked like they all passed, but agree probably shouldn't keep this here.  There's doesn't seem to be a built in mechanism now to handle a config parameter?  I can add another argument to accept config parameters and values if that makes sense?

ThomasSchellenbergNextCentury: Hmm, I'm definitely seeing errors at the oracle metadata level (for example, test 005 is failing on step 0 with "objects_count: 0 != 1").

You can create a config file for a specific test, naming it `<test_name>.<metadata_level>.config.ini` and putting it in the `integration_tests/data/` folder. See test 119 for an example.

phile-caci: Ah.. thanks!

docs/source/install.rst
phile-caci: I can make the change if you like, I only went that route because it looked like the code was mostly setup for it being enabled by default.

ThomasSchellenbergNextCentury: I didn't mean for you to change the existing code. I just meant that the options could be "disable_whatever", and "false" by default, which would mean they're actually enabled by default. Just a semantics thing.

phile-caci: Ahh got it. I'll make the change

phile-caci:
@ThomasSchellenbergNextCentury , 

I think I hit all the updates, let me know if I missed anything, thanks!

ThomasSchellenbergNextCentury (commented):
It would be nice if the new integration test for `only_return_goal_object: true` actually had the targets being picked up, and verified that the "reward" changes, but I won't let that hold up my approval, since I know were' on a deadline.

phile-caci:
> It would be nice if the new integration test for `only_return_goal_object: true` actually had the targets being picked up, and verified that the "reward" changes, but I won't let that hold up my approval, since I know were' on a deadline.

Added `only_return_goal_object` true integration test.  There are 12 objects in the scene, only 4 are goal target objects.  At oracle level, should only return 4 objects.

ThomasSchellenbergNextCentury:
I've run this through twice now, and I keep getting an error on integration test 171:

```
    171.triggered_by_correct_sequence (oracle): Step 37 failed:
        objects chest_2 visible: False != True
```

This only happens if I run the entire test suite (not the specific test).

If no one else is seeing this, then maybe something is funky with my local build. I don't want it to hold up approval. (And I suppose Jenkins will let us know if the tests fail after merging.)

integration_tests/data/186.only_return_goal_objects_true.actions.txt
```diff
@@ -0,0 +1 @@
+MoveAhead
```
ThomasSchellenbergNextCentury, line 1: @phile-caci I think you misunderstood my comment. I meant the test should involve walking around and picking up all of the soccer balls, and checking the "reward" in the "*.outputs.json" files.

Giving the timing, I'm fine if we just want to handle this in a separate ticket.

phile-caci: Ugh.. sry.. I'll go ahead and create the ticket.

phile-caci:
> I've run this through twice now, and I keep getting an error on integration test 171:
> 
> ```
>     171.triggered_by_correct_sequence (oracle): Step 37 failed:
>         objects chest_2 visible: False != True
> ```
> 
> This only happens if I run the entire test suite (not the specific test).
> 
> If no one else is seeing this, then maybe something is funky with my local build. I don't want it to hold up approval. (And I suppose Jenkins will let us know if the tests fail after merging.)

I ran the entire suite, I didn't get an error on 171 but I did get an error with `run_depth_and_segmentation_test`, but it looks more like an approx issue:

```
FAILED INTEGRATION TESTS:
    run_depth_and_segmentation_test (level1): Step 0 level1 failed: depth_map_list[0][339][376] value of 1.0898563861846924 is not close enough to expected value of 1.0899707078933716
    run_depth_and_segmentation_test (level2): Step 0 level2 failed: depth_map_list[0][339][376] value of 1.0898563861846924 is not close enough to expected value of 1.0899707078933716
    run_depth_and_segmentation_test (oracle): Step 0 oracle failed: depth_map_list[0][339][376] value of 1.0898563861846924 is not close enough to expected value of 1.0899707078933716
```

PR #645, opened 2023-02-06
MCS-1668 Added terminal_output option to MCS config file.

ThomasSchellenbergNextCentury:
See below for details.

I'm not sure whether we should include this with code freeze (0.6.5), or wait until our 0.7.0 release. It's just a config change, but it's also low priority.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -228,6 +228,21 @@ step_penalty
 
 Changes the negative penalty recieved for every step. Default: 0.001
 
+terminal_output
+^^^^^^^^^^^^^^^
+
+(boolean or string, optional)
+
+The terminal output mode controls what output metadata is logged to the terminal during a run (using the debug logger). Multiple modes can be configured together using commas-delimited strings (for example: "actions,performer").
+
+- `all` or `true` (default): Prints all terminal output (listed below).
+- `minimal`: Prints minimal output, including step number, selected action, return status, goal, and current reward (if any).
+- `actions`: Prints all actions available for the next step.
+- `objects`: Prints all object metadata.
+- `performer`: Prints metadata for the performer agent, including camera and frame rate information.
+- `scene`: Prints metadata for the scene's room, including all structural objects.
+- `false` or `none`: Prints no terminal output, besides initialization, warning, and error messages.
+
 top_down_camera
 ^^^^^^^^^^^^^^^
 
--- a/machine_common_sense/config_manager.py
+++ b/machine_common_sense/config_manager.py
@@ -307,6 +307,14 @@ class SceneObject(BaseModel):
     stack_target: Optional[bool]
 
 
+class TerminalOutputMode(Enum):
+    ACTIONS = 'actions'
+    MINIMAL = 'minimal'
+    OBJECTS = 'objects'
+    PERFORMER = 'performer'
+    SCENE = 'scene'
+
+
 class ConfigManager:
 
     DEFAULT_ROOM_DIMENSIONS = Vector3d(x=10, y=3, z=10)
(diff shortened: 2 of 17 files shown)


docs/source/install.rst
```diff
@@ -228,6 +228,21 @@ step_penalty
...
+
+The terminal output mode controls what output metadata is logged to the terminal during a run (using the debug logger). Multiple modes can be configured together using commas-delimited strings (for example: "actions,performer").
+
+- `all` or `true` (default): Prints all terminal output (listed below).
+- `minimal`: Prints minimal output, including step number, selected action, return status, goal, and current reward (if any).
+- `actions`: Prints all actions available for the next step.
+- `objects`: Prints all object metadata.
+- `performer`: Prints metadata for the performer agent, including camera and frame rate information.
+- `scene`: Prints metadata for the scene's room, including all structural objects.
+- `false` or `none`: Prints no terminal output, besides initialization, warning, and error messages.
```
ThomasSchellenbergNextCentury, line 244: New config to set the terminal output mode, which affects the amount of output logged by the ControllerLogger. The default is "all" (current behavior), and can also be "none", or subset(s) of the output. The "all" and "none" options are the parts most important to me -- the rest is somewhat arbitrary and can be refactored if you have better suggestions.

machine_common_sense/config_manager.py
```diff
@@ -307,31 +307,40 @@ class SceneObject(BaseModel):
...
     CONFIG_STEP_PENALTY = 'step_penalty'
-    CONFIG_GOAL_REWARD = 'goal_reward'
+    CONFIG_STEPS_ALLOWED_IN_LAVA = 'steps_allowed_in_lava'
+    CONFIG_TEAM = 'team'
+    CONFIG_TERMINAL_OUTPUT = 'terminal_output'
+    CONFIG_TIMEOUT = 'timeout'
     CONFIG_TOP_DOWN_PLOTTER = 'top_down_plotter'
     CONFIG_TOP_DOWN_CAMERA = 'top_down_camera'
-    CONFIG_TIMEOUT = 'timeout'
+    CONFIG_VIDEO_ENABLED = 'video_enabled'
```
ThomasSchellenbergNextCentury, line 343: Mostly just alphabetizing variables.

machine_common_sense/controller_logger.py
```diff
@@ -13,50 +15,184 @@ class ControllerLogger(AbstractControllerSubscriber):
...
+                    f"{step_output.goal.steps_allowed_in_lava}"
+                )
+            if step_output.goal.triggered_by_target_sequence:
+                logger.debug(
+                    f"  TRIGGERED-BY-TARGET-SEQUENCE: "
+                    f"{step_output.goal.triggered_by_target_sequence}"
+                )
+
+        if TerminalOutputMode.PERFORMER in mode:
+            logger.debug("PERFORMER METADATA:")
```
ThomasSchellenbergNextCentury, line 63: Here, and through the rest of this file, I expanded on the output that was being logged to the terminal. This all seemed like useful information, and the parts that don't change over the course of the run (like the goal category, or the camera properties) I moved into "on_start_scene" so they are only displayed once. Many of these logs use the "restricted_step_output" and thus don't appear at non-oracle metadata levels. The formatting is arbitrary and can be reworked if people have a strong preference.

phile-caci: Looks good, as you stated above, these looks like it's only displayed once:

STARTING NEW SCENE:
METADATA TIER:
GOAL METADATA:
PERFORMER METADATA:
SCENE METADATA:

phile-caci (approved):
Looks good, tested all options for terminal_output and ran through all the unittest.

machine_common_sense/controller_logger.py
brianpippin: Like the fact that not all data needs to be terminal logged every step.

brianpippin (approved):

jaudick (approved):

PR #651, opened 2023-03-02
MCS-1707 Added optional pickup_number property in goal metadata for multi-retrieval reward calculations of ambiguous scenes.

ThomasSchellenbergNextCentury:
Throwing this up here for discussion. I believe we've done this in the past where we've "patched" TA1 EC2 instances by overwriting specific files (in this case, `reward.py`) that have been downloaded in their local python virtual environment.

Changes:
- Added an optional "pickup_number" property to the goal metadata. If a scene's goal has a "pickup_number", that's the number of targets which must be picked up in order to attain the reward. Therefore in ambiguous Number Comparison and Arithmetic scenes, we can label ALL soccer balls as "targets", but have the "pickup_number" be half of the total.
- Scenes without a "pickup_number" should still work as before: all of the "targets" must be picked up in order to attain the reward.

TODOs:
- Add an integration test as part of this PR.
- Update the Number Comparison and Arithmetic hypercube code (MCS-1708)

### DIFF
--- a/machine_common_sense/reward.py
+++ b/machine_common_sense/reward.py
@@ -76,19 +76,27 @@ class Reward(object):
         metadata = goal.metadata or {}
         # Different goal categories may use different property names
         target_names = ['target', 'targets']
+        # Get the total number of targets.
+        pickup_number = metadata.get('pickup_number') or 0
         for target_name in target_names:
             # Some properties may be dicts, and some may be lists of dicts
             targets = metadata.get(target_name) or []
             targets = targets if isinstance(targets, list) else [targets]
+            # If pickup_number was not defined, use the list's length.
+            if not pickup_number:
+                pickup_number = len(targets)
             for target in targets:
                 goal_id = target.get('id')
                 goal_object = Reward.__get_object_from_list(objects, goal_id)
                 if goal_object:
                     goal_objects.append(goal_object)
 
-        # Only attain the reward if all targets are picked up
-        if goal_objects and all(
-                [obj.get('isPickedUp', False) for obj in goal_objects]):
+        picked_up = len([
+            object_metadata for object_metadata in goal_objects
+            if object_metadata.get('isPickedUp', True)
+        ])
+        # Attain the reward if the required number of targets were picked-up.
+        if goal_objects and pickup_number and picked_up >= pickup_number:
             reward = goal_reward
 
         return round(reward, 4)
--- a/machine_common_sense/reward.py
+++ b/machine_common_sense/reward.py
@@ -93,7 +93,7 @@ class Reward(object):
 
         picked_up = len([
             object_metadata for object_metadata in goal_objects
-            if object_metadata.get('isPickedUp', True)
+            if object_metadata.get('wasPickedUp')
         ])
         # Attain the reward if the required number of targets were picked-up.
         if goal_objects and pickup_number and picked_up >= pickup_number:
--- a/integration_tests/data/193.multi_retrieval_pickup_number.actions.txt
+++ b/integration_tests/data/193.multi_retrieval_pickup_number.actions.txt
@@ -1,0 +1,2 @@
+﻿LookDown
+PickupObject,objectId=soccer_ball_1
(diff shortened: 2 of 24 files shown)


ThomasSchellenbergNextCentury:
Per our conversion in today's standup, we'll be addressing this issue for Eval 6 via the ingest code, so this PR isn't a high priority, but we'll still need to have this PR (or implement another solution) for Eval 7.

phile-caci (approved):
Looks good to me:

added pickup_number to one of the NumberComparison scenes
```
  "goal": {
    "category": "retrieval",
    "metadata": {
      "target": {
        "id": "470ba0e4-8f65-43d9-a780-73eed87eee88",
        "pickup_number": 2
      }
    },
```

In the console, saw:
```
GOAL METADATA:
  CATEGORY: multi retrieval
  DESCRIPTION: Find and pick up as many objects as possible of type: small light black white rubber ball.
  LAST STEP: 2500
  TARGET 1 / 2: 835a1f54-6d87-4084-8771-d85608e9e72b
  TARGET 2 / 2: 69536878-cdb5-4557-b3cd-0db7dccdb482
```

brianpippin:
I am thinking the best way to handle this that would require the least amount of logic is the same way we are doing the scoring in Ingest for this evaluation.  The goal has the pickup number and then rewards keeps tracking of the number of unique objects picked up.  If the unique number of pick up equals the number required in the goal then we add the +1 to the score.  I feel like this would be the cleanest solution and not involve doing any half calculations in the scoring module.  I think this would also allow all balls to be labeled as targets.   I think we shouldn't rush this fix, and definitely want to get an integration test and make sure we examine all the hypercube cells so we don't miss any cases.  Thoughts?

ThomasSchellenbergNextCentury:
> The goal has the pickup number and then rewards keeps tracking of the number of unique objects picked up. If the unique number of pick up equals the number required in the goal then we add the +1 to the score. I feel like this would be the cleanest solution and not involve doing any half calculations in the scoring module. I think this would also allow all balls to be labeled as targets.

This is the way it's currently implemented in this PR, right? So we're on the same page :)

Unless you're talking about labeling all of the balls as targets in ALL of the scenes (not just ambiguous ones), in which case, that seems fine. I made ticket MCS-1708 to handle the corresponding Scene Generator updates, so we can tackle it there.

One thing I just realized, that I'll need to do next as part of testing, is to avoid returning the "pickup_number" at lower-than-oracle metadata levels, since that seems like revealing too much information. Leaving that here as a reminder to myself.

machine_common_sense/reward.py
```diff
@@ -76,19 +76,27 @@ def _calc_retrieval_reward(
...
                 goal_object = Reward.__get_object_from_list(objects, goal_id)
                 if goal_object:
                     goal_objects.append(goal_object)
 
-        # Only attain the reward if all targets are picked up
-        if goal_objects and all(
-                [obj.get('isPickedUp', False) for obj in goal_objects]):
+        picked_up = len([
+            object_metadata for object_metadata in goal_objects
+            if object_metadata.get('isPickedUp', True)
```
brianpippin, line 96: Doesn't this code here only count the number of objects currently picked up?  We need to keep track of how many different objects were already picked up, even if they were dropped, so I think this code will need some modifications to keep track of anytime an object was ever picked up.  Unless I am mistaken about the "isPickedUp" flag.

ThomasSchellenbergNextCentury: That's true. I don't think we had discussed the "picking up a target and then dropping it" case before I posted this. We'll need to fix that as you said (by tracking the picked-up targets) before we merge this.

ThomasSchellenbergNextCentury:
@brianpippin and team: I've addressed the dropped target issue.

Because the Reward class is stateless (its functions are static), I decided it was better to track this new state -- whether the object has ever been picked up -- using the object metadata. Since we're already using the object metadata to access the `isPickedUp` property, changing the property (to `wasPickedUp`) was fairly simple. This property will change from `false` to `true` when an object is first picked up, and stay `true` even if the object is dropped.

The property is implemented in this PR (you'll need to build from this branch to run the new tests): https://github.com/NextCenturyCorporation/ai2thor/pull/315

I've also added new integration tests for both the `pickup_number` and dropping targets.

jaudick (approved):

PR #652, opened 2023-03-07
MCS-1713 Fixed a bug with the new MCS config options accidentally removing important metadata from the scene history files.

ThomasSchellenbergNextCentury:
Problem: Previously the checks for the new MCS config file options, like `is_position_disabled()`, incorporated the metadata tier, automatically overriding the config file setting. This doesn't work when we need to save that data in the scene history files for use in the scorecard. We need Unity to return this information during the evals, regardless of the metadata tier used, and include it in the unrestricted step metadata, so it makes its way into the scene history file.

Solution: Remove metadata tier overrides in checks like `_is_position_disabled()`, so when we run the evals at metadata level 2, we can still save important metadata like position in the scene history. Only setting the specific MCS config file properties (like `disable_position = true`) will affect the checks. The StepMetadata output is still correctly restricted (as it was before) due to the `get_restrictions` function in `controller_output_handler.py`.

We'll probably need to patch the affected file (`config_manager.py`) for running the eval.

Comments or concerns about this approach?


phile-caci (approved):
Looks good, all test pass.

I didn't see the position output in the console at metadata level2 (I assume that's right?) , but I did see position information in the history file after the fix:

BEFORE:
```
                "performer_radius": 0.251,
                "performer_reach": 1.0,
                "physics_frames_per_second": 20.0,
                "position": null,
```

AFTER:
```
                "performer_radius": 0.251,
                "performer_reach": 1.0,
                "physics_frames_per_second": 20.0,
                "position": {
                    "x": -8.936,
                    "y": 0.762,
                    "z": 9.265
                },
```

rartiss55 (approved):
Did some testing with step output vs history files + ingest, LGTM! 

I did notice that segmentation_colors is now in the history files -- wondering whether or not we need it there/if there's any worry about that making the history files unnecessarily large (may not be a concern though/I don't want to make this patch more complicated than needed, just thought I'd mention it).

ThomasSchellenbergNextCentury:
@rartiss55 I think the segmentation_colors were there previously... For example, look at `mcs-ingest/tests/test_data/imitation_eval_5_ex_1_history.json`

If we think it (or any other properties) won't be necessary, we can probably remove it via the HistoryWriter, but I'm hesitant to do that now, given the timeframe.

jaudick (approved):

PR #653, opened 2023-04-04
MCS 1650 - Added "only_return_goal_object" integration tests

NDCACI:
Single-retrieval goal. Used test 109 to make test 189.

Intuitive physics scene. Used test 150 to make test 190.

Imitation scene. Used test 176 to make test 191.

Normal “passive” scene. Used test 69 to make test 192.


integration_tests/data/192.intuitive_physics_only_return_goal_object.scene.json
```diff
@@ -0,0 +1,119 @@
...
+                  "y": -0.20,
+                  "z": 0
+                }
+              }
+            ]
+          }
+    ],
+    "goal": {
+      "category": "passive"
+    }
```
ThomasSchellenbergNextCentury, line 118: I'm getting an error if there isn't a `metadata` property defined in the `goal`. You can change it to something like this:

```json
"goal": {
    "category": "passive",
    "metadata": {}
}
```

integration_tests/data/189.single-retrieval_only_return_goal_object.scene.json
```diff
@@ -0,0 +1,45 @@
+﻿{
+    "name": "soccer ball can be picked up and will roll an expected distance",
```
ThomasSchellenbergNextCentury, line 2: You should update the `name` property in each of these scene files to be more descriptive for the new tests. For example, this `name` could be something like "only return goal object in single retrieval scene"

integration_tests/data/190.intuitive_only_return_goal_object.scene.json
```diff
@@ -0,0 +1,116 @@
...
+        "position": {
+            "x": 0,
+            "z": 0
+        },
+        "rotation": {
+            "x": 0,
+            "y": 0
+        }
+    },
+    "objects": [
```
ThomasSchellenbergNextCentury, line 18: I'm realizing for this test (190), and test 192, we should probably include a non-structural object in the scene too. You can just copy the `soccer_ball` from test 189 and add it to the `objects` list here -- I think that will work fine.

ThomasSchellenbergNextCentury:
There's also a note at the bottom of the ticket description to update either test 182 or 186 (not both) to pick up all of the soccer balls.

ThomasSchellenbergNextCentury (changes requested):
1. I was still getting an error when trying to run test 192. I think we weren't handling an empty metadata dict correctly. I pushed a fix for the issue to this branch (hope you don't mind).
2. The names of the files for test 192 still have "intuitive_physics" in them; it should be replaced with "passive".
3. I'm seeing errors for tests 190 and 192 at oracle metadata level. You can run the integration tests with the `--dev --autofix` flags to update the values automatically (I'm assuming they just haven't been updated yet due to your most recent changes).

integration_tests/data/186.only_return_goal_objects_true.actions.txt
```diff
@@ -1 +1,9 @@
 MoveAhead
+PickupObject,objectId=a9b7bbb2-107e-4ed2-b2cf-b789f288396c
+DropObject,objectId=a9b7bbb2-107e-4ed2-b2cf-b789f288396c
+PickupObject,objectId=1244a5fc-dfb9-48ba-a2b9-ec88b89bfbe1
+DropObject,objectId=1244a5fc-dfb9-48ba-a2b9-ec88b89bfbe1
+PickupObject,objectId=755a6b61-523a-47ee-8704-50fe856ff178
+DropObject,objectId=755a6b61-523a-47ee-8704-50fe856ff178
+PickupObject,objectId=460d9b68-63ca-433c-8b7d-84ae1d50c1ee
+DropObject,objectId=460d9b68-63ca-433c-8b7d-84ae1d50c1ee
```
ThomasSchellenbergNextCentury, line 9: This test (186) also wasn't working for me. I think you need to expand the 3 `outputs.json` files for test 186 to account for the newly added steps. (The lists in each of those files should contain one element per step, plus one for initialization; so, for this test with 9 steps, the lists should contain 10 elements.)

NDCACI: I went through all your comments and I think this is good for review again

phile-caci (approved):
Looks good to me.  No issues running the updated integration tests.

PR #654, opened 2023-04-07
MCS-1726 - added timeout

NDCACI:
First pass, seemed too simple so I feel like I missed something. No test added yet.

### DIFF
--- a/machine_common_sense/__init__.py
+++ b/machine_common_sense/__init__.py
@@ -48,7 +48,7 @@ def get_controller_with_timeout(unity_exec: str, config: ConfigManager):
     with concurrent.futures.ThreadPoolExecutor(max_workers=2) as executor:
         future = executor.submit(get_controller, unity_exec, config)
         try:
-            controller = future.result(timeout=TIME_LIMIT_SECONDS)
+            controller = future.result(timeout=config.get_controller_timeout)
             return controller
         except concurrent.futures.TimeoutError as Msg:
             logger.error("Timeout error in creating controller", exc_info=Msg)
--- a/machine_common_sense/__init__.py
+++ b/machine_common_sense/__init__.py
@@ -30,10 +30,6 @@ logger = logging.getLogger(__name__)
 # Set default logging handler to avoid "No handler found" warnings
 logger.addHandler(logging.NullHandler())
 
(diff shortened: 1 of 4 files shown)


ThomasSchellenbergNextCentury (changes requested):
Oof, OK. My fault, for not realizing there are actually two different timeouts in the code:
- A timeout for each action step, which is set by CONFIG_TIMEOUT, which you've modified here (see `_check_for_step_timeout` in `controller.py`)
- A timeout for initializing the controller, which was the one we wanted to be able to set for this ticket (see `TIME_LIMIT_SECONDS` and `get_controller_with_timeout` in `__init__.py`)

I'm thinking it's important to keep these settings separate (at least for now), so, in order to focus on the second timeout, if you could please:
1. Add a new config setting to `config_manager.py` (maybe `CONFIG_CONTROLLER_TIMEOUT = 'controller_timeout'` ?), with a corresponding getter (you can use `180` as the default value)
2. In `__init__.py`, in `get_controller_with_timeout`, use the configured "controller_timeout" instead of TIME_LIMIT_SECONDS
3. Add new unit tests
4. Add a new section for the "controller_timeout" to “Config File Properties” in `docs/source/install.rst`

You knew this was too easy! :)

machine_common_sense/__init__.py
```diff
@@ -48,7 +48,7 @@ def get_controller_with_timeout(unity_exec: str, config: ConfigManager):
     with concurrent.futures.ThreadPoolExecutor(max_workers=2) as executor:
         future = executor.submit(get_controller, unity_exec, config)
         try:
-            controller = future.result(timeout=TIME_LIMIT_SECONDS)
+            controller = future.result(timeout=config.get_controller_timeout)
```
ThomasSchellenbergNextCentury, line 51: You should remove the TIME_LIMIT_SECONDS variable.

ThomasSchellenbergNextCentury, line 47: Need parentheses after the function call (`config.get_controller_timeout()`)

ThomasSchellenbergNextCentury (approved):
Looks great!

jaudick (approved):

brianpippin (approved):

PR #658, opened 2023-04-20
MCS-1731 tools move tools

jaudick:
[Unity](https://github.com/NextCenturyCorporation/ai2thor/pull/317)

The names in the scene.json explain what each does test in more depth.


ThomasSchellenbergNextCentury:
1. I'm getting a bunch of integration test failures due to missing files.
2. Not important at all, but I think you skipped some of the test numbers: 198, 201, 203, 205, 207

jaudick:
@ThomasSchellenbergNextCentury Jeez there are so many missing. I will do some investigating

jaudick:
@ThomasSchellenbergNextCentury I believe the tests are all here now. I changed 212 and 213 to test a new condition. Previously they were rect on rect tool pushing the tool out of the way but I spent a whole day trying to get the oracle output to be consistent. It would always be 0.1 off here or there so I changed it to test that they dont stick if they are at the minimal distance of seperation. The hooked tool test was consistent and tests pushing the tool out of the way so I think thats good enough for now.

phile-caci (approved):
Rebuilt AI2THOR with companion ticket. Ran all integration test
Viewed 197-203 - ball movement looks good
Viewed 204-217 - tool movement looks good (Question on 209 below)

Question, on 209 when the rect tool was directly behind the hooked tool it's ok that it can move the hooked tool to the side when it touches (I'm assuming it is)?

rartiss55 (approved):
LGTM + tests all pass

ThomasSchellenbergNextCentury (approved):
Great tests. I particularly appreciated number 200, which checks the use case MESS presented to us last fall. Thank you for doing such a thorough job!

PR #661, opened 2023-05-11
MCS-1796 Fixed frame off-by-one error in webenabled UI

ThomasSchellenbergNextCentury:
1. When performing an action, wait for the new image to become ready
2. When selecting a scene, display the initial image


jaudick:
@ThomasSchellenbergNextCentury The only way I can get this actually show an image after frame 0 is by holding down a button which forces it to be one frame behind. Then the images appear. So if I dont hold a button and input something every second or so then I can never see the image appear.

ThomasSchellenbergNextCentury:
@jaudick I'm assuming this happens fairly consistently? Maybe we should try a screen share tomorrow so I can see what you're talking about.

Would you mind share the complete terminal logs?

phile-caci (approved):
This worked for me.  I'm no longer a frame behind when I move around a scene.

ThomasSchellenbergNextCentury:
@jaudick @rartiss55 I've tested with Python 3.10 on Ubuntu (with the same versions of the python dependencies that Jacob's using) and am still not seeing any issues. Since the problem appears to affect rendering the images on the webpage, I'm wondering if it's actually a browser issue? What browser versions do you have?

Chrome: 112.0.5615.165
Firefox: 113.02

rartiss55:
@ThomasSchellenbergNextCentury For me this seems like a timing issue. Might be worth mentioning that my human input mode has been one step behind for a while now. When I went through things on this branch, it seemed like when I made a step, the new image would load in the specified directory AFTER the reference to the older one was already sent as the "up to date" version in the HTML. Additionally, sometimes the image hadn't even finished loading (especially on start) when I looked at things in the network tab, so I would sometimes get a broken image link or half loaded image like Jacob was seeing.

I played around with adding time.sleep(0.1) in a couple different places, and that seems to work for me + the most up to date image is loaded fully. We still have the issue of pressing buttons too fast that will cause issues with the images, but I believe that's logged as a separate ticket somewhere for the future.

webenabled/mcs_interface.py
```diff
@@ -123,15 +125,21 @@ def get_image_name(self):
                     if file != latest_file:
                         os.unlink(file)
 
+                # wait to make sure we've finished loading the new image
+                # (not sure why the is_file_open check below didn't
+                # do the trick?)
+                time.sleep(0.1)
```
rartiss55, line 131: I added some logger lines for testing with is_file_open -- this never returned False for me, so not sure if it just doesn't work for me, or the timing is off, or if it works at all? Might be worth revisiting.

rartiss55 (approved):
Hopefully the time.sleep() calls work for everyone. Approving until told otherwise :)

PR #662, opened 2023-05-11
Minor improvements to webenabled UI

ThomasSchellenbergNextCentury:
1. Separated the scene name and action list on the webpage
2. Added a temporary step number to the webpage
3. The "cmd" and "img" directories now use the same datetime string and also include the session user ID
4. Updated the gitignore with the webenabled run files


phile-caci (approved):
Looks good on my end!

rartiss55 (approved):
Code looks good + confirmed the new file names in my img directory.

PR #674, opened 2023-06-16
MCS-1869 Fixes for Doc Publish

phile-caci:
Fixes to for Doc Publish (hot fixed on master to get release deployed)

Looks like deploy has gone through successfully.  There will be errors with the PYPI push since that was successful the first time, subsequent workflow runs (since doc-publish and pypi push are part of the same workflow) show the PYPI fails because it already exists.

### DIFF
--- a/.github/workflows/docs-publish.yaml
+++ b/.github/workflows/docs-publish.yaml
@@ -6,17 +6,24 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    permissions:
+      pages: write 
+      id-token: write
+      contents: write
+      pull-requests: write
     steps:
     - name: Checkout
-      uses: actions/checkout@master
+      uses: actions/checkout@v3
       with:
         fetch-depth: 0
     - name: Build and Commit
-      uses: sphinx-notes/pages@master
+      uses: sphinx-notes/pages@v3
       with:
         documentation_path: docs/source
+        publish: false        
     - name: Push changes
       uses: ad-m/github-push-action@master
       with:
-        github_token: ${{ secrets.GITHUB_TOKEN }}
+        github_token: ${{ secrets.PLE_PAT_ACTIONS }}
         branch: gh-pages
+        force: true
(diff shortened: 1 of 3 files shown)


NDCACI (approved):

.github/workflows/docs-publish.yaml
```diff
@@ -6,17 +6,24 @@ on:
...
       with:
         documentation_path: docs/source
+        publish: false        
     - name: Push changes
       uses: ad-m/github-push-action@master
       with:
-        github_token: ${{ secrets.GITHUB_TOKEN }}
-        branch: gh-pages
\ No newline at end of file
+        github_token: ${{ secrets.PLE_PAT_ACTIONS }}
```
ThomasSchellenbergNextCentury, line 27: You had to create a new github token for this?

docs/requirements.txt
```diff
@@ -0,0 +1,35 @@
...
+requests==2.31.0
+scikit-image==0.19.3
+Shapely==1.7.1; python_version<"3.10"
+Shapely==1.8.5; python_version>="3.10"
+Sphinx==4.3.2; python_version<"3.8"
+Sphinx==6.1.3; python_version>="3.8"
+sphinxcontrib-websupport==1.2.4
+sphinx-markdown-builder==0.5.5
+typeguard==3.0.2
+git+https://github.com/sphinx-contrib/video.git
```
ThomasSchellenbergNextCentury, line 35: I'm assuming there's no way to use the requirements file in the root folder for this? I fully expect to update one file in the future, and forget to update the other...

ThomasSchellenbergNextCentury (approved):
Left a few questions, but looks fine overall.

docs/requirements.txt
phile-caci: I probably could, right now it looks like it only uses the docs folder, but I can probably add another github action or edit the jenkins build script to copy the root requirements to the docs folder.

PR #677, opened 2023-06-21
MCS-1808 - Fix shortcut keys for Pass and EndScene in human input mode 

rartiss55:
This was originally a webenabled ticket to enable the "spacebar = Pass" action, but I realized the problem was actually in human input mode:
1.) The spacebar was being used as a shortcut for two different actions (Pass and EndScene)
2.) Input in human input mode is stripped, so the spacebar didn't actually do anything.

This should be fixed now + can be tested in human input mode (spacebar will now pass, "q" to EndScene).

### DIFF
--- a/machine_common_sense/action.py
+++ b/machine_common_sense/action.py
@@ -749,7 +749,6 @@ class Action(Enum):
         Unexpected error; please report immediately to development team.
     """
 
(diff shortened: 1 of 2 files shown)


phile-caci (approved):
Looks good. No issues in unittest/integration

`q` now ends scene
` ` now passes

PR #678, opened 2023-06-26
MCS-1807 - [Webenabled] Avoid getting out of sync images with quick keypresses

rartiss55:
For webenabled: if you pressed the keys too fast, the images would get out of sync and sometimes not load properly. I went with the option of ignoring extra keypresses if an action was currently being processed.


phile-caci (approved):
@rartiss55, this looks good.  I like going with the option of ignoring the key presses. Tested by rapid firing a key, looks like they got ignored and never got out of sync.

PR #680, opened 2023-07-05
MCS-1809 [Webenabled] loading spinner

rartiss55:
Added a loading spinner for when:
- The user first goes to the page and is loading the unity app for the first time
- When a scene is loaded by the user
- When actions are being processed


webenabled/templates/mcs_page.html
```diff
@@ -48,10 +48,6 @@
             margin: 0;
         }
 
-        div#header {
```
rartiss55, line 51: Removing old/unused css here.

phile-caci (approved):
LGTM, saw spinner when:

- Loading Page
- Loading Scene
- Performing Action

PR #682, opened 2023-07-13
MCS-1811 - [Webenabled] support all actions

rartiss55:
Based off of the 1810 branch:

- All actions are supported with the exception of "EndScene"
- Added inputs for the various parameters that can be used for certain actions. There is also client side validation in case somehow an invalid value is entered, but these checks will only happen if an action related to the input is triggered.
- Changed the default image coordinates to be the center of the image (that's how human input mode works currently, and I figured this way would be more consistent with how the newly added parameters are working, since they all have defaults).
- Everything is sort of just dumped on the screen right now (information about parameters + inputs), we have another ticket to redo the layout, so I'm punting until then. 
- Updated the keyboard shortcut infographic
- minor bug fixes


webenabled/mcs_interface.py
```diff
@@ -22,7 +22,7 @@
 
 def convert_key_to_action(key: str, logger):
     for action in mcs.Action:
-        if key == action.key:
+        if key.lower() == action.key:
```
rartiss55, line 25: ignoring case now for action keys.

webenabled/templates/mcs_page.html
```diff
@@ -225,9 +279,15 @@ <h1>MCS Demo</h1>
     // ---------------------------------------------------------
     // Handle clicking on one of the scene file names
     var lis = document.getElementById("scenes_unordered_list").getElementsByTagName('li');
-    let actionKeysWithParams = ['1', '3', '4', '5', '6', '7', '8', '9', 'm', 't']
+    let actionKeysWithParams = ['1', '3', '4', '5', '6', '7', '8', '9', 'm', 'M', 't', 'T']
+    let arrowKeys = ["ArrowDown", "ArrowUp", "ArrowLeft", "ArrowRight"]
```
rartiss55, line 283: Decided to ignore arrow key input entirely for keypress event/action processing, since perhaps someone could just be using those keys to scroll.

NDCACI (approved):

PR #683, opened 2023-07-19
MCS-1812 - [Webenabled] incorporate step output metadata

rartiss55:
Branched off MCS-1811...

Created a very minimal json output file with the following step metadata (we can always add other things if we deem them valuable later):
- reward
- return status
- steps_on_lava
- step_number

+ Also added last action taken, goal information, and the available actions list should be updated on a per step basis now. 

Most of this information is displayed below the unity image in the UI for now.

I will likely leave this one open a while to allow people on different platforms to test. With the creation of another output file, I did run into some timing issues that I hopefully resolved, but want to make extra certain that there are no issues since we've seen differences on different platforms with that before.


webenabled/README.md
```diff
@@ -54,7 +54,7 @@ http://<machine.ip.address>:8080/mcs
...
-3. The `run_scene_with_dir` script creates the MCS Controller and starts a watchdog Observer to watch for changes to the `cmd_<time>/` directory. Whenever a new "command" text file is created or modified, it triggers the Observer, which reads the command (either a scene filename ending in ".json" or an MCS action like Pass or MoveAhead), gives it to the MCS Controller via either its `start_scene` or `step` function, and saves the output image in the `img_<time>` directory.
+2. The `MCSInterface` class creates the `cmd_<time>/` and `output_<time>/` directories in `static/mcsinterface/` for the input and output data for the current user session. It starts a subprocess in a separate thread which runs `run_scene_with_dir.py`. It watches the `output_<time>/` directory for step output and images the subprocess saves in that directory, which are the images returned by action steps.
```
rartiss55, line 57: Ended up renaming a lot of "img" specific things to "output", since I'm now saving step output metadata as well as image files in an output directory.

NDCACI (approved):

webenabled/README.md
ThomasSchellenbergNextCentury: Makes sense. And thanks for updating these docs!

webenabled/templates/mcs_page.html
```diff
@@ -189,14 +189,18 @@ <h1>MCS Demo</h1>
...
                 <span id="action_list">N/A</span>
             </div></p>
+            <p><div id="goal_cat_div">
+                <span>Goal Category: </span>
+                <span id="goal_cat">N/A</span>
+            </div></p>
+            <p><div id="goal_desc_div">
+                <span>Goal Description: </span>
+                <span id="goal_desc">N/A</span>
+            </div></p>
```
ThomasSchellenbergNextCentury, line 203: For discussion: I'm wondering if we should reorder the info in this middle column. I'm thinking this order:

- Scene and goal info
- Output from previous step, including next available actions
- Image
- Input for next step, including instructions

Thoughts?
-

rartiss55: This definitely needs to change, I vote punting until the later UI ticket. I was actually thinking we might want to utilize more of the left hand side space for some information (if it makes sense) since we'll likely have a dropdown for the scenes.

ThomasSchellenbergNextCentury (approved):
LGTM. Tested it a bunch and didn't find any bugs related to this PR.

phile-caci (approved):
Looks good on my end.  No need to reply, just a thought, maybe for a future implementation we can keep all the metadata for each step on the browser (like in a textarea).

PR #684, opened 2023-08-03
MCS-1910 [Webenabled] error handling

rartiss55:
This was to slightly improve the experience of when the user has an unexpected error from the Unity app. The main things I tested were:
- invalid actions (i.e. only available action is "Pass" but I tried "MoveAhead")
- trying an action when the unity app is no longer running

There should be an error displayed in these cases instead of return status (and an error file saved in the output folder).


phile-caci (approved):
Looks good on my end:

image

image

ThomasSchellenbergNextCentury (approved):
Code looks good. Works for me.

PR #686, opened 2023-08-14
MCS-1816 [Webenabled] Improve styling

rartiss55:
Some notes:
- Ended up making things look similar to the analysis UI website, since we already kind of do that for the home page as well with the similar colors + didn't want to get too whacky with color scheme. 
- Used the footer we use everywhere -- I can always change it to something else. 
- I had originally intended to utilize more space on the left for some of the outputs, but it didn't actually make sense in practice, so I went with Thomas' suggestion on how to order things in the middle. 
- Wanted to find different ways to display the shortcut panel (tried making it a collapsible div but it didn't quite work out right with the new layout). Originally left it more or less the same as it is, but then last minute changed things so that its on the bottom left (thought it'd be nice to save some screen real estate on a smaller monitor, but maybe that doesn't matter so much). Here's how the two different versions look in case anyone has a preference:

Screenshot 2023-08-14 at 10 57 28 AM
vs 
Screenshot 2023-08-14 at 10 50 05 AM

- Some of my changes were nitpicks about naming conventions ("some_div" vs "some-div" for instance), so apologies for the diffs!

### DIFF
--- a/webenabled/templates/mcs_page.html
+++ b/webenabled/templates/mcs_page.html
@@ -39,7 +39,7 @@
 
         div#scene-nav-info {
             flex: 1;
-            min-width: 350px;
+            min-width: 400px;
         }
 
         .section {
@@ -66,7 +66,6 @@
         }
 
         #scenes-dropdown {
-            max-width: 300px;
             margin: 5px 5px 10px 10px;
         }
 
@@ -143,6 +142,26 @@
             }
         }
 
+        select {
+            width: 90%;
+            font-family: inherit;
+            font-size: inherit;
+            line-height: inherit;
+            padding: 3px;
+        }
+
+        input {
+            font-family: inherit;
+            font-size: inherit;
+            margin-left: 4px;
+            margin-right: 4px;
+        }
+
+        input.checkbox {
+            width: 16px;
+            height: 16px;
+        }
+
         .input_fields {
             display: flex;
             justify-content: space-between;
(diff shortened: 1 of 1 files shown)


webenabled/templates/mcs_page.html
```diff
@@ -190,138 +226,139 @@
...
-            <p>
-                <div>
-                    For image coordinates, the default is the center of the scene image. 
-                    Click a point on the Unity image to populate with a different coordinate.
+    <div id="content-wrapper">
+        <div class="left-panel column">
+            <div id="scene-nav-info" class="section">
+                <p><strong>Scenes</strong>
+                <p>Pick from one of the following scenes:</p>
+                <select id="scenes-dropdown" size="15">
```
rartiss55, line 236: Note that I did make this a select + originally had it as a dropdown, but IMO it looks nicer as a scrollable select since there's so many options.

NDCACI (approved):

ThomasSchellenbergNextCentury (commented):
Overall, I love it! Awesome job. I approve removing the right column. I think it only makes sense to keep all three columns if we need to add more content to them, but we can always add the column back if it's ever necessary.

Action items:
- If you reduce the width of the browser, the keyword image needs to scale down appropriately.
- I'd suggest combining the Goal Category and Goal Description into one line. Something like "Goal: (Retrieval) Find and pick up the ball."
- I think we should move the Image Coordinates and input fields above the help text ("Parameters will be used if applicable..."), so they're easier to see.
- Can we display a message when the page loads saying "Please select a scene"?

rartiss55:
@ThomasSchellenbergNextCentury Thanks for the feedback! I believe I addressed all of your comments, let me know if there's anything else.

ThomasSchellenbergNextCentury (approved):

PR #687, opened 2023-08-14
MCS-1925 - [Webenabled] Ignoring non-action keys

rartiss55:
Original ticket was to not submit a pass action for function keys -- decided to simply ignore all non-action keys. Also addressed a small bug where you could submit an action before actually choosing a scene. 

Note that this is branched off of MCS-1816-styling.

### DIFF
--- a/webenabled/templates/mcs_page.html
+++ b/webenabled/templates/mcs_page.html
@@ -360,8 +360,7 @@
     // Handle clicking on one of the scene file names
     var sceneSelect = document.getElementById("scenes-dropdown");
     let actionKeysWithParams = ['1', '3', '4', '5', '6', '7', '8', '9', 'm', 'M', 't', 'T']
-    let allOtherActionKeys = ['2', 'w', 'W', 'a', 'A', 's', 'D', 'h', 'H', 'i', 'I', 'j', 
-    'J', 'k', 'K', 'L', 'l', ' ', 'q', 'Q']
+    let allOtherActionKeys = ['2', 'W', 'A', 'S', 'D', 'H', 'I', 'J', 'K', 'L', ' ', 'Q']
     var processingKeypress = false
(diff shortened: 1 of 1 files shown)


ThomasSchellenbergNextCentury (approved):
That does the trick for me.

Quick observation (probably a separate ticket): When I hit spacebar, my browser automatically scrolls down, which is unfortunate.

webenabled/templates/mcs_page.html
```diff
@@ -360,7 +360,8 @@ <h1>Machine Common Sense</h1>
     // Handle clicking on one of the scene file names
     var sceneSelect = document.getElementById("scenes-dropdown");
     let actionKeysWithParams = ['1', '3', '4', '5', '6', '7', '8', '9', 'm', 'M', 't', 'T']
-    let arrowKeys = ["ArrowDown", "ArrowUp", "ArrowLeft", "ArrowRight"]
+    let allOtherActionKeys = ['2', 'w', 'W', 'a', 'A', 's', 'D', 'h', 'H', 'i', 'I', 'j',
```
phile-caci, line 363: Think we're missing lower case 'd'.

ThomasSchellenbergNextCentury: I think we could just use uppercase letters and say `!allOtherActionKeys.includes(e.key.toUpperCase())`

rartiss55:
> That does the trick for me.
> 
> Quick observation (probably a separate ticket): When I hit spacebar, my browser automatically scrolls down, which is unfortunate.

Good catch, the spacebar shouldn't scroll anymore.

webenabled/templates/mcs_page.html
rartiss55: Thanks for catching this -- fixed.

PR #692, opened 2023-08-29
MCS-1817 Close unity process when browser closes

rartiss55:


webenabled/mcsweb.py
```diff
@@ -131,6 +132,48 @@ def handle_keypress():
...
+
+    app.logger.info(
+        "Attempting to clean up processes after browser has been closed.")
+
+    for p in psutil.process_iter(['pid']):
+        if p.info['pid'] == controller_pid:
+            children = p.children(recursive=True)
+            for c_process in children:
+                app.logger.info(
+                    f"Found child process of controller: {c_process}, "
```
rartiss55, line 153: Child process here should be the unity process. Did it this way so we don't delete an incorrect unity process (in case two different ones are running).

webenabled/mcsweb.py
```diff
@@ -131,6 +132,48 @@ def handle_keypress():
...
+
+    app.logger.info(
+        f"Clear user session for: {unique_id}")
+    del session[unique_id]
+
+    resp = jsonify(
+        ended_controller_process=controller_pid,
+        ended_session=unique_id
+    )
+    resp.delete_cookie('uniq_id')
```
rartiss55, line 172: I don't think this cookie actually gets deleted from future requests because the browser window is closed, but for future requests, it checks/knows that the controller associated with this id is no longer running, so it'll create a new "user".

webenabled/templates/mcs_page.html
```diff
@@ -652,6 +652,15 @@ <h1>Machine Common Sense</h1>
...
     }
 
+
+    function cleanup(e) {
+        resp = fetch("{{url_for('exit_unity')}}",
+            {method: 'POST', body: JSON.stringify(this.innerHTML)})
+            .then(parseJsonResponse)
+    }
+
+    window.addEventListener('beforeunload', this.cleanup, false);
```
rartiss55, line 662: Note that this works for closing the window, but it also is activated in the case where the user refreshes the page.

ThomasSchellenbergNextCentury (commented):
If you open a second tab, it now uses the same Unity application/process, rather than starting a second one. This means you can change scenes and/or perform actions using the second tab, which also affects the first tab, but the first tab doesn't update. I'm not sure if we care, but thought I'd point it out for discussion.

rartiss55:
> If you open a second tab, it now uses the same Unity application/process, rather than starting a second one. This means you can change scenes and/or perform actions using the second tab, which also affects the first tab, but the first tab doesn't update. I'm not sure if we care, but thought I'd point it out for discussion.

@ThomasSchellenbergNextCentury For me, this was the behavior before. The only way I would get two unity windows in two tabs is if I started them very quickly / even then webenabled might end up using the same process anyway. Was this never the case for you?

Perhaps I can use localStorage or something to see if a tab is already opened, and at least display a popup if a new tab is opened warning the user. I will look into it.

ThomasSchellenbergNextCentury:
@rartiss55 I can reproduce the problem on this branch if I open a second tab before the Unity process from the first tab finishes loading. I get two Unity windows, and then things start to go a little haywire when I try selecting a scene. At some point additional Unity windows started to pop up, so I think something got out-of-sync.

However, since I remember seeing this issue before, I'm assuming the issue's not directly related to this ticket. I'm fine if we want to investigate this as a separate ticket.

ThomasSchellenbergNextCentury (approved):

phile-caci (approved):
Unity closes when I exit the browser.  LGTM!

PR #693, opened 2023-09-11
MCS-1945 [Webenabled] Restrict to one tab

rartiss55:
This is to prevent multiple dangling controller processes in webenabled. You should be able to still have multiple sessions across browsers, but now it'll be restricted to one session/controller per browser. If you are in the same browser and try to open up webenabled in another tab, you should get a warning/error message.


ThomasSchellenbergNextCentury (approved):
Works great.

phile-caci (approved):
LGTM!

PR #698, opened 2023-09-19
Mcs 1955 mcs deployment updates

phile-caci:
Tested on the release of 0.7.2, looks like there aren't any issues.

Just some notes... there's a regex replacement in the mcs version bump job that replaces the old version string with the new version string.  However, the string "0.7.1" was also found in the requirements.txt for recommonmark and replaced it with 0.7.2 which cause the build to fail.

Also updated gh-action-pypi-publish to the latest version, previous version got sunset.

Manually bumped these versions and also corrected the "bump-version-0.7.2" branch.


.github/workflows/pypi-publish.yaml
```diff
@@ -25,11 +25,11 @@ jobs:
       - name: create wheel file
         run: python setup.py bdist_wheel sdist clean --all 
       - name: Publish distribution to Test PyPI
-        uses: pypa/gh-action-pypi-publish@master
+        uses: pypa/gh-action-pypi-publish@release/v1
         with:
           password: ${{ secrets.TEST_PYPI_API_TOKEN }}
-          repository_url: https://test.pypi.org/legacy/
+          repository-url: https://test.pypi.org/legacy/
```
phile-caci, line 31: repository_url was deprecated and updated to repository-url

ThomasSchellenbergNextCentury (approved):

PR #702, opened 2023-10-13
Mcs 1957 windows auto downloader

phile-caci:
Adds downlaod support for windows.

The windows addressables and unity exec are in `development` only atm. To test, in `machine_common_sense/_version.py` change the version to 'development'.

To test:

You can then cache addressables:
```
python ./machine_common_sense/scripts/cache_addressables.py
```

Run integration:
```
cd integration_tests
python run_handmade_tests.py
```

You should see it download the `develop` version of the unity executable (MCS-AI2-THOR-Unity-App-vdevelop-{linux || win64 || mac}.zip:). I've tested the auto downloader in windows and linux, but a second set of eyes would be good. Also, can someone make sure nothing is broken in MacOS.

Thanks!


ThomasSchellenbergNextCentury (approved):
The auto-downloader worked for me on Windows and Linux.

All the unit tests now pass on Windows.

On Windows, all of my integration tests passed, except 108 (level1) and 107 (oracle). Both of these seemed like minor rounding errors, so I don't think we need to solve them immediately. I'll add a ticket to deal with it later though.

I'm still running the integration tests on Linux (I got my auto-downloader to work again -- yay), but if I run into any issues with them, they're probably not due to this PR.

Approved.

PR #707, opened 2023-11-07
MCS-1976 - Fix integration test for windows

phile-caci:
Updated output for failing test.. Reran integration for both Windows and Linux builds/addressables.   Looks like the int tests are all good (change .5 to .51 which is now passes the math.isclose check).


ThomasSchellenbergNextCentury (approved):
The tests passed for me on both Linux and Windows.

PR #717, opened 2023-12-14
MCS-1990 Trying to get the webenabled downloads to work

ThomasSchellenbergNextCentury:
Ideally, we should be able to double-click on the "mcsweb" package file (binary/executable) to start the application.

NOTES BEFORE YOU BEGIN:
- I had problems using pyinstaller with a python virtual environment installed in the MCS root folder. I'd recommend installing a new python virtual environment in the webenabled folder (see the webenabled README for details).
- You'll need to `pip install portalocker` before you use pyinstaller (I've just added it to the MCS `setup.py` file, but since it wasn't there for the latest release, you'll need to do it manually for now).
- From what I've read, you can't use pyinstaller to build an application on Linux, and then run that application on Windows; you need to use pyinstaller to build the application on Windows, and have separate packages for each OS.
- When running everything via the pyinstaller file, to open the webpage, I had to use `127.0.0.1:8080/mcs` instead of `localhost:8080/mcs`

STATUS REPORT:

Previously we tested by executing `./dist/mcsweb/mcsweb` on the command line (from the `webenabled/` folder), and this still works on Linux.

As an intermediary step, I tried executing `./mcsweb` (from the `webenabled/dist/mcsweb/` folder). There were a few problems that I've since resolved, so that also works on Linux. As you'll see in the "Package" instructions on the README, the resolution involved copying a bunch of files into the `dist/mcsweb/` folder (not ideal, but, hey, it works).

PROBLEM 1: Unfortunately, double-clicking on the binary in Linux doesn't work; you can still load the webpage, but Unity doesn't start, so clicking on a scene doesn't do anything. Since running things this way means you can't see log messages in a terminal, I don't know what sort of error is being thrown. You could try temporarily hard-coding the logger in `mcsweb.py` to write to a log file, which might get you some information.

PROBLEM 2: I can't get the Windows package to run at all, for any of these three methods; the flask server starts, but when I open the webpage, I get a ModuleNotFoundError for the watchdog python module, which should be installed correctly on my machine... I don't remember who originally tested this on Windows (maybe it was Phi?), but I'm not sure why it's not working at all anymore.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,6 +60,7 @@ setuptools.setup(
         'numpyencoder==0.3.0',
         'opencv-python==4.4.0.46; python_version<="3.9"',
         'opencv-python==4.5.4.60; python_version>="3.10"',
+        'portalocker',
         'pydantic==1.10.7',
         'requests==2.31.0',
         'scikit-image==0.19.3',
--- a/webenabled/mcs_task_desc.py
+++ b/webenabled/mcs_task_desc.py
@@ -1,7 +1,6 @@
-from enum import Enum, unique
+from enum import Enum
 
 
-@unique
 class TaskDescription(Enum):
     """
     The names for each enum correspond to the naming conventions currently
--- a/webenabled/README.md
+++ b/webenabled/README.md
@@ -1,6 +1,14 @@
-# Running MCS on the web
+# Machine Common Sense Web Browser Demo
 
-## Install
+This MCS web app lets you run scenes in your web browser, using your keyboard to move around and perform actions like an AI would.
+
+## Usage
+
+TODO
+
+## Development
+
+### Install
 
 First, "git clone" this repository. Then run the following commands from this folder to setup your python environment:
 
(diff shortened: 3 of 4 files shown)


webenabled/mcs_task_desc.py
```diff
@@ -1,7 +1,6 @@
-from enum import Enum, unique
+from enum import Enum
 
 
-@unique
 class TaskDescription(Enum):
```
ThomasSchellenbergNextCentury, line 4: For some reason this particular decorator was causing pyinstaller to fail...

setup.py
```diff
@@ -60,6 +60,7 @@ def get_version():
         'numpyencoder==0.3.0',
         'opencv-python==4.4.0.46; python_version<="3.9"',
         'opencv-python==4.5.4.60; python_version>="3.10"',
+        'portalocker',
```
ThomasSchellenbergNextCentury, line 63: If this branch never gets merged, it'd probably be good to add this to the `setup.py` on the development branch

PR #718, opened 2023-12-15
MCS - 2014 - Set Rotation from 36 to 9

NDCACI:
The quick action of turn 9 still turns the camera ~90 degrees.


phile-caci (approved):
haha, thanks for jumping on it so quickly, I will be demoing this at the PI Meeting. tested and works as expected

